PR #5, opened 2025-01-04
Logo square

cvanelteren:
Replaced the logos for the docs and updated some of the capitalization


cvanelteren:
Note probably missing a bunch but the front page is covered.

PR #6, opened 2025-01-04
license update

cvanelteren:


beckermr:
Please backout this PR. The license you chose is not appropriate for software. See

https://opensource.stackexchange.com/questions/1717/why-is-cc-by-sa-discouraged-for-code

Further, we cannot just remove the old MIT license without permission from the old contributors.

beckermr:
ping @cvanelteren

Issue #12, opened 2025-01-07
Add possibility of providing triangles while plotting tricontour


thibaultDrt:
Hello there,

I revive a ProPlot [issue](https://github.com/proplot-dev/proplot/issues/439) that is quite limiting in my field.

# Context

In Matplotlib, it is possible to plot contours or filled contours using unstructured grid with triangles, using [tricontour](https://matplotlib.org/stable/api/_as_gen/matplotlib.pyplot.tricontour.html) and [tricontourf](https://matplotlib.org/stable/api/_as_gen/matplotlib.pyplot.tricontourf.html).
The documentation shows that several ways of providing arguments are available :

```python
ax.tricontourf(x, y, z, ...)
ax.tricontourf(x, y, triangles, z, ...)
ax.tricontourf(triangulation, z, ...)
```

1. In the first case, no connectivity is provided and a Delaunay triangulation algorithm is used internally to create triangles. When dealing with concave geometries, this algorithm fills the empty spaces with undesired triangles, and therefore requires to provide the connectivities.
2. In the second case, an array of connectivity is provided, allowing to deal with concave geometries for example.
3. In the 3rd case, an object [Triangulation](https://matplotlib.org/stable/api/tri_api.html#matplotlib.tri.Triangulation) is provided, which mainly consists of coordinates and connectivities.

The management of the arguments (with or without the triangles connectivities array) is dealt [here](https://github.com/matplotlib/matplotlib/blob/75c53626f7355b4079f76b1992e6d1a45ebcf20c/lib/matplotlib/tri/_triangulation.py#L170C1-L191C51) in matplotlib.

# Anomaly in UltraPlot

However, in pro-ultraplot, [tricontour](https://github.com/Ultraplot/ultraplot/blob/6b8223a8c3dbb77662bcb8ce0afd38f133dfd0d8/ultraplot/axes/plot.py#L4404C1-L4425C17) and its filled counterpart only considers `(x, y)` arguments, and forces the Delaunay triangulation algorithm to be used.
It would be great to modify this behaviour and allow to provide the `triangles `array or `triangulation` object as argument of the function.

Thanks a lot for the help :)

cvanelteren:
Which version of MPL are you using? The docs state that two call signatures are allowed ([here](https://matplotlib.org/stable/api/_as_gen/matplotlib.pyplot.tricontourf.html))
```
tricontourf(triangulation, z, [levels], ...)
[...]

PR #13, opened 2025-01-07
Allow triangulation object in tricountour(f)

cvanelteren:
Addresses https://github.com/Ultraplot/ultraplot/issues/12

I am awaiting test cases to add to the docs to ensure that this would be fixed and tested for in the future.


thibaultDrt:
For information, the arguments for this function in matplotlib are handled [here](https://github.com/matplotlib/matplotlib/blob/75c53626f7355b4079f76b1992e6d1a45ebcf20c/lib/matplotlib/tri/_triangulation.py#L142C1-L191C51). 
It also allow to provide directly a Triangulation object instead of x, y and triangles.

For the decorator, I would know how to handle it here, with different signatures.

beckermr (changes requested):
I am pretty sure the changes here break existing functionality that allows people to pass data frames with x/y/z columns to the plotting code. This needs to be checked.

cvanelteren:
No. The input parser that is in place checks whether kwargs contains a data dict which may contain the keys of "x,y,z" which are checked for. I am not sure why this way of parsing is preferred over keeping closer to the mpl way of doing it. The parser checks if some keywords are present. Tbh it seems a bit convoluted to me in general.

cvanelteren:
I have made an adjustment where it passes through both functions now to ensure backwards compatibility. I makes the code a bit overly complicated in my opinion, but perhaps this is the way forward. Happy to know what you think.

beckermr (approved):
LGTM and nice job on the tests!

PR #15, opened 2025-01-10
Fix contourf for matplotlib 3.10 (fixes Issue #14)

austin-hoover:

Issue #14, opened 2025-01-10
Error in contourf for matplotlib version 3.10


austin-hoover:
WIth matplotlib==3.10 (ultraplot==0.99.3), `ax.contourf` throws an error:

```python
import numpy as np
import ultraplot as uplt

rng = np.random.default_rng(0)
x = rng.normal(size=(1000, 2))
values, edges = np.histogramdd(x, bins=25)

fig, ax = uplt.subplots()
ax.contourf(edges[0], edges[1], values.T)
```

```
Traceback (most recent call last):
  File "/Users/46h/repo/ultraplot/tests/test_contourf.py", line 9, in <module>
    ax.contourf(values.T)
  File "/Users/46h/repo/ultraplot/ultraplot/internals/inputs.py", line 327, in _preprocess_or_redirect
    return func(self, *args, **kwargs)  # call unbound method
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/46h/repo/ultraplot/ultraplot/axes/plot.py", line 4224, in contourf
    self._fix_patch_edges(m, **edgefix_kw, **contour_kw)  # no-op if not contour_kw
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/46h/repo/ultraplot/ultraplot/axes/plot.py", line 1811, in _fix_patch_edges
    for contour in obj.collections:
                   ^^^^^^^^^^^^^^^
AttributeError: 'QuadContourSet' object has no attribute 'collections'
```

It works with matplotlib==3.9.3. It looks like this collections attribute was removed in matplotlib version 3.8: https://matplotlib.org/stable/api/prev_api_changes/api_changes_3.8.0.html#contourset-is-now-a-single-collection.

austin-hoover:
Removing the for loop below fixes the problem. 

https://github.com/Ultraplot/ultraplot/blob/6b8223a8c3dbb77662bcb8ce0afd38f133dfd0d8/ultraplot/axes/plot.py#L1805-L1814

thibaultDrt:
Tricontourf encounters a similar issue.

This is working 👍 
```python
fig, ax = uplot.subplot()
ax.tricontourf(x, y, z, ec="none")
``` 

This is not working 👎 
```python
fig, ax = uplot.subplot()
ax.tricontourf(x, y, z)
```

```
---------------------------------------------------------------------------
[...]

PR #17, opened 2025-01-11
mpl 3.10

cvanelteren:
Fixes #15 #14  an backward compatible with mpl >= 3.8


beckermr:
Did you check that this doesn't break on earlier versions of matplotlib, namely 3.9?

cvanelteren:
Yes

cvanelteren:
it basically adds changes that have been there since 3.8. Since my latest pull are compatible from 3.9.x this does not affect the code base. I confirmed prior to working on this that it worked with 3.8, and 3.9.2/3. Should be no issue :smile:

PR #18, opened 2025-01-14
Axis sharing fix

cvanelteren:
Addresses https://github.com/Ultraplot/ultraplot/issues/11#issuecomment-2590028154

The MPL api changed in v3.6 where get_[x/y]_shared_axes returns an immutable object.

- [x] check remaining code base for use of this object
- [x] write unittests
   - [x]  to test for sharing limits
   - [x] to test for sharing labels (only)

### DIFF
--- a/ultraplot/axes/cartesian.py
+++ b/ultraplot/axes/cartesian.py
@@ -576,7 +576,7 @@ class CartesianAxes(shared._SharedAxes, plot.PlotAxes):
             if ax1.get_autoscalex_on() and not ax2.get_autoscalex_on():
                 ax1.set_xlim(ax2.get_xlim())  # non-default limits
         # Copy non-default locators and formatters
-        self.get_shared_x_axes().joined(self, sharex)  # share limit/scale changes
+        self.sharex(sharex) # get_shared_x_axes now returns an immutable object
         if sharex.xaxis.isDefault_majloc and not self.xaxis.isDefault_majloc:
             sharex.xaxis.set_major_locator(self.xaxis.get_major_locator())
         if sharex.xaxis.isDefault_minloc and not self.xaxis.isDefault_minloc:
@@ -598,7 +598,7 @@ class CartesianAxes(shared._SharedAxes, plot.PlotAxes):
                 ax1.set_yscale(ax2.get_yscale())
             if ax1.get_autoscaley_on() and not ax2.get_autoscaley_on():
                 ax1.set_ylim(ax2.get_ylim())
-        self.get_shared_y_axes().joined(self, sharey)  # share limit/scale changes
+        self.sharey(sharey) # get_shared_y_axes now returns an immutable object
         if sharey.yaxis.isDefault_majloc and not self.yaxis.isDefault_majloc:
             sharey.yaxis.set_major_locator(self.yaxis.get_major_locator())
         if sharey.yaxis.isDefault_minloc and not self.yaxis.isDefault_minloc:
--- a/ultraplot/axes/cartesian.py
+++ b/ultraplot/axes/cartesian.py
@@ -576,7 +576,10 @@ class CartesianAxes(shared._SharedAxes, plot.PlotAxes):
             if ax1.get_autoscalex_on() and not ax2.get_autoscalex_on():
                 ax1.set_xlim(ax2.get_xlim())  # non-default limits
         # Copy non-default locators and formatters
-        self.sharex(sharex) # get_shared_x_axes now returns an immutable object
+        # self.sharex(sharex)
+        self._shared_axes["x"].join(self, sharex)
+
+        # self.get_shared_x_axes().joined(self, sharex)  # share limit/scale changes
         if sharex.xaxis.isDefault_majloc and not self.xaxis.isDefault_majloc:
             sharex.xaxis.set_major_locator(self.xaxis.get_major_locator())
         if sharex.xaxis.isDefault_minloc and not self.xaxis.isDefault_minloc:
@@ -598,7 +601,8 @@ class CartesianAxes(shared._SharedAxes, plot.PlotAxes):
                 ax1.set_yscale(ax2.get_yscale())
             if ax1.get_autoscaley_on() and not ax2.get_autoscaley_on():
                 ax1.set_ylim(ax2.get_ylim())
-        self.sharey(sharey) # get_shared_y_axes now returns an immutable object
+        # self.sharey(sharey) # get_shared_y_axes now returns an immutable object
+        self._shared_axes["y"].join(self, sharey)
         if sharey.yaxis.isDefault_majloc and not self.yaxis.isDefault_majloc:
             sharey.yaxis.set_major_locator(self.yaxis.get_major_locator())
         if sharey.yaxis.isDefault_minloc and not self.yaxis.isDefault_minloc:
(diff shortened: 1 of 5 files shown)


cvanelteren:
Note to self: not sure exactly how the sharing is operating at this moment. There was an api change that made the original implementation in proplot invalid and ultraplot inherited this legacy. The proposed change makes it work as it should, but the actual labels of the axis are still internally held the same. That is, we get a proper output image, but when I check if the labels of the individual axes are in fact shared, they are not. I feel uncomfortable merging this without this feature being there -- I will need to check if older versions in fact pass this test or not.

cvanelteren:
image

cvanelteren:
Quick update.
- Proplot has the same issue with 0.95 that the axis sharing is a bit funky
- I noticed that draw is not called pre-plot. Manually drawing it does update the shared labels correctly.

cvanelteren:
Should be good now -- open for review

cvanelteren:
Pinging @beckermr` @pratiman-91 :bell:

pratiman-91 (commented):
It looks good to me! Maybe check with another reviewer.

beckermr (commented):
Relying on private class attributes is fragile. It'd be better to only use public methods / attributes.

cvanelteren:
The MPL documentation does not allow to share one axis across multiple subplots (see [here](https://github.com/matplotlib/matplotlib/blob/8d64f03a1f501ba0019279bf2f8db3930d1fe33f/lib/matplotlib/axes/_base.py#L1239-L1256)).  I tried implementing this:

```python

def _safe_share(self, axis_name, other_axis):
    if hasattr(self, f"_share{axis_name}"):
        setattr(self, f"_share{axis_name}", None)

    self._shared_axes[axis_name].join(self, other_axis)
    
    # Update the share reference
    setattr(self, f"_share{axis_name}", other_axis)
```

However this results in the axis labels not being shared properly and the ticks being messed up:

image

cvanelteren:
Note unsharing axis is also not allowed according to their docs. This would force us to either (a) write our own implementation, or abuse the api as is done here.

cvanelteren:
Perhaps we could do this

```python

   def share_axis(self, which, other):
        if not isinstance(other, plot.PlotAxes):
            raise ValueError("other must be an Axes instance")
        shared = getattr(self, f"get_shared_{which}_axes")()
        if which == "x":
            if other in shared:
                return
            self._shared_axes["x"].join(self, other)
        else:
            if other in shared:
                return
            self._shared_axes["y"].join(self, other)
  ```
  To circumenvent the error and enable proper sharing?

**Comment by beckermr** [comment]
Then it appears developing an implementation using public apis is the correct approach. 

There appears to be api instability in MPL even for public attributes. One should therefore expect that relying explicitly on private ones will be even more unstable. I think the history of new MPL versions breaking with proplot/ultraplot matches these expectations.

**Comment by cvanelteren** [comment]
Actually scratch that these methods produce error. Not sure how to proceed here.. open for feedback.

**Comment by cvanelteren** [comment]
To maintain Ultraplot's multiple axis sharing capability while respecting Matplotlib's API changes, we can override Matplotlib's public sharing methods with our own implementation. This approach allows us to bypass Matplotlib's single-axis sharing limitation while preserving Ultraplot's existing functionality. Although not ideal, this serves as a pragmatic interim solution until a more permanent resolution can be implemented.

File: ultraplot/axes/shared.py [hunk 1]
```diff
@@ -184,3 +184,40 @@ def _update_ticks(
         if kwtext_extra:
             for lab in obj.get_ticklabels():
                 lab.update(kwtext_extra)
+
+    # Override matplotlib defaults ot handle multiple axis sharing
```
beckermr, line 188:
```suggestion
    # Override matplotlib defaults to handle multiple axis sharing
```

ultraplot/tests/test_subplots.py
```diff
@@ -172,3 +172,44 @@ def test_reference_aspect():
...
+@pytest.mark.parametrize("share", ["limits", "labels"])
+def test_axis_sharing(share):
+    fig, ax = uplt.subplots(ncols=2, nrows=2, share=share, span=False)
+    labels = ["A", "B", "C", "D"]
+    for idx, axi in enumerate(ax):
+        axi.scatter(idx, idx)
+        axi.set_xlabel(labels[idx])
+        axi.set_ylabel(labels[idx])
+
+    # TODO: the labels are handled in a funky way. The plot looks fine but the label are not "shared" that is the labels still exist but they are not visible and instead there are new labels created. Need to figure this out.
```
beckermr, line 187:
```suggestion
    # TODO: the labels are handled in a funky way. The plot looks fine but
    # the label are not "shared" that is the labels still exist but they
    # are not visible and instead there are new labels created. Need to
    # figure this out.
```

ultraplot/axes/shared.py
```diff
@@ -184,3 +184,40 @@ def _update_ticks(
...
+    def sharex(self, other):
+        return self.share_axis(which="x", other=other)
+
+    def sharey(self, other):
+        self.share_axis(which="y", other=other)
+
+    # Internal function to share axes
+    def share_axis(self, which, other):
```
beckermr, line 196:
```suggestion
    # Internal function to share axes
    def _share_axis(self, which, other):
```

If this is meant to be an internal API and not public, then it should be prefixed with an underscore. Otherwise, we should remove the comment.

ultraplot/axes/cartesian.py
```diff
@@ -563,6 +563,17 @@ def _is_panel_group_member(self, other):
             and other._panel_parent is self._panel_parent  # other is sibling panel
         )
 
+    def _safe_share(self, axis_name, other_axis):
+        if hasattr(self, f"_share{axis_name}"):
+            setattr(self, f"_share{axis_name}", None)
+        getattr(self, f"share{axis_name}")(other_axis)
```
beckermr, line 569: Where is this function used? I cannot find it used in the diff, but I may have missed it. If it is not used, we should remove it. Otherwise, we need to use it somewhere, but IDK where it should go.

beckermr (commented):
Comments on the code. I still see one private attribute accessed, but this PR is closer in that we are adding APIs that we needed and/or were removed.

ultraplot/axes/cartesian.py
cvanelteren: Ah this should be removed! Apologies

ultraplot/axes/shared.py
cvanelteren: Depends, I put this currently as public as I think other objects may use this imo. Internal should be rephrased to ultraplot. We could also decide to make it internal -- open to both :man_shrugging:

ultraplot/tests/test_subplots.py
cvanelteren: Do we want multiple lines or just rely on the editor(s)?

ultraplot/axes/shared.py
cvanelteren: matplotlib does not have this function but separate functions for x and y. I think due to this discrepancy it is better to make it internal.

ultraplot/tests/test_subplots.py
beckermr: multiple lines

PR #21, opened 2025-01-16
Update workflow to test for multiple python versions.

cvanelteren:
Addresses https://github.com/Ultraplot/ultraplot/pull/20#discussion_r1918927419

Perhaps the current supported versions is a bit overkill. I intended it to read from pyproject.toml to test the versions that are explicitly supported. I usually run whatever the latest is on conda-forge, but can imagine that some systems may lag in their upgrade schedules.


cvanelteren:
Closing to merge with #20

Issue #23, opened 2025-01-17
bug of ultraplot documentation


huddao:
Very happy the documentation was online, but there is a question. I'm not sure if this is a display problem or a code error.

[The page](https://ultraplot.readthedocs.io/en/latest/2dplots.html#Special-normalizers) 
![Image](https://github.com/user-attachments/assets/d11fc15e-7708-4216-a39e-efb9f89e6a2e)

cvanelteren:
Will have a look, thanks!

cvanelteren:
@huddao I believe I fixed the issue. It was a remnant from #13

cvanelteren:
The issue was a bit  tricky to narrow down as the tests create in #13  were passing. The issue only appeared after drawing the figure . Which puts some question marks on the testing pipeline right now. I will setup some baselines for the mpl images once I feel a bit more comfortable that the code is doing what I think it should be doing.

huddao:
thanks, I wish this can be resolved as soon as possible.

beckermr:
The relevant code was committed in PR #17, not #13.

cvanelteren:
Fixed: https://app.readthedocs.org/projects/ultraplot/builds/27028525/

cvanelteren:
Note it is fixed on latest but main will slowly come back online as it is built against an older commit.

PR #24, opened 2025-01-17
Hotfix contour

cvanelteren:
Addresses #23  partly reverts #13


beckermr:
The PR this reverts is #17, not #13 AFAICT.

beckermr (changes requested):
The PR that added this bit of code is #17. 

The diff there removed this block of code:

```
                for contour in obj.collections:
                    contour.set_linestyle("-")
                    contour.set_linewidth(linewidth)
                    contour.set_edgecolor("face")
```

Presumably one needs something to set how the contours are drawn. I don't see how simply removing the offending code is actually fixing anything.

cvanelteren:
From my understanding, the `collections` property used to hold the lines for the different levels. This property is removed and you can now access the lines directly using `get_paths` which returns  flat list of all the paths for all the levels. If I look at the api ([here](https://github.com/matplotlib/matplotlib/blob/8d64f03a1f501ba0019279bf2f8db3930d1fe33f/lib/matplotlib/contour.py#L1273C3-L1281C39)), the draw command sets the properties for the paths manually which makes the `collections` part redundant. Why? well my initially impulse would be to replace the collections with a similar command for the paths, but the api is already doing that on draw -- hence the removal.

cvanelteren (commented):
See main thread

beckermr:
OK. Let's add a test against an older version of the code to ensure a plot that accesses these features comes out the same.

cvanelteren:
I went back to proplot and copied the tests -- for which many won't work -- to recreate the contour which generated:

image

I think, however, that this is wrong. The reason why I think this is that the contour shows high and lows, but  the linestyles do not reflect these differences. The current unittests will produce 
image
I think the major difference is created through the path effects doing something funky to the ContourSet. I can recreate the left subplot by merely setting the correct keywords in the function call of the contour on `ax[1]` to:

```
ax.contour(data, ls = '-', lw = 1.5, edgecolor = 'k')
```
producing

image


 
Digging a bit further, I went to the proplot docs to look for some tests. From the proplot docs I can generate the following figure
image
<details> <summary> using this </summary>

```
import xarray as xr
import numpy as np
import pandas as pd

# DataArray
linspace = np.linspace(0, np.pi, 20)
data = 50 * state.normal(1, 0.2, size=(20, 20)) * (
    np.sin(linspace * 2) ** 2
    * np.cos(linspace + np.pi / 2)[:, None] ** 2
)
lat = xr.DataArray(
    np.linspace(-90, 90, 20),
    dims=('lat',),
    attrs={'units': '\N{DEGREE SIGN}N'}
)
plev = xr.DataArray(
    np.linspace(1000, 0, 20),
    dims=('plev',),
    attrs={'long_name': 'pressure', 'units': 'hPa'}
)
da = xr.DataArray(
    data,
    name='u',
    dims=('plev', 'lat'),
    coords={'plev': plev, 'lat': lat},
    attrs={'long_name': 'zonal wind', 'units': 'm/s'}
)

# DataFrame
data = state.rand(12, 20)
df = pd.DataFrame(
    (data - 0.4).cumsum(axis=0).cumsum(axis=1)[::1, ::-1],
    index=pd.date_range('2000-01', '2000-12', freq='MS')
)
df.name = 'temperature (\N{DEGREE SIGN}C)'
df.index.name = 'date'
df.columns.name = 'variable (units)'
import ultraplot as pplt
fig = pplt.figure(refwidth=2.5, share=False, suptitle='Automatic subplot formatting')

# Plot DataArray
cmap = pplt.Colormap('PuBu', left=0.05)
ax = fig.subplot(121, yreverse=True)
ax.contourf(da, cmap=cmap, colorbar='t', lw=0.7, ec='k')

# Plot DataFrame
ax = fig.subplot(122, yreverse=True)
ax.contourf(df, cmap='YlOrRd', colorbar='t', lw=0.7, ec='k')
ax.format(xtickminor=False, yformatter='%b', ytickminor=False)
pplt.show(block = 1)
```

</details>

Which "looks" the same as on [their docs](https://proplot.readthedocs.io/en/latest/2dplots.html).

# Conclusion:

I think the current implementation correctly implements the contour that matplotlib intends, and prior versions of proplot do not.

I propose to set the baseline for this particular plot at the current version rather to check it further back. But I am open if someone has the time or gusto to check it with even older versions.


## Subnote
I used the latest official build from proplot, with matplotlib 3.4.3  and manually rebuilding basemap -- I can share the env if one needs it.

beckermr:
I don't follow your message at all. In the plot from the current unittests, the right-hand subplot looks really funky. The black contour lines only show when there is a contour label. This is fixed by the extra call specified and the fixed plot looks like the output from proplot previously. 

I think we should include this call in the code so that the behavior is unchanged. Proplot deviates from matplotlib on purpose to produce better defaults and nicer plots. IDK why we would change that.

cvanelteren:
The unittest situation you mentioned - those were actually brought over from proplot since ultraplot didn't have any tests to begin with. Most of the carried over unittests that are carried over cannot fail as it needs a newer version of matplotlib. This makes it hard to include a test that generates baselines without retroactively fixing proplot -- which I don't feel like is necessary as I will show next.

I've thoroughly checked the contour plotting and can confirm it's working properly. The main difference you're seeing is just in how the default styles work:
- MPL normally doesn't show lines without labels
- Proplot uses the same line style everywhere
- My version uses dashed lines for negative values and solid for positive ones

The good news is you can still recreate those proplot-style images exactly like before. The changes are really just about giving users better default options out of the box.

Hopefully this clears the issues up!

beckermr:
Can you send me the code that made the second plot above? That would help me understand what is going on.

cvanelteren:
```python
import ultraplot as uplt

import numpy as np
data = state.rand(5, 5) * 10 - 5
fig, axs = uplt.subplots(ncols=2)
ax = axs[0]
ax.contourf(
    data,
    edgecolor="k",
    linewidth=1.5,
    labels=True,
    labels_kw={"color": "k", "size": "large"},
)
ax = axs[1]
m = ax.contourf(data, ls = '-', lw = 1.5, edgecolor = 'k')
ax.clabel(m, colors="black", fontsize="large")  # looks finewithout this

import matplotlib.patheffects as pe

m.set_path_effects([pe.Stroke(linewidth=1.5, foreground="k), pe.Normal()])
uplt.show(block = 1)

 ```

**Comment by cvanelteren** [comment]
Note it is from the unittest ;-)!

**Comment by beckermr** [comment]
Ah can you send a link to the unittest code?

**Comment by cvanelteren** [comment]
https://github.com/Ultraplot/ultraplot/blob/46616ab0bb5114b1444c3d68da3e6782bc2c7ed4/ultraplot/tests/test_2dplots.py#L106

**Comment by cvanelteren** [comment]
I am merging this with devel. After locally testing the open PRs it all looks good for now.

**Comment by cvanelteren** [comment]
is that ok @beckermr?

File: pyproject.toml [hunk 1]
```diff
@@ -19,14 +19,13 @@ maintainers = [
 ]
 description = "A succinct matplotlib wrapper for making beautiful, publication-quality graphics."
 readme = "README.rst"
-requires-python = ">=3.6.0"
+requires-python = ">=3.8, <3.12"
```
beckermr, line 22: Why the upper bound here?

beckermr (approved):
Sorry for the delay here. One question about the python bound but otherwise LGTM.

beckermr:
Feel free to merge to main if you'd like.

pyproject.toml
cvanelteren: basemap won't install for python>=3.12. Eventhough we don't have a hard requirement for it some of the tests will fail. I contacted the people from basemap and they are working on a fix. I put a crude PR up there that monkey patches basemap for these higher versions, but it isn't (and will likely never be) merged. 

I checked the code with this monkey patch for basemap and ultraplot will pass all the tests for these higher versions.

Issue #25, opened 2025-01-18
Inconsistent Cycling Behavior in ProPlot's Cycler Implementation


cvanelteren:
Currently, ProPlot's internal implementation of cycling functionality exhibits inconsistent behavior. While ProPlot uses a custom function to maintain compatibility with Matplotlib's cycler API, there appears to be a discrepancy in how cycling is executed.

Specifically:
1. The plotting functions are correctly implementing manual cycling
2. However, the underlying cycle object itself is not properly advancing through its sequence

This creates a mismatch between the expected cycling behavior and the actual output, potentially affecting any functionality that relies on proper cycler state management.

Steps to reproduce:
```python
import ultraplot as plt
import numpy as np
state = np.random.RandomState(51423)
fig, ax = plt.subplots()
cmap = plt.colormaps.get_cmap("538")
cycle = plt.Cycle(
    "538",
    marker=["X", "o", "s", "d"],
    sizes=[20, 100],
    edgecolors=["r", "k"]
)
ax.scatter(
    state.rand(10, 4),
    state.rand(10, 4),
    cycle=cycle,
    area_size=False,
)
plt.show(block = 1)
```
![Image](https://github.com/user-attachments/assets/beb054a0-38a0-43a8-8b1d-1b95dd46dc5b)
Expected behavior:
The cycle object should advance through its sequence in sync with the plotting functions.

Current behavior:
While plotting functions manually cycle through values, the cycle object remains static and doesn't progress through its sequence as expected.

cvanelteren:
Note here the colors are cycling but none of the other properties are.

cvanelteren:
![Image](https://github.com/user-attachments/assets/f8f3a83d-5fd2-49a3-8385-d71dc89875e0)

cvanelteren:
Got it working on the PR 😄

cvanelteren:
Fixed!

PR #26, opened 2025-01-18
Fix prop cycling not working

cvanelteren:
Addresses #25 

~~Note this is currently a draft as we are directly interfacing with private functions and this is not preferred.~~ But it does solve the issue. I will give it some thought on how this could be solved.


cvanelteren:
I would propose to change the cycle function to a cycle class and handle the cycling through there. This monkey patch here relies too much on internal mpl functions/variables and could be changed in the future.

cvanelteren:
Now the cycling is more robust, and passing the tests without checking with baseline compare. Will do it later.

cvanelteren:
Pinging @beckermr @pratiman-91

cvanelteren:
Could we go ahead and merge this? @pratiman-91 @beckermr

pratiman-91:
I am out on holiday until the first week of Feb. If someone else can review. 
Thanks!

ultraplot/axes/plot.py
```diff
@@ -3482,14 +3479,17 @@ def _apply_scatter(self, xs, ys, ss, cc, *, vert=True, **kwargs):
...
 
         guide_kw = _pop_params(kw, self._update_guide)
         objs = []
         for _, n, x, y, s, c, kw in self._iter_arg_cols(xs, ys, ss, cc, **kw):
             # Don't set 'c' explicitly unless it was provided
-            kw["s"] = s
-            if c is not None:
-                kw["c"] = c
+            kw["s"], kw["c"] = s, c
+            kw = self._parse_cycle(n, cycle_manually=cycle_manually, **kw)
```
beckermr, line 3492: This change doesn't match the comment. Why was it made? We need to either adjust the code or the comment. Also, why did you add another _parse_cycle call here?

ultraplot/axes/plot.py
```diff
@@ -3482,14 +3479,17 @@ def _apply_scatter(self, xs, ys, ss, cc, *, vert=True, **kwargs):
                 infer_rgb=infer_rgb,
                 **kw,
             )
+        # Move _parse_cycle before _parse_color
+        kw = self._parse_cycle(
+            xs.shape[1] if xs.ndim > 1 else 1, cycle_manually=cycle_manually, **kw
+        )
```
beckermr, line 3485: The comment mentions moving the cycler parsing before the parsing of the color. However, this change undoes that. Why? Should we delete the comment or do we need to move this back?

ultraplot/axes/plot.py
```diff
@@ -3459,12 +3459,9 @@ def _apply_scatter(self, xs, ys, ss, cc, *, vert=True, **kwargs):
         ys, kw = inputs._dist_reduce(ys, **kw)
         ss, kw = self._parse_markersize(ss, **kw)  # parse 's'
 
-        # Move _parse_cycle before _parse_color
-        kw = self._parse_cycle(xs.shape[1] if xs.ndim > 1 else 1, **kw)
-
         # Only parse color if explicitly provided
+        infer_rgb = True
         if cc is not None:
-            infer_rgb = True
```
beckermr, line 3467: This appears to be a change in defaults. Can you add a comment as to why or put it back?

ultraplot/constructor.py
```diff
@@ -762,148 +762,187 @@ def _pop_modification(key):
...
-        colors of the ``'538'`` color cycle.
-        For `~ultraplot.colors.ContinuousColormap`\\ s, this is either a
-        sequence of sample coordinates used to draw colors from the colormap, or
-        an integer number of colors to draw. If the latter, the sample coordinates
-        are ``np.linspace(0, 1, samples)``. For example, ``Cycle('Reds', 5)``
-        divides the ``'Reds'`` colormap into five evenly spaced colors.
+class Cycle(cycler.Cycler):
+    def __init__(self, *args, N=None, samples=None, name=None, **kwargs):
+        """
+        Generate and merge `~cycler.Cycler` instances in a variety of ways.
```
beckermr, line 768: The docs here no longer match the fact that this is a class as opposed to a function. We should adjust.

ultraplot/constructor.py
```diff
@@ -762,148 +762,187 @@ def _pop_modification(key):
...
+            Aliases for the above keywords.
+        **kwargs
+            If the input is not already a `~cycler.Cycler` instance, these are passed
+            to `Colormap` and used to build the `~ultraplot.colors.DiscreteColormap`
+            from which the cycler will draw its colors.
+
+        Returns
+        -------
+        cycler.Cycler
+            A `~cycler.Cycler` instance that can be passed
```
beckermr, line 837: This is no longer true. We now have a new class.

beckermr (changes requested):
Verifying that the Cycler object is correctly instantiated by our sublcass is difficult. I'd be good to add specific unit tests of this so that we verify it is correct and we can catch regressions. Ideally, we'd write the tests against a working version of the factory function first and then move the change to a class to a new PR and ensure the tests are still passing.

ultraplot/axes/plot.py
cvanelteren: This reverts the behavior of proplot.

ultraplot/axes/plot.py
cvanelteren: I moved this back t o how proplot does it. I initially thought I monkey patched in cycling behavior which broke on the original PR I had with proplot. The colors were cycling but the other properties were not. Not is does all.

ultraplot/constructor.py
cvanelteren: Good catch. I just copied the doc string over -- will fix.

cvanelteren:
> Verifying that the Cycler object is correctly instantiated by our sublcass is difficult. I'd be good to add specific unit tests of this so that we verify it is correct and we can catch regressions. Ideally, we'd write the tests against a working version of the factory function first and then move the change to a class to a new PR and ensure the tests are still passing.

Using #20, we can check for validity. That is, as soon as we have main producing a good baseline, any further tests can be used to check whether changes to the codebase will produce the same validity as we will expect (using pytest-mpl). Currently main does not pass these tests, but I am expecting that the joint of #26, #24 will.

ultraplot/axes/plot.py
cvanelteren: See comment below

cvanelteren:
I do like the idea of just testing if cycling happens correctly. I will write a simple test to see if it does that is lower level than the high level tests proposed above.

cvanelteren:
> I am out on holiday until the first week of Feb. If someone else can review. 
> Thanks!

Have fun😬

cvanelteren (commented):
Applied all fixed, added unittests, fixed up doc strings

ultraplot/constructor.py
```diff
@@ -762,148 +773,182 @@ def _pop_modification(key):
     return cmap
 
 
-def Cycle(*args, N=None, samples=None, name=None, **kwargs):
-    """
-    Generate and merge `~cycler.Cycler` instances in a variety of ways.
+class Cycle(cycler.Cycler):
+    def __init__(self, *args, N=None, samples=None, name=None, **kwargs):
```
beckermr, line 777: Doc strings for classes are always added to the class, not the __init__ function.

beckermr (changes requested):
One more item.

cvanelteren:
image
putting it back to draft -- I have been scratching my head why this is not working. Will keep this posted until I can fix this. The cycler properly generates a map from the keyword cycler but it is not properly passed on to pie.

cvanelteren:
Rebased to changes to clean up the mess -- should be working now. I cleaned up the logic inside `_parse_cycle` as I found it difficult to follow exactly what was returned where and when. This way I can more easily follow. As far as I can tell the cycling is working properly now. However, I noticed one crucial fact. UltraPlot performs a step at draw that may change the look of the images. This is not picked up when the image is passed to pytest-mpl.  I will run some further tests tomorrow and then this is done.

I'll put it on the todo to check what is causing this.

cvanelteren:
I am not very with how I cycle the properties here: there are calls to private properties and I cycle the object manually. After digging a bit, I can see that there is  currently no public api for accessing the cycling object (see https://github.com/matplotlib/matplotlib/issues/26831#issuecomment-1781552042). This leads us at a crossroad

1. We can either provide a public api
2. Or cycle it manually 

UltraPlot does override some public function that exist already (e.g. see `set_prop_cycle`).

I am also not sure if this goes beyond this PR -- intended as a quick fix -- rather than a larger api change. Open for suggestions.

cvanelteren:
Also see https://github.com/matplotlib/matplotlib/pull/29469
Which makes me lean towards merging it as is and waiting for potential future api to change such that we can follow suit.

cvanelteren:
Then we could add a test this future PR with

```python

def test_manual_cycle():
    cycle = uplt.Cycle(
        ["red", "green", "black"],
        marker=["X", "o"],
        sizes=[20, 100],
        edgecolors=["r", "k"],
    )
    fig, ax = uplt.subplots()
    ax.set_prop_cycle(cycle)
    ax.scatter([1], [1])

    ax = ax[0]  # force to get single axis rather than GridSpec
    parser = ax._get_lines
    current_pos = parser._idx
    cycled_pos = (current_pos + 1) % len(parser._cycler_items)
    props = ax._active_cycle.get_next()
    assert cycled_pos == parser._idx
    assert props == parser._cycler_items[parser._idx]
```

or something similar

cvanelteren:
I belief the root cause of the baseline images not showing the proper formatting when using pytest-mpl and they look fine when manually plotting is that in pytest-mpl they remove the `rcContex` prior to plotting: https://github.com/matplotlib/pytest-mpl/blob/29fd89095ced454e1f82007aeb592268a430aba0/pytest_mpl/plugin.py#L730

cvanelteren:
I am merging this with devel. After locally testing the open PRs it all looks good for now.

cvanelteren:
is that ok @beckermr?

ultraplot/axes/plot.py
```diff
@@ -2477,54 +2477,43 @@ def _parse_cycle(
...
+            self.set_prop_cycle(resolved_cycle)
+
+        # Apply manual cycle properties
+        if cycle_manually:
+            current_prop = self._get_lines._cycler_items[self._get_lines._idx]
+            self._get_lines._idx = (self._get_lines._idx + 1) % len(self._active_cycle)
+            for prop, key in cycle_manually.items():
+                if kwargs.get(key) is None and prop in current_prop:
+                    value = current_prop[prop]
+                    kwargs[key] = pcolors.to_rgba(value) if key == "c" else value
```
beckermr, line 2514: With the new changes, this block of code is skipped if `return_cycle` is `True`. Why is that?

ultraplot/constructor.py
```diff
@@ -845,65 +850,103 @@ def Cycle(*args, N=None, samples=None, name=None, **kwargs):
...
+        # Ensure at least a default color property exists
+        if not props:
+            props = {"color": ["black"]}
+
+        # Build cycler with matching property lengths
+        lengths = [len(value) for value in props.values()]
+        maxlen = np.lcm.reduce(lengths)
+        props = {key: value * (maxlen // len(value)) for key, value in props.items()}
+        mcycler = cycler.cycler(**props)
+        super().__init__(mcycler)
```
beckermr, line 939: Can you add a comment on what this call does exactly? Does it use matplotlib's init function to update the current cycle to the computed one?

ultraplot/constructor.py
```diff
@@ -845,65 +850,103 @@ def Cycle(*args, N=None, samples=None, name=None, **kwargs):
...
-            return
-        elif not np.iterable(value) or isinstance(value, str):
-            value = (value,)
-        props[key] = list(value)  # ensure mutable list
 
-    # If args is non-empty, means we want color cycle; otherwise is black
-    if not args:
-        props.setdefault("color", ["black"])
+    def __init__(self, *args, N=None, samples=None, name=None, **kwargs):
+        self.name = "_no_name"  # default value
```
beckermr, line 855: This is a change in behavior. In the previous code, the cycler always got the input name unless the function hit the early return. I'd use the `_not_none` here too.

ultraplot/axes/plot.py
beckermr, line 2514: This is to_hex previously. Why the change?

beckermr (changes requested):
This PR needs changes. Please wait next time.

cvanelteren:
Ah my bad!

ultraplot/constructor.py
cvanelteren: `cycler.cycler` is not a class but a function in mpl. The logic of the init from the `Cycle` class is not compatible with the function -- that is we need to pass the output of cycler directly to the parent of the class we introduced.

beckermr: Cool. Let's add a comment on this in the code.

cvanelteren:
I will open another PR to fix the fixes.

beckermr:
Thank you! Sorry for the delay in my code review.

ultraplot/axes/plot.py
cvanelteren: You are right! Shouldn't have happened. I believe I was testing something here and forgot to change it back. See #29

ultraplot/constructor.py
cvanelteren: Right again!

cvanelteren:
No worries ;-)

Issue #27, opened 2025-01-24
Error when trying to add inset axis


smups:
Info:
- python 3.13
- ultraplot 1.0
- matplotlib 3.10.0
- OS: fedora 41 (x86_64)

Code that breaks:
```python
fig = uplt.figure(figwidth=6.2, share=False, ref=4)
axs = fig.subplots([[1,2,3],[4,5,6],[4,5,6]])
fig.format(toplabels=['Model A', 'Model B', 'Model C']

m = axs[0,0].imshow(np.transpose(a), extent=extent_a)
axs[1,0].plot(*trajectory_a)

axs[1,1].imshow(np.transpose(b), extent=extent_b)
axs[0,1].plot(*trajectory_b)

# extent_c = (0, 35, 0, 4π ≈ 12)
axs[1,2].imshow(np.transpose(c), extent=extent_b)
axs[0,2].plot(*trajectory_c)

# (code works without this line)
ix = axs[1,2].inset([5, 5, 5, 5]) # <- error: `NotImplementedError: cannot remove artist`

fig.colorbar(m, loc='r', extend='max', formatter='sci')
```

Complete error:
```log
---------------------------------------------------------------------------
NotImplementedError                       Traceback (most recent call last)
File ~/.local/lib/python3.13/site-packages/matplotlib/pyplot.py:279, in _draw_all_if_interactive()
    277 def _draw_all_if_interactive() -> None:
    278     if matplotlib.is_interactive():
--> 279         draw_all()

File ~/.local/lib/python3.13/site-packages/matplotlib/_pylab_helpers.py:131, in Gcf.draw_all(cls, force)
    129 for manager in cls.get_all_fig_managers():
    130     if force or manager.canvas.figure.stale:
--> 131         manager.canvas.draw_idle()

File ~/.local/lib/python3.13/site-packages/matplotlib/backend_bases.py:1891, in FigureCanvasBase.draw_idle(self, *args, **kwargs)
   1889 if not self._is_idle_drawing:
   1890     with self._idle_draw_cntx():
-> 1891         self.draw(*args, **kwargs)

File ~/.local/lib/python3.13/site-packages/ultraplot/figure.py:467, in _add_canvas_preprocessor.<locals>._canvas_preprocess(self, *args, **kwargs)
    465 ctx3 = rc.context(fig._render_context)  # draw with figure-specific setting
    466 with ctx1, ctx2, ctx3:
--> 467     fig.auto_layout()
[...]

PR #29, opened 2025-01-26
Cycling hotfix2

cvanelteren:
I merge too quickly with the other PR -- these address the points raided by  @beckermr

### DIFF
--- a/ultraplot/axes/plot.py
+++ b/ultraplot/axes/plot.py
@@ -2491,10 +2491,6 @@ class PlotAxes(base.Axes):
         # Ignore cycle for single-column plotting
         resolved_cycle = None if ncycle == 1 else resolved_cycle
 
-        # Return or apply cycle
-        if return_cycle:
-            return resolved_cycle, kwargs
-
         if resolved_cycle and resolved_cycle != self._active_cycle:
             self.set_prop_cycle(resolved_cycle)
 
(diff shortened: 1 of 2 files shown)


cvanelteren:
Not entirely sure why the tests are failing -- they look fine here. The tests inside `test_axes` are acting up, related to #27

cvanelteren:
Ah I see what the issue is; the pytest-mpl stuff needs to be removed from the workflow.

beckermr:
It may be worth vendoring our own pytest plugin or making our own decorator. We'll be heavy users.

cvanelteren:
> It may be worth vendoring our own pytest plugin or making our own decorator. We'll be heavy users.

Good idea! Was working already working on this and currently have an issue tracker on pytest-mpl.

cvanelteren:
I figure out that pytest-mpl does in fact allow you to overwrite the style. I have done some local tests and it seems to be working nicely with (in nushell): 

```nushell
pytest --mpl-generate-path=ultraplot/tests/baseline --mpl-default-style (python -c "import ultraplot as plt; print(plt.config.Configurator())")

```

beckermr:
Oh nice! Is there a way to set those arguments as a default in our pyproject.toml?

cvanelteren:
Yes will add it in another PR. I do expect that some stuff will most likely break still. I manually inspected the images a few times and some of them still look a bit odd to me but the very fast majority looks great!

beckermr (approved):

beckermr:
OK, sounds good to me. Let's go ahead with this one despite the tests failing.

cvanelteren:
The failure is related to the fix #33. Should be fine.

PR #33, opened 2025-01-27
Compatiblity update for mpl 3.10

cvanelteren:
Addresses #27

# Reason
Updates `indicate_inset_zoom` to support matplotlib 3.10+ while maintaining backwards compatibility.

## Changes
- Adds version-specific implementation for matplotlib ≥3.10 using the new `InsetIndicator` class
- Maintains existing implementation for matplotlib <3.10 using rectangle and patches
- Both implementations share the same interface and return format

## Technical Details
- For mpl ≥3.10: Uses `InsetIndicator` class which manages both rectangle and connectors in a single object
- For mpl <3.10: Continues to use separate rectangle patch and connector lines
- Implements version checking to automatically select the appropriate implementation

### DIFF
--- a/ultraplot/axes/base.py
+++ b/ultraplot/axes/base.py
@@ -2914,7 +2914,6 @@ class Axes(maxes.Axes):
 
             from matplotlib.inset import InsetIndicator
 
-            kwargs.setdefault("facecolor", "none")
             indicator = InsetIndicator(
                 bounds=bounds, inset_ax=self, **kwargs)
 
--- a/ultraplot/axes/base.py
+++ b/ultraplot/axes/base.py
@@ -2914,10 +2914,9 @@ class Axes(maxes.Axes):
 
             from matplotlib.inset import InsetIndicator
 
-            # NOTE: Explicitly set facecolor here to none, as it seems to be ignore -- not sure why as it is the default
+            kwargs.setdefault("facecolor", "none")
             indicator = InsetIndicator(
-                bounds=bounds, inset_ax=self, facecolor="none", **kwargs
-            )
+                bounds=bounds, inset_ax=self, **kwargs)
 
             if self._inset_zoom_artists:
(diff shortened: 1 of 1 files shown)


beckermr (approved):
I'm happy to merge this. We should develop a specific policy on MPL version support.

cvanelteren:
I can also leave the warning out for now. It is more for the maintainers to remind to rid of some legacy code. We could also leave it out and naturally drop support when the return of this function will break. The return signature still uses the tuple format -- which is compatible with the Indicator class but may change in the future. The mpl docs list the indicator class as an experimental feature. Curious to hear what you think.

beckermr:
I think we should drop support for python versions and MPL versions on the same schedule as MPL. 

I also think we should move to calver or [effver](https://jacobtomlinson.dev/effver/). 

Semver is just too hard for this package due to how much churn we experience on the MPL side of things. 

Using experimental features / apis is fine imho. We can read the MPL changelogs to find out when they change or are dropped.

beckermr (approved):

cvanelteren:
Yeah let's do that. Effver seems to work fine -- usually not too fan of the date releases. The question now is do we mark the recent changes as a 1.1 or a 1.0.1?

cvanelteren:
Regarding
>I think we should drop support for python versions and MPL versions on the same schedule as MPL.

I fully agree too. However with regard to the feature itself do you suggest making the mpl compatibility more explicit in pyproject.toml? For example if you install mpl 3.10 you need ultraplot version x as well. 

Looking at the codebase now I sometimes look at how cases are handled and am confused if the situation is still recent or we just delete the older supported mpl versions. Not sure what the best way about going it. For now we could leave it in with a comment and if it becomes a problem we just remove it.

beckermr:
Yes we should ensure we have proper upper bounds on mpl in the pyproject.toml.

PR #40, opened 2025-01-28
prod: drop python 3.9 per MPL and NEP29

beckermr:
We need to drop 3.9 support for two reasons:

1. The use of the `match cycle` syntax means we are no longer compatible with 3.9
2. matplotlib follows numpy's version support in NEP29. Per [their docs](https://numpy.org/neps/nep-0029-deprecation_policy.html#support-table), support for python 3.9 was dropped for any release after December 18, 2024.


environment-dev.yml
```diff
@@ -2,8 +2,8 @@ name: ultraplot-dev
 channels:
   - conda-forge
 dependencies:
-  - python>=3.9,<3.12
-  - numpy>=2.2
+  - python>=3.9,<3.14
```
beckermr, line 5:
```suggestion
  - python>=3.10,<3.14
```

environment-dev.yml
```diff
@@ -2,8 +2,8 @@ name: ultraplot-dev
...
+  - numpy
   - matplotlib>=3.9.3
```
beckermr, line 7:
```suggestion
  - matplotlib>=3.9
```

cvanelteren:
I anticipate potential challenges regarding our Basemap integration (reference: https://github.com/matplotlib/basemap/blob/develop/packages/basemap/pyproject.toml).

While Basemap wasn't explicitly listed as a dependency in Proplot, it effectively functions as one in practice. We should make a deliberate decision to either formally include it or remove the functionality entirely. The Basemap project is currently undergoing a revival after several years of deprecation, with ongoing efforts to modernize it. I've submitted a pull request to help catalyze this process (https://github.com/matplotlib/basemap/actions/runs/12949899747/job/36122328672?pr=611), though I must be mindful of maintaining focus across multiple projects.

I agree with the previous points raised, and maintaining version parity with Matplotlib would be  smart!

environment-dev.yml
cvanelteren: This would be an issue with basemap having no support for numpy < 2. We need to address that first before we add support higher than 3.12.

environment-dev.yml
```diff
@@ -13,13 +13,13 @@ dependencies:
   - pytest-mpl
   - jupyter
   - pip
+  - basemap
```
beckermr, line 16:
```suggestion
```

environment-dev.yml
```diff
@@ -13,13 +13,13 @@ dependencies:
...
+  - sphinx-automodapi
   - pip:
-      - basemap
-      - pint
-      - sphinx
-      - jupytext
-      - nbsphinx
-      - sphinx-copybutton
-      - sphinx-autoapi
       - git+https://github.com/ultraplot/UltraTheme.git
```
beckermr, line 25:
```suggestion
      - git+https://github.com/ultraplot/UltraTheme.git
      - basemap
```

cvanelteren:
We should perhaps remove `docs/environment.yml`  as it is currently causing the readthedocs to fail due to a mismatch of the python versions.

environment-dev.yml
```diff
@@ -13,13 +13,14 @@ dependencies:
...
-      - sphinx-automodapi
+      - basemap
```
beckermr, line 26:
```suggestion
```

environment-dev.yml
```diff
@@ -13,13 +13,14 @@ dependencies:
...
   - jupyter
   - pip
+  - pint
+  - sphinx
+  - nbsphinx
+  - jupytext
+  - sphinx-copybutton
+  - sphinx-autoapi
+  - sphinx-automodapi
+  - sphinx-rtd-theme
```
beckermr, line 23:
```suggestion
  - sphinx-rtd-theme
  - base
```

environment-dev.yml
```diff
@@ -13,13 +13,14 @@ dependencies:
...
+  - base
```
beckermr, line 24:
```suggestion
  - basemap
```

beckermr:
This is blocked on a release of basemap 1.4.1 from conda-forge. See https://github.com/conda-forge/basemap-feedstock/pull/137

environment-dev.yml
```diff
@@ -2,9 +2,9 @@ name: ultraplot-dev
 channels:
   - conda-forge
 dependencies:
-  - python>=3.9,<3.12
-  - numpy>=2.2
-  - matplotlib>=3.9.3
+  - python>=3.10,<3.14
```
beckermr, line 5:
```suggestion
  - python>=3.10,<3.13
```

pyproject.toml
```diff
@@ -19,7 +19,7 @@ maintainers = [
 ]
 description = "A succinct matplotlib wrapper for making beautiful, publication-quality graphics."
 readme = "README.rst"
-requires-python = ">=3.9, <3.12"
+requires-python = ">=3.10,<3.14"
```
beckermr, line 22:
```suggestion
requires-python = ">=3.10,<3.13"
```

cvanelteren (commented):
The pin below 3.13 works for the available basemap? Otherwise looks good.

beckermr:
yes that pin works for the available basemap packages.

PR #45, opened 2025-01-30
Add baseline comparison to workflow

cvanelteren:
Adding the baseline tests back into ensure future fidelity and make the tests more robust. This would make it easier to track potentially hard to see errors. It would further document future changes in case we wish to deviate from a style in the future.


cvanelteren:
There is a failure on main (apparently) that is fixed in this PR; initially thought the issue was on this one. I can also separate that and then retest this PR. Let me know how you would like to proceed @beckermr

beckermr (approved):

PR #46, opened 2025-01-30
Revert "Add baseline comparison to workflow"

cvanelteren:
Reverts Ultraplot/ultraplot#45


beckermr (changes requested):
Why is this being reverted?

cvanelteren:
It needs to be amended there is a folder error in the second part

beckermr:
OK. How about submit a bug fix instead of just backing the whole thing out?

PR #48, opened 2025-01-30
Readthedocs fix

cvanelteren:
Read the docs are failing. Needed to add install to ensure the imports work in the doctests

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -6,7 +6,7 @@ version: 2
 build:
   os: ubuntu-22.04
   tools:
-    python: mambaforge-latest
+    python: miniforge-latest
 conda:
   environment: ./environment.yml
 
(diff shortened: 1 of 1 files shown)


.readthedocs.yml
```diff
@@ -7,19 +7,21 @@ build:
   os: ubuntu-22.04
   tools:
     python: mambaforge-latest
```
beckermr, line 9:
```suggestion
    python: miniforge-latest
```

mambaforge is deprecated and miniforge is identical to it now. I think this will work.

beckermr:
I can't see the read the docs admin panel. Assuming this build works there, Lgtm! 

Is there a way for rtd to post the status of the build to github?

cvanelteren:
Added you to the admin panel 😬

cvanelteren:
> Is there a way for rtd to post the status of the build to github? 

We have a badge that does this (and currently reads failing). Other than the badge, not sure if we can report it directly somehow.

PR #57, opened 2025-01-31
Dep build

### LINKED ISSUE #56: Publish to pypi still being skipped
Not sure why this is being skipped but it doesn't trigger well. I also think we can let go of release and just focus on tags?

cvanelteren:
This fixes #56 as it was creating a dependency between test and prod whereas test was trigger on tags whereas prod is triggered on release. Furthermore this integrates the two rather than having tags and releases

### DIFF
--- a/.github/workflows/build-ultraplot.yml
+++ b/.github/workflows/build-ultraplot.yml
@@ -81,3 +81,10 @@ jobs:
         with:
           name: failed-comparisons-${{ inputs.python-version }}-${{ inputs.locale }}
           path: results/*
+
+  build-success:
+    needs: [build-ultraplot, compare-baseline]
+    runs-on: ubuntu-latest
+    steps:
+      - name: Build succeeded
+        run: echo "All builds and tests passed successfully!"
--- a/.github/workflows/build-ultraplot.yml
+++ b/.github/workflows/build-ultraplot.yml
@@ -87,10 +87,3 @@ jobs:
         with:
           name: failed-comparisons-${{ inputs.python-version }}-${{ inputs.matplotlib-version }}
           path: results/*
-
-  build-success:
-    needs: [build-ultraplot, compare-baseline]
-    runs-on: ubuntu-latest
-    steps:
-      - name: Build succeeded
-        run: echo "All builds and tests passed successfully!"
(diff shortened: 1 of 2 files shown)


cvanelteren:
What do you think of letting go of release and just using tags @beckermr ?

beckermr:
I am confused. Can you explain the issue more clearly?

cvanelteren:
Currently prod is pushed only when testpy is finished, but they are both run on different conditions. Test is run on tag releases whereas prod is run on releases. So it ends up being mutually exclusive.

cvanelteren:
To continue. We currently have releases and tags, but one of them is sufficient I think. We can let one go and still keep the dep between uploading to test (catching potential issues) and releasing to pypi.

beckermr (changes requested):
We can use this for the if condition to solve this problem


```
if: github.event_name == 'push' || github.event_name == 'release'
```

beckermr:
We should keep the official releases on github.

.github/workflows/publish-pypi.yml
```diff
@@ -1,7 +1,7 @@
 name: Publish to PyPI
 on:
   release:
-    types: [published]
+    tags: ["v*"]
```
beckermr, line 4:
```suggestion
    tags: [published]
```

.github/workflows/publish-pypi.yml
```diff
@@ -1,7 +1,7 @@
 name: Publish to PyPI
 on:
   release:
-    types: [published]
+    tags: [published]
```
beckermr, line 4:
```suggestion
    types: [published]
```

beckermr (approved):

beckermr (changes requested):
the build-sucess step is in the wrong file.

beckermr:
We should back out that latest change and just merge my other PR.

.github/workflows/build-ultraplot.yml
```diff
@@ -87,3 +87,10 @@ jobs:
         with:
           name: failed-comparisons-${{ inputs.python-version }}-${{ inputs.matplotlib-version }}
           path: results/*
+
+  build-success:
+    needs: [build-ultraplot, compare-baseline]
+    runs-on: ubuntu-latest
+    steps:
+      - name: Build succeeded
+        run: echo "All builds and tests passed successfully!"
```
beckermr, line 96:
```suggestion
```

beckermr (approved):

PR #61, opened 2025-02-01
fixing publish workflow

### LINKED ISSUE #60: Publish work does not work
I am a bit at a loss why the workflow is not running properly. Will look at this again but now putting this up as a reminder.

cvanelteren:
fixes #60

The workflow indicates that pypi does not allow for the custom classifier for matplotlib framework (https://pypi.org/classifiers/). I moved them to a custom header and updated the workflow appropriately.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -62,21 +62,7 @@ jobs:
           min_version = re.search(r">=(\d+\.\d+)", mpl_req)
           max_version = re.search(r"<(\d+\.\d+)", mpl_req)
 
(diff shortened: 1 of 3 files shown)


beckermr (changes requested):
The MPL version code does not rely on the classifiers. It uses the version range for the MPL dep listing directly. So we do not need the extra pyproject.toml section and we can put back the code you deleted.

beckermr:
@cvanelteren see this PR: https://github.com/Ultraplot/ultraplot/pull/62

PR #63, opened 2025-02-01
fix: correct chain of logic for publish workflow

beckermr:
This PR fixes the logic in the publish workflow. It uses test pypi for releases and tag pushes, but only pushes releases to pypi.

### DIFF
--- a/.github/workflows/publish-pypi.yml
+++ b/.github/workflows/publish-pypi.yml
@@ -30,6 +30,7 @@ jobs:
         uses: pypa/gh-action-pypi-publish@release/v1
         with:
           repository-url: https://test.pypi.org/legacy/
+          verbose: true
 
   publish-prod:
     name: Publish to PyPI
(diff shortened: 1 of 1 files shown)


.github/workflows/publish-pypi.yml
```diff
@@ -39,7 +38,7 @@ jobs:
       name: prod
       url: https://pypi.org/project/ultraplot/
     runs-on: ubuntu-latest
-    if: github.event_name == 'release' || (github.event_name == 'push' && needs.publish-pypi-test.result == 'success')
```
cvanelteren, line 42: Not sure if we  can remove that last part, but otherwise LGTM.

cvanelteren (changes requested):
We should also add  the verbose tag in case it does break in the future

.github/workflows/publish-pypi.yml
beckermr: We can since the `needs` block will only run the second job if the first suceeds.

cvanelteren (approved):

PR #64, opened 2025-02-02
fix: get pypi publish working

beckermr:
I am copying over the configs from conda-forge-metadata which has a working publish action: https://github.com/conda-forge/conda-forge-metadata/blob/main/.github/workflows/pypi.yml

Hopefully this finally fixes things. 

Right now, setuptools scm is not computing the version from the tag correctly.


beckermr:
OK @cvanelteren - let's try this version

cvanelteren (approved):

PR #65, opened 2025-02-02
add case where cycler is already a cycle

cvanelteren:
Ran into this edge when trying to fix the deprecation warnings


beckermr (changes requested):
Needs a test.

cvanelteren:
The proplot docs have 
image
We get now:

image

What kind of test are you thinking off? We already have some tests currently.

beckermr:
I'd write a direct unittest of the class itself at minimum.

cvanelteren:
It fails on some tests here locally, but the failures reflect how it should look.

cvanelteren:
In particular: 
image
image
image

PR #69, opened 2025-02-03
Deprecation fix mpl 3.10 and beyond

cvanelteren:
There are a lot of deprecation warnings from MPL and or Ultraplot existing in the current build. The majority of these fixes can be addressed by modifying the tests. Some of the deprecation warnings ensures keywords of usage to be compatible with mpl 3.10 which will result in errors in later versions. 

There are a few errors I have not addressed that warrants some discussion.

- <details>
    <summary> The first one </summary>  
      
    ```sh
    ultraplot/tests/test_1dplots.py::test_boxplot_vectors
    /home/casper/micromamba/envs/ultraplot-dev/lib/python3.11/site-packages/pytest_mpl/plugin.py:125: UltraPlotWarning: Passing lists to fillalpha was removed in v0.9. Please specify different opacities using the property cycle colors instead.
        store.return_value[test_name] = obj(*args, **kwargs)
    ```
  </details>
- <details>
    <summary> The second one </summary>  
      
    ```sh
    ultraplot/tests/test_subplots.py::test_axis_sharing[labels]
  /home/casper/micromamba/envs/ultraplot-dev/lib/python3.11/site-packages/pytest_mpl/plugin.py:125: UltraPlotWarning: Calling arbitrary axes methods from SubplotGrid was deprecated in v0.8 and will be removed in a future release. Please index the grid or loop over the grid instead.
    store.return_value[test_name] = obj(*args, **kwargs)
    ```
  </details>

For both of these, I don't agree with the deprecation warning. The first issue indicates that we should move towards a cycler object. However, adjusting the subplot to match the target baseline cannot be readily achieved through the cycler object. It requires setting the patches to a different alpha and leaving the lines with alpha = 1. This would (I think) require an extension of what the cycler object can handle which is to set all potential properties of all the artists. I don't think it can currently do this.

The second error implies the removal of multi-axis formatting. Personally, I would be against removing this feature as it is one of the most powerful features of UltraPlot.


cvanelteren:
Making this a draft as we have not updated to 3.10 yet.

cvanelteren:
Note I had to rename the warning UltraplotWarning to UltraPlotWarning due to some weird pytest-xdist issue.

cvanelteren:
@beckermr 

I had some spare time so pushed this along. There is one warning I could use your input for. There is a warning for the boxplot that keyword `fillalphas` is removed in 0.9 for proplot. I cannot determined directly why this was deprecated from the source but I have a hunch.

The warning implies that properties have to be fed through the `Cycle` object. However, the source is currently geared towards dealing with line properties only, setting color, alpha and marker style accordingly. This creates a stress point as the deprecation warning is not compatible with the current implementation of the source.  

The major problem is that it is not trivial to determine from a singular entry point how to cycle general objects. Line objects have different properties from boxplots, and contours. As such, setting all the properties through a general cycler won't be possible unless we write the code for this. There are some properties that can be leveraged for this in `ultraplot.internals.__init__.py` that categorize the properties for the major artist types.

In short, we either 
a) allow for the `fillalphas` to remain
b) extend the cycler object to allow cycling for different parts of the plot. For example we could have an internal map to `line, patch, fliers, medians` etc, and cycle the properties accordingly.

For `b` however, retooling would be necessary that will touch much of the codebase and I a don't know what will be gained tbh. I could be missing  the point of the warnings and Luke's intentions, hence this msg to get some input on this issue. Happy to hear your thoughts.

P.S. This warning can be found by running `ultraplot/tests/test_1dplots.py::test_boxplot_vectors` (for example).

ultraplot/axes/plot.py
```diff
@@ -3892,6 +3892,7 @@ def _apply_boxplot(
             x, y, autoy=False, autoguide=False, vert=vert, **kw
         )
         kw = self._parse_cycle(x.size, **kw)  # possibly apply cycle
+        print(kw)
```
beckermr, line 3895: This needs to be removed before merging.

ultraplot/figure.py
```diff
@@ -816,14 +816,9 @@ def _context_authorized(self):
...
-        Handle deprication of basemap and cartopy package.
+        Handle potential deprication of basemap and cartopy package.
         """
-        if basemap is not None:
-            backend = ("cartopy", "basemap")[bool(basemap)]
-            warnings._warn_ultraplot(
-                f"The 'basemap' keyword was deprecated in version 0.10.0 and will be "
-                f"removed in a future release. Please use backend={backend!r} instead."
-            )
+        # Basemap is currently being developed again so are removing the deprecation warning
```
beckermr, line 821: This is not right. This deprecation warning was about how to specify the backend (via a string) and does not depend on whether or not basemap is being developed. We need to keep this code and force folks to write `backend='basemap'` if they want basemap.

ultraplot/tests/test_1dplots.py
```diff
@@ -150,14 +150,15 @@ def test_boxplot_vectors():
     datas = np.array(datas, dtype=object)
     assert len(datas) == len(coords)
     fig, ax = uplt.subplot(refwidth=3)
+    cycle = uplt.Cycle("538")
```
beckermr, line 153: Is there a reason the cycle is made directly here?

beckermr (changes requested):
I left some comments.

ultraplot/figure.py
cvanelteren: The backend is not getting removed imo. At the time Luke developed this basemap was EOL.

beckermr:
Yes I know, sorry. I don't fully understand why things were deprecated either. I think not deprecating it fow now is fine.

cvanelteren:
Should be good. I think the failed tests are wrong currently in main. Why? The tests
- `test_drawing_in_projection_without_globe`
- `test_drawing_in_projection_with_globe` 
Are matched horizontally and vertically in this PR but not from main. The data within each test is the same just what is plotted is different which would mean that we should see a similar pattern horizontally but not vertically. This is the case for the PR but not for main. 
The other tests that fail were not included (e.g. `test_panels_suplabels_three_hor_panels_donotinlcude`). The one that is odd is `test_multi_formatting` which I believed is caused by double pcolor line  that exist. 

This PR ended up combining too many things in one swoop -- which is not preferred, but  hopefully is still clear. The codebase should be ready for 3.10 now.

cvanelteren:
@beckermr could you review these changes. Codecov is indicating missing tests for the patch, but I wan to address these in a larger focus. I will just take 1 file a week and add missing tests. For example, I changed the `UltraplotWarning` to `UltraPlotWarning` and now codecov indicates that these should be tested.

I am in favor of pushing this through and fixing the missing coverage on a weekly basis to focus my efforts more constructively (as it has to happen anyway).

cvanelteren:
Tests look as they should and fail on newly added once or correct existing errors.

cvanelteren:
Can we go ahead an merge this @beckermr?

ultraplot/axes/base.py
```diff
@@ -1583,6 +1591,64 @@ def _artist_fully_clipped(self, artist):
...
+        bounds: tuple[float, float, float, float],
+        parent: "Axes",
+        **kwargs,
+    ) -> "InsetIndicator":
+        # Implementation for matplotlib >= 3.10
+        # NOTE: if the api changes we need to deprecate the old one. At the time of writing the IndicateInset is experimental and may change in the future. This would require us to change potentially the return signature of this function.
+        kwargs.setdefault("label", "_indicate_inset")
+
+        # If we already have a zoom indicator we need to update the properties or add them
+        # Note the first time we enter this function, we create the object. Afterwards the function is accessed again but with different updates
```
beckermr, line 1616: Let's breakup the long comments across multiple lines.

ultraplot/axes/base.py
```diff
@@ -2223,6 +2289,8 @@ def _legend_tuple(*objs):  # noqa: E306
             else:
                 hs = hs[0] if len(hs) == 1 else hs  # unfurl for better error messages
                 label = label if label is not None else labs[0] if labs else "_no_label"
+                if label.startswith("_"):
```
beckermr, line 2292: Is this a bug fix?

beckermr (changes requested):
I had a few more comments, plus there are test failures in the plots that do not look like we expect them.

beckermr:
Also the docs build is failing.

cvanelteren:
Will fix this and perhaps split this PR up for clarity.

ultraplot/axes/base.py
cvanelteren: My bad it is my editor that autowraps it at 65 :)

ultraplot/axes/base.py
cvanelteren: Ah I see broke it. Thanks!

ultraplot/axes/base.py
```diff
@@ -2273,6 +2339,8 @@ def _legend_tuple(*objs):  # noqa: E306
             else:
                 hs = hs[0] if len(hs) == 1 else hs  # unfurl for better error messages
                 label = label if label is not None else labs[0] if labs else "_no_label"
+                if label and str(label).startswith("_"):
+                    continue
```
beckermr, line 2343: @cvanelteren This change is not a simple refactor for clarity. The continue statement means that labels that start with `_` are never added to the list. What happens if a user passes a label like `_my_custom_label`?

beckermr (changes requested):
This continue statement needs more discussion.

ultraplot/axes/base.py
cvanelteren: This is how it was  before:
https://github.com/Ultraplot/ultraplot/blob/eb7fef59bfa1fc43a4cba61eb55b10f9977b2178/ultraplot/axes/base.py#L2214

cvanelteren: ah oops listing the wrong thing. I will revert this

cvanelteren: For context, I think this was there due to the internal maps having weird `_` starting labels. But can't recall exactly why I wrote this. It has been a little while :)

beckermr (changes requested):
Code looks good, though I do not understand all of the changes.

The test failures need fixing, and then I think we can merge.

cvanelteren:
Ok merging this as the newer plots more correct. We may need to revert the draw in `fig.show` as it caused the image sizes to differ.

beckermr:
I was about to comment that the newer plots were more correct as well. 

If you want code review, please wait until the reviewer approves before merging.

cvanelteren:
Ah ok I gathered it was ok. Will wait next time.

PR #76, opened 2025-02-14
use norm explicitly

### LINKED ISSUE #75: norm ignored when (when clipping) for coloring scatter plots; normalizer is modified
when passing a norm to scatter / pcolor, the norm is ignored for coloring but also its vmin/vmax values are modified

```python
import numpy as np
import ultraplot as uplt

x = np.arange(10)
y = x**2
c = y

cmap = uplt.Colormap("viridis")
norm = uplt.Norm("linear", 0, 10)
fig, ax = uplt.subplots()
ax.scatter(x, y, c=c, cmap=cmap, norm=norm)
print(norm.vmin, norm.vmax)


fig, ax = plt.subplots()
norm = uplt.Norm("linear", 0, 10)
ax.scatter(x, y, c=c, cmap=cmap, norm=norm)
```
```
>>> 0.0 81.0
```

ultraplot result

Image

matplotlib result

Image

See also the discussion on the proplot github: https://github.com/proplot-dev/proplot/issues/400

A workaround is to pass vmin, vmax directly. 
norm not correctly applying colors I can live with but the modification of the norm object is unexpected thus dangerous

cvanelteren:
Should fix #75

### DIFF
--- a/ultraplot/axes/plot.py
+++ b/ultraplot/axes/plot.py
@@ -2296,7 +2296,8 @@ class PlotAxes(base.Axes):
         cmap_kw = cmap_kw or {}
         norm_kw = norm_kw or {}
         # If norm is given we use it to set vmin and vmax
-        if norm:
+
+        if isinstance(norm, mcolors.Normalize):
             vmin = norm.vmin
             vmax = norm.vmax
         vmin = _not_none(vmin=vmin, norm_kw_vmin=norm_kw.pop("vmin", None))
--- a/ultraplot/axes/plot.py
+++ b/ultraplot/axes/plot.py
@@ -2444,7 +2444,6 @@ class PlotAxes(base.Axes):
             kwargs.update({"levels": levels, "extend": extend})
         else:
             guides._add_guide_kw("colorbar", kwargs, extend=extend)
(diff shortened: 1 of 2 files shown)


beckermr (changes requested):
For this one, we should compare the output figures directly as well.

cvanelteren:
As in make the test an image compare?

beckermr:
Yes. It will be a no-op in this case, but will catch any regressions in the future.

cvanelteren:
The test will fail due to a newly added test (probably).

ultraplot/axes/plot.py
```diff
@@ -2296,6 +2296,9 @@ def _parse_cmap(
         cmap_kw = cmap_kw or {}
         norm_kw = norm_kw or {}
         # If norm is given we use it to set vmin and vmax
+        if (vmin is not None or vmax is not None) and norm is not None:
+            print(vmin, vmax, norm)
```
beckermr: This print statement needs to be removed.

cvanelteren: done was too quick to ping you

beckermr (approved):
test failure is indeed expected and correct. LGTM!

PR #78, opened 2025-02-15
fix: remove race condition for pushes of tags

beckermr:
I missed one of the race conditions.


cvanelteren (changes requested):
The path is issue is caused by the racing condition? The test workflow contains the path directly.

beckermr:
The race condition is duplicate uploads to test pypi, not the path thing you fixed.

cvanelteren:
Any clue why the test fails? 
image

cvanelteren:
Ah I think I might. Will fix. LGTM

PR #79, opened 2025-02-15
use seed for reproducibility

cvanelteren:
Think the test introduced in #76 didn't use the seed fixed in the test causing a potential reproducibility issue in #78


cvanelteren:
Merging as the error is expected.

PR #83, opened 2025-02-19
Fix demo function not extracting colormaps

### LINKED ISSUE #82: Color Cycles Overview not rendered
I am just trying out ultraplot for the first time and realized that on [the color cycles pages](https://ultraplot.readthedocs.io/en/latest/cycles.html), the overview of color cycles seems currently not to be rendered (`uplt.show_cycles(rasterized=True)`).

Thank you for trying to keep proplot/ultraplot alive and up to date!

cvanelteren:
fixes #82

### DIFF
--- a/ultraplot/demos.py
+++ b/ultraplot/demos.py
@@ -832,12 +832,12 @@ def show_cycles(*args, **kwargs):
         ]
         ignore = ()
     else:
-        cycles = [
-            cmap
-            for cmap in pcolors._cmap_database.values()
-            if isinstance(cmap, pcolors.DiscreteColormap)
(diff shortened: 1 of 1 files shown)


cvanelteren:
Converting to draft to figure out why some are missing

cvanelteren:
The colormap names are changed by the class `_Colormap` inside ultraplot's color. It seems to copy the name and append a "_" which the original proplot code checks for but then ignores. I converted the code to accept a dictionary and passing along the name that exists in the database. This seems to work, but the naming conversion  still puzzles me and is a legacy artifact from proplot.

Issue #84, opened 2025-02-19
Colormaps having odd naming when converted internally


cvanelteren:
Colormaps are converted internally and renamed. This caused a bug in #82 but is recently fixed in main. The issue can be seen from

```python
import ultraplot as plt
for name, cmap in plt.colormaps.items():
    print(f"{cmap.name=} {name=}")
# prints  for example cmap.name='_BR_copy' name='br'
```

The name of the colormaps if correct, and hence all the functions work correctly. I have the  intuition by integration mpl's colormaps and ultraplot's colormaps  that this  renaming is performed to avoid potential  naming collision. I believe this would be unnecessary and is rather confusing.

The relevant code can be found in `colors.py`

```python

 def _make_name(self, suffix=None):
        """
        Generate a default colormap name. Do not append more than one
        leading underscore or more than one identical suffix.
        """
        name = self.name
        name = name or ""
        if name[:1] != "_":
            name = "_" + name
        suffix = suffix or "copy"
        suffix = "_" + suffix
        if name[-len(suffix) :] != suffix:
            name = name + suffix
        return name
```

cvanelteren:
Note that the internal lookup for colormaps ignore cases.

cvanelteren:
For the user this may never cause an issue unless the explicit `cmap.name`  is used. I don't really know why the maps are internally converted to this plus the suffix, but I think it should be removed.

cvanelteren:
Still an issue but closing as it is not crucial for now. Will reopen if there is a need from the community.

PR #85, opened 2025-02-20
Renaming colormaps fix

cvanelteren:
Addressing #84. I am pushing these changes to invoke a pull that compares it with the baseline. It replaces the logic that converts the colormap names to a `_` leading name with `copy` added to it.


cvanelteren:
I am leaving this issue for now as it does not affect performance in any way as the colormaps are still retrieved through the ColormapDatabase with the normal names. Just a note that ultraplot seems to rely on converting the names internally such that the colormaps are correctly merged inside `Colormap`. Retrieval of the colormap may mutate the colormap in an erroneous way that I don't fully comprehend. For example

```python
def test_input_parsing_cycle():
    """
    Test the potential inputs to cycle
    """
    # # The first argument is a string or an iterable of strings
    with pytest.raises(ValueError):
        cycle = uplt.Cycle(None)

    # # Empty should also be handled
    cycle = uplt.Cycle()

    # # # Test singular string
    cycle = uplt.Cycle("Blues")
    # target = uplt.colormaps.get_cmap("Blues")
    target = uplt.colormaps.get_cmap("blues")
    first_color = cycle.get_next()["color"]
    first_color = uplt.colors.to_rgba(first_color)
    assert np.allclose(first_color, target(0))

    # Test composition
    cycle = uplt.Cycle("Blues", "Reds", N=2)
    lower_half = uplt.colormaps.get_cmap("blues")
    upper_half = uplt.colormaps.get_cmap("reds")
    first_color = uplt.colors.to_rgba(cycle.get_next()["color"])
    last_color = uplt.colors.to_rgba(cycle.get_next()["color"])
    assert np.allclose(first_color, lower_half(0.0))
    print(last_color, upper_half(1.0))
    assert np.allclose(last_color, upper_half(1.0))
```

These test can pass if one of the final two tests commented out (either one). It halts on an error where the colormap is trying to merge continuous and discrete maps. With the original naming convention they are both converted internally to continuous -- even if they only contain finite size color.  When the naming is changed to:

```python
        name = self.name or ""
        suffix = suffix or "copy"
        if suffix:
            suffix += "_" + suffix
        if suffix and not name.endswith(suffix):
            name += suffix
```
inside `_make_name` it does not do this step but it can. Again it is opague to me when the conversion happens exactly, but it does not affect performance unless one attempts to access the `.name` property and need the unmodified name somehow.

PR #87, opened 2025-02-20
allows cycle to be a tuple

### LINKED ISSUE #86: Cycle object not parsed correctly as tuple
From the docs

```python
import ultraplot as pplt
import numpy as np
gs = pplt.GridSpec(nrows=3, ncols=2)
fig = pplt.figure(refwidth=2.2, span=False, share='labels')

# Vertical vs. horizontal
data = (state.rand(10, 5) - 0.5).cumsum(axis=0)
ax = fig.subplot(gs[0], title='Dependent x-axis')
ax.line(data, lw=2.5, cycle='seaborn')
ax = fig.subplot(gs[1], title='Dependent y-axis')
ax.linex(data, lw=2.5, cycle='seaborn')

# Vertical lines
gray = 'gray7'
data = state.rand(20) - 0.5
ax = fig.subplot(gs[2], title='Vertical lines')
ax.area(data, color=gray, alpha=0.2)
ax.vlines(data, negpos=True, lw=2)

# Horizontal lines
ax = fig.subplot(gs[3], title='Horizontal lines')
ax.areax(data, color=gray, alpha=0.2)
ax.hlines(data, negpos=True, lw=2)

# Step
ax = fig.subplot(gs[4], title='Step plot')
data = state.rand(20, 4).cumsum(axis=1).cumsum(axis=0)
cycle = ('gray6', 'blue7', 'red7', 'gray4')
ax.step(data, cycle=cycle, labels=list('ABCD'), legend='ul', legend_kw={'ncol': 2})

# Stems
ax = fig.subplot(gs[5], title='Stem plot')
data = state.rand(20)
ax.stem(data)
fig.format(suptitle='Line plots demo', xlabel='xlabel', ylabel='ylabel')

```

Bottom left plot is not correctly showing the colors. Can be fixed by wrapping it in a cycler object.

cvanelteren:
Fixes #86

### DIFF
--- a/ultraplot/axes/plot.py
+++ b/ultraplot/axes/plot.py
@@ -2488,7 +2488,7 @@ class PlotAxes(base.Axes):
                 resolved_cycle = constructor.Cycle(rc["axes.prop_cycle"])
             case str() if cycle.lower() == "none":
                 resolved_cycle = None
-            case str() | int():
+            case str() | int() | tuple():
                 resolved_cycle = constructor.Cycle(cycle, **cycle_kw)
             case constructor.Cycle():
                 resolved_cycle = constructor.Cycle(cycle)


ultraplot/axes/plot.py
```diff
@@ -2488,7 +2488,7 @@ def _parse_cycle(
                 resolved_cycle = constructor.Cycle(rc["axes.prop_cycle"])
             case str() if cycle.lower() == "none":
                 resolved_cycle = None
-            case str() | int():
+            case str() | int() | tuple():
```
beckermr, line 2491: Do we want to support any iterable here, not just a tuple?

beckermr (changes requested):

cvanelteren:
Pinged a little too early -- it was a bit of a rabbit hole I went into but I think this should be correct.

cvanelteren:
Some tests fail -- particularly on the colormaps. Judging  from the actual differences they are minor. I am not exactly sure why they are different. Perhaps @beckermr can spot a mistake with some fresh eyes. Can confirm that the plot on #86 looks at it should now.

beckermr:
My guess is that the Cycle object itself is an iterate so you need to move that case statement above the one with Iterable or add a new case statement after the one that catches Cycle objects.

cvanelteren:
Good catch. That fixed it. Results look good now. The failure relates to a missing test and a fixed error.

cvanelteren:
image
image

beckermr:
That helped a ton. There is one failure in the box plots that needs investigating.

beckermr:
Ah so the new boxplot plot is the correct one?

beckermr (approved):
Assuming the new boxplot is correct, Lgtm!

cvanelteren:
New one is correct I think. Thanks will push some styling and then merge it.

PR #91, opened 2025-02-21
Fixes heatmap not showing labels.

cvanelteren:
This PR makes the logic for `pcolormesh` specific to `QuadMesh` which is the object returned by `pcolormesh` and is more specific than a `Collection`. 

This ensures that labels are added correctly again when using a heatmap

```python
import ultraplot as plt, numpy as np

x = np.random.rand(10, 10)
fig, ax = plt.subplots(figsize=(5, 5))
ax.heatmap(x, labels=True)
fig.show()

```
image

### DIFF
--- a/ultraplot/axes/plot.py
+++ b/ultraplot/axes/plot.py
@@ -1388,12 +1388,70 @@ class PlotAxes(base.Axes):
         formatter = constructor.Formatter(
             formatter, precision=precision, **formatter_kw
         )  # noqa: E501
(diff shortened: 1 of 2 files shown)


beckermr (changes requested):
Let's add a empty grid cells on the edges and in the middle of the plot to ensure that looks OK too.

cvanelteren:
Looks good. Baseline comp should fail on the added test.

beckermr (approved):

PR #92, opened 2025-02-21
Doc link fix

cvanelteren:
Updates the docs to ensure that the links are visualized. It further also improves styling to use `UltraPlot` in text and `ultraplot` in code. 

Addresses #90

PR #93, opened 2025-02-21
prod: lint for docstring style

beckermr:
This might help with #92.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -18,4 +18,4 @@ repos:
     rev: 0.6.2
     hooks:
       - id: pydoclint
-        args: [--style=numpy]
+        args: [--style=numpy, --arg-type-hints-in-signature=False]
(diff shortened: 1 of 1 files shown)


cvanelteren:
Good idea. Type hinting the whole codebase would be no fun task however :\

beckermr:
We can turn off some of the rules depending on what we need/want.

beckermr:
Config options are here: https://jsh9.github.io/pydoclint/config_options.html

We could also use ruff/ruff-format which might offer automated fixes depending on the rule.

.pre-commit-config.yaml
```diff
@@ -14,3 +14,8 @@ repos:
     rev: 25.1.0
     hooks:
       - id: black
+  - repo: https://github.com/jsh9/pydoclint
+    rev: 0.6.2
+    hooks:
+      - id: pydoclint
+        args: [--style=numpy]
```
beckermr, line 21:
```suggestion
        args: [--style=numpy, --arg-type-hints-in-signature=False]
```

beckermr:
I am closing this for now. We can try it again later if we want.

PR #94, opened 2025-02-21
Doc link fix clean

cvanelteren:
The styling introduced in #92  was not necessary. The autoapi was causing issues with autodoc. The indicators `:class:` and so on are taken care of with `py:obj` implicitly which was messed up by autoapi. I did add the styling from the first two commits on #92.


cvanelteren:
Closing this. What I thought to be the case is in fact not the case.

PR #99, opened 2025-02-25
filter out property if not set

cvanelteren:
#98 Addresses an issue where explicit key word is given or not. By default the first argument is assumed to be the color spec but that does not have to be the case.

### DIFF
--- a/ultraplot/constructor.py
+++ b/ultraplot/constructor.py
@@ -881,10 +881,10 @@ markeredgecolors, markerfacecolors
 
     def _handle_empty_args(self, props, kwargs):
         """Handle case when no positional arguments are provided."""
-        props.setdefault("color", "black")
+        props.setdefault("color", ["black"])
         if kwargs:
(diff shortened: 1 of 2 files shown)


beckermr (changes requested):
This PR needs to add the exact test case from the original issue.

cvanelteren:
They are in there. Just refactored and expanded.

beckermr (changes requested):
I see this test

```
def test_cycler_factory():
    """Test the factory function that creates cycler objects"""
    # Test basic factory creation
    cycler = uplt.Cycle(colors=["red", "blue"])
    assert isinstance(cycler, uplt.Cycle)
```

but it doesn't check the list of colors coming out. Please add that.

cvanelteren:
`    assert props1["color"] != props2["color"]  # color should cycle
` 

This line does it. But I  will make it more explicit.

beckermr (approved):

PR #102, opened 2025-02-27
Texgyre fix

cvanelteren:
The glyph apparently is not available in tex-gyre. Setting a fallback font.


cvanelteren:
Note reverts #101

Issue #89, opened 2025-02-20
show image comparison failures on github pages


beckermr:
It'd be nice for the CI to publish a github pages website for image comparison failures.

cvanelteren:
Yes agreed. That would help with the wackamole I have been playing today.

cvanelteren:
The docs take pretty long to build. Putting a deep before the pytest would be an inconvenience. How do you reckon we could handle this?

beckermr:
I think we can publish just the image comparison artifacts with the html page via pushing to a branch in github pages. I'll try and mock something up when I get a chance.

cvanelteren:
How about something like #97?

beckermr:
OK So I fully debugged the plot failure script and we're stuck at a trick permissions issue. This [workflow](https://github.com/Ultraplot/ultraplot/actions/runs/14297675863/job/40066929509) has the error:

```
[feat-abcloc 956ba149] Update plot comparison archives [skip ci]
 6 files changed, 314[19](https://github.com/Ultraplot/ultraplot/actions/runs/14297675863/job/40066929509#step:7:20) insertions(+)
 create mode 100644 docs/_static/mpl-results/[20](https://github.com/Ultraplot/ultraplot/actions/runs/14297675863/job/40066929509#step:7:21)250406_232754_c9b3c85_pr--1/3.10-3.9/fig_comparison.html
 create mode 100644 docs/_static/mpl-results/20250406_232754_c9b3c85_pr--1/3.11-3.9/fig_comparison.html
 create mode 100644 docs/_static/mpl-results/20250406_232754_c9b3c85_pr--1/3.12-3.9/fig_comparison.html
 create mode 100644 docs/_static/mpl-results/20250406_[23](https://github.com/Ultraplot/ultraplot/actions/runs/14297675863/job/40066929509#step:7:24)2754_c9b3c85_pr--1/ultraplot.tests.test_axes.test_outer_labels/result.png
 create mode 100644 docs/_static/mpl-results/archives.txt
 create mode 100644 docs/plot_comparison_results.rst
remote: Permission to cvanelteren/ultraplot.git denied to github-actions[bot].
fatal: unable to access 'https://github.com/cvanelteren/ultraplot/': The requested URL returned error: 403
Error: Process completed with exit code 1[28](https://github.com/Ultraplot/ultraplot/actions/runs/14297675863/job/40066929509#step:7:29).
```

[...]

PR #103, opened 2025-02-27
Upload errors to pages

cvanelteren:
Addresses #89


beckermr (changes requested):
This approach will bloat the repo. Let's use readthedocs instead.

cvanelteren:
Readthedocs does not host the data but merely pulls it from the repo. We can create a separate github page (and repo) for this. I also wouldn't wanna include this in the docs as it is more for devs or contributors.

beckermr:
I'm pretty sure read the docs hosts the data directly, but a separate repo is fine too.

beckermr:
I'd not store the baseline there, but rather just push the diff data to the repo. Note that github pages supports only one version of the site, so we'll need to put each one under a different subdirectory within a single main page.

cvanelteren:
Ok I added RtD  to the access the organization. It turns out I never gave it permission through OAuth -- not sure why it worked in the first place. PR builds are enabled now. I will start working on integrating the PR artifacts to push to RtD.

cvanelteren:
Closing this in favor of RtD approach.

PR #106, opened 2025-02-27
prod: only build ultraplot when ultraplot src changes

cvanelteren:
This PR will streamline the runs. When updating the docs, the src does not have to be built unless the src was changed.

### DIFF
--- a/.github/workflows/main-no-changes.yml
+++ b/.github/workflows/main-no-changes.yml
@@ -17,4 +17,4 @@ jobs:
   build-success:
     runs-on: ubuntu-latest
     steps:
-      - run: echo "All tests passed successfully!"
+      - run: echo "No tests were run since no ultraplot Python code changes were detected! Marking the test suite as passed!"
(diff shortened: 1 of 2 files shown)


beckermr (changes requested):
This breaks automerge.

cvanelteren:
How will this break it?

beckermr:
When you skip the whole test suite, the required status "build-success" is never run. Here is a screen shot from this PR.

Screenshot 2025-02-27 at 6 51 53 AM

beckermr:
One approach would be to add a separate job that runs on the same events but using the `paths-ignore` key instead and posts the build success.

.github/workflows/main-no-changes.yml
```diff
@@ -0,0 +1,20 @@
...
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  build-success:
+    runs-on: ubuntu-latest
+    steps:
+      - run: echo "All tests passed successfully!"
```
beckermr, line 20:
```suggestion
      - run: echo "No tests were run since no ultraplot Python code changes were detected! Marking the test suite as passed!"
```

beckermr (approved):

Issue #105, opened 2025-02-27
tick labels overlap when extend colorbar in contourf


cmyangcn:
As shown, when the "extend" paramenter in "contourf" is not set to "neither", after executing the following code, the resulting plot's colorbar displays two additional tick values at both ends.
```python
import pandas as pd
import ultraplot as uplt
state = np.random.RandomState(51423)
fig, ax = uplt.subplots(
    ncols=1, nrows=1
)
data = (state.rand(50, 50) - 0.1).cumsum(axis=0)
m = ax.contourf(data, cmap="grays",levels=uplt.arange(0,25,2.5),extend='both')
ax.colorbar(m, length=0.8, label="colorbar label", loc="b",ticks=5,minorticks=2.5)
```

![Image](https://github.com/user-attachments/assets/b56b08f2-a289-42b1-ab94-0ffcc07f2f6b)

cvanelteren:
if you leave out `ticks` it works. I think I will rework the code to not allow for both to be set when using a quadmesh.

cmyangcn:
If skip ticks, the colorbar’s major ticks will be set by default, and that might not be what we are after.

cvanelteren:
I did some digging but I don't fully understand the problem yet. 

There seems to be an issue that when `ticks` or `minorticks` are  given the major ticks shift the bins over by 1. We see the beginning range being plot; this should not happen.

cvanelteren:
The difference by 1 comes directly from the mpl source

```python
    def tick_values(self, vmin, vmax):
        if vmax < vmin:
            vmin, vmax = vmax, vmin
        step = self._edge.step
        vmin -= self._offset
        vmax -= self._offset
        vmin = self._edge.ge(vmin) * step
        n = (vmax - vmin + 0.001 * step) // step
        locs = vmin - step + np.arange(n + 3) * step + self._offset
        return self.raise_if_exceeds(locs)
```

cvanelteren:
![Image](https://github.com/user-attachments/assets/21a3fd54-9547-411d-8ad6-e5d33dd52ecb)

Think I figured it out. Fix following soon.

cvanelteren:
[...]

PR #108, opened 2025-02-28
Fix colorbar ticks

cvanelteren:
Addresses #105 

When `extend != 'neither'` the additional `update_ticks` is creating ghosting on the ticks. I am not sure why this happened but the `update_ticks` command is overridden by Ultraplot but deprecated by mpl. Removing the function outright breaks code elsewhere that I currently do not fully understand. I would be more in favor of removing this additional function but for now we can fix this by removing this additional call. Unittest is added to safe guard against future changes.

### DIFF
--- a/ultraplot/axes/base.py
+++ b/ultraplot/axes/base.py
@@ -1143,6 +1143,7 @@ class Axes(maxes.Axes):
         # DiscreteLocator or else get issues (see mpl #22233).
         norm = mappable.norm
         formatter = _not_none(formatter, getattr(norm, "_labels", None), "auto")
+        formatter_kw.setdefault("tickrange", (norm.vmin, norm.vmax))
         formatter = constructor.Formatter(formatter, **formatter_kw)
         categorical = isinstance(formatter, mticker.FixedFormatter)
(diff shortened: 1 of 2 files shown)


beckermr (changes requested):
Let's add `baseline/*` to the gitignore.

cvanelteren:
I thought I fixed it but now I am seeing locally weird stuff again. Back to the drawing board.

cvanelteren:
I think I had some test code messing with the formatter. I fixed the issue by forcing the tickrange to be set by default. The code creates an internal formatter that somehow does not have a default range for their ticks which makes it plot the entire range. Within the context of `extend != neither` it puts the values past the range on the location of the min and max of the original capped range (with `extend == nether`). I will check later if this can be traced by the way mpl handles these things but for now this would work --> more comfortable with what is causing the error.

cvanelteren:
I rebased and cleaned up the history here so that future people can track the changes. The fix should be ok and set default behavior will still do the things above but for now its beer time.

beckermr (approved):
Test failure is expected.

Issue #107, opened 2025-02-27
2D Xarray plotting not working for string coordinates


Ockenfuss:
Not sure if this is officially supported, since I am using the xarray plot method. But the following code seems to plot only part of the array along the axis with string labels.
```python
import numpy as np
import xarray as xr
import ultraplot as uplt

test=xr.DataArray(np.random.rand(6,6), dims=('x', 'y'), coords={'x':['a', 'b', 'c', 'd', 'e', 'f']})
fig, ax=uplt.subplots()
test.plot(x='x', ax=ax)
```

![Image](https://github.com/user-attachments/assets/eea516e7-d5aa-474b-9db7-c24fb6258e19)

Ultraplot Version: 1.0.7
Xarray Version: 2025.1.2

cvanelteren:
Hi thanks for reporting. I am very not familiar with xarray as I have never used in my work -- so I am flying a bit blind here, but judging from the output this should be working. However, I think this is using matplotlib as the backend. The `test.plot` is using probably matplotlib on the backend.

If I change it to ultraplot I get this

```python
import numpy as np
import xarray as xr
import ultraplot as uplt

test = xr.DataArray(
    np.random.rand(6, 6),
    dims=("x", "y"),
    coords={"x": "a b c d e f".split()},
)
fig, ax = uplt.subplots()
ax.plot(x="x", y="y", data=test)
uplt.show(block=1)
```

![Image](https://github.com/user-attachments/assets/86b0187c-398d-41bb-8391-d53fb977c827)

which is at least showing the correct columns but not exactly what you have in mind. Imshow will work fine but then you have to parse the x, y labels separately.

cvanelteren:
Also see the docs https://ultraplot.readthedocs.io/en/latest/2dplots.html#Pandas-and-xarray-integration

Ockenfuss:
Thanks for the quick response! Yes, I don't know how it works internally, but I think the xarray plotting is only tested for matplotlib. This is the result if providing a matplotlib axes instead:
```python
import numpy as np
import xarray as xr
import matplotlib.pyplot as plt

test=xr.DataArray(np.random.rand(6,6), dims=('x', 'y'), coords={'x':['a', 'b', 'c', 'd', 'e', 'f']})
fig, ax=plt.subplots()
test.plot(x='x', ax=ax)
```

![Image](https://github.com/user-attachments/assets/f480d613-014e-423d-b018-1f10e03a577a)

[...]

PR #111, opened 2025-03-01
fix issue where cat. data is now shown correctly

cvanelteren:
When trying to address #107, I noticed that the strings are not formatted correctly.
 For example,
image

This can be fixed by converted it to numpy

which yields

image


cvanelteren:
Shows only failed test where the size differs. Unsure why...but visually no artifacts.

beckermr (changes requested):
No I see two plots where now the axis labels have disappeared. 
Screenshot 2025-03-02 at 7 15 47 AM

Screenshot 2025-03-02 at 7 16 15 AM

cvanelteren:
Ah I see. Then let's ignore this PR for now. Will do some more digging. Converting it to draft.

cvanelteren:
Closing and not fixing as the issue resides with xarray.

PR #113, opened 2025-03-02
Only store failed tests mpl-pytest

cvanelteren:
Pytest does not allow to only store failed tests. This PR addresses this issue by providing a plugin that removes all successful tests from the results dir. It needs to be called as

> pytest --mpl --mpl-baseline-path=baseline --mpl-default-style="./ultraplot.yml" --store-failed-only --mpl-results-path=./results ultraplot/tests/test_1dplots.py::test_heatmap_labels --mpl-generate-summary=html

This builds upon the issues raised in #103 #89


cvanelteren:
Note this reasons from the perspective that all tests are run. It should be better to inverse the pattern (only store failed) but this way is easier and does not require an additional directory.

cvanelteren:
The idea would be to host the full baseline on a separate repo and push the failed to readthedocs as suggested by @beckermr

beckermr (changes requested):
We need to test this live by adding the flag, making a test fail, and ensuring the output looks correct.

cvanelteren:
Did this locally, but I will push an error now.

beckermr:
We need to add the flag to the pytest call too in order to test.

beckermr (changes requested):
The artifacts are still ~90 MB and I see all of the plots in them. Something is not working here...

cvanelteren:
Will check later.

cvanelteren:
Should be good now @beckermr

beckermr:
I am trying a failure live again.

ultraplot/tests/test_1dplots.py
```diff
@@ -31,7 +31,7 @@ def test_auto_reverse():
     axs[4].pcolor(x, y[::-1], z)
     axs[5].format(xlim=(0, 9), ylim=(0, 9))  # manual override!
     axs[5].pcolor(x, y[::-1], z)
-    fig.format(suptitle="Auto-reverse test", collabels=["reverse", "fixed"])
+    fig.format(suptitle="auto-reverse test", collabels=["reverse", "fixed"])
```
beckermr, line 34:
```suggestion
    fig.format(suptitle="Auto-reverse test", collabels=["reverse", "fixed"])
```

beckermr (approved):

PR #114, opened 2025-03-02
Attempt fix workflow docs build

cvanelteren:


.github/workflows/build-docs.yml
```diff
@@ -24,13 +24,10 @@ jobs:
       - uses: mamba-org/setup-micromamba@v2.0.4
         with:
           environment-file: ./environment.yml
-      - name: Check dependencies
-        run: |
-          micromamba activate ultraplot-dev
-          micromamba list
```
beckermr: You needed `shell: bash -el {0}` in this step. The eval command you added below is not needed.

environment.yml
```diff
@@ -12,6 +12,7 @@ dependencies:
   - pytest
   - pytest-mpl
   - jupyter
+  - ipython <= 9.0.0
```
beckermr, line 15:
```suggestion
  - ipython <=9.0.0
```

environment.yml
```diff
@@ -12,9 +12,10 @@ dependencies:
   - pytest
   - pytest-mpl
   - jupyter
+  - ipython >= 9.0.0
```
beckermr, line 15:
```suggestion
  - ipython >=9.0.0
```

ultraplot/axes/geo.py
```diff
@@ -5,6 +5,18 @@
...
+
+try:
+    from cartopy.io import DownloadWarning
+
+    warnings.filterwarnings("ignore", category=DownloadWarning)
+except ImportError:
+    # In case cartopy isn't installed yet when conf.py is executed
+    pass
+
+
```
beckermr, line 19: I don't like ignoring warnings here by default for users. We should move this back to the docs build itself.

beckermr (changes requested):
A quick review while I have time.

Besides the comments in the review, the change to the ipython version requirement is problematic. We should support older versions there, so I think we'll want the config module code to try the old syntax and if it fails, try the new syntax, and if that fails, then raise.

ultraplot/axes/geo.py
cvanelteren: So the conf did not remove the warnings from the test, do you know why?

ultraplot/axes/geo.py
```diff
@@ -5,6 +5,10 @@
 import copy
 import inspect
 
+# Surpress warnings from cartopy when downloading data inside docs env
+import warnings
+
+
```
beckermr, line 11:
```suggestion
```

ultraplot/axes/geo.py
beckermr: I do not know why. Can you point me to the code that makes the plot? I wonder if we need to embed the warnings filter inside the doc strings maybe?

PR #115, opened 2025-03-02
fix: matrix test for no changes is running when it should not be

beckermr:
I am seeing this dummy jobs run even when there are changes to the python code. This PR is trying to fix that.


.pre-commit-config.yaml
```diff
@@ -14,3 +14,8 @@ repos:
     rev: 25.1.0
     hooks:
       - id: black
+  - repo: https://github.com/jsh9/pydoclint
+    rev: 0.6.2
+    hooks:
+      - id: pydoclint
+        args: [--style=numpy, --arg-type-hints-in-signature=False]
```
beckermr, line 21:
```suggestion
```

beckermr:
@cvanelteren This one is ready for review. The previous fix I pushed for the skips was not working properly. The logic in the GHA syntax is not quite what we needed. This logic works better, but runs a few more dummy jobs.

cvanelteren (changes requested):
Can't we put the build under a separate step with a singular if statement rather than adding them everywhere?

beckermr:
I have not figured out a way to do that. I'll keep digging.

.github/workflows/build-ultraplot.yml
```diff
@@ -85,7 +86,6 @@ jobs:
 
       # Return the html output of the comparison even if failed
       - name: Upload comparison failures
```
beckermr, line 88:
```suggestion
      - name: Upload comparison failures
        if: always()
```

.github/workflows/build-ultraplot.yml
```diff
@@ -51,6 +51,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
+
```
beckermr, line 54:
```suggestion
```

beckermr:
OK @cvanelteren. I have fixed things up and tested that it works properly in all three cases.

- no python changes
- python changes that pass
- python changes that fail

This should be ready to merge!

.github/workflows/main.yml
```diff
@@ -110,7 +122,10 @@ jobs:
           echo "matplotlib-versions=$(echo $MPL_VERSIONS | jq -c)" >> $GITHUB_OUTPUT
 
   build:
-    needs: get-versions
+    needs:
+      - get-versions
+      - run-if-changes
+    if: always() && needs.run-if-changes.outputs.run == 'true' && needs.get-versions.result != 'skipped'
```
beckermr, line 128:
```suggestion
    if: always() && needs.run-if-changes.outputs.run == 'true' && needs.get-versions.result == 'success'
```

PR #118, opened 2025-03-06
prod: only use readthedocs for PR tests

beckermr:
As it says...


cvanelteren:
I am not entirely sure I follow this PR. Could you elaborate a bit? The workflow is now skipped entirely.

cvanelteren (changes requested):
Wouldn't we need to change this then to tag?

beckermr:
The docs workflow on github actions isn't what makes the docs available on readthedocs or does builds for readthedocs. That is all done on their machines. This GHA job is pointless. 

Further, with PR builds enabled for readthedocs, we can now see drafts of the documentation and ensure it builds in PRs and deploys properly.

cvanelteren:
Ah I see let's remove the file then.

beckermr:
Yep. It is entirely separate.

PR #124, opened 2025-03-14
bugfix for Axes.legend when certain keywords are set to str

syrte:
Bug: `kwargs` just has popped out key, and thus, `kwargs[key]` won't work.  It raises KeyError when the value is a string (e.g.,`columnspacing='1em'`)


cvanelteren:
Will fail on the added test.

cvanelteren:
Thanks!

Issue #125, opened 2025-03-16
[Feature request] Adding ticks to geographic plots


664787022:
Thanks, @cvanelteren , for your continued development of Proplot into Ultraplot

Adding ticks to geographic plots is a practical  feature in geoscience studies.

Although this feature was requested in Proplot, it was not merged into the codebase, see  [Support adding tick marks to geographic plots](https://github.com/proplot-dev/proplot/pull/253)

Hopefully, @cvanelteren  can add this feature to Ultraplot !

cvanelteren:
I believe this is already possible:

![Image](https://github.com/user-attachments/assets/4bff9984-999b-4ca9-9410-ae4387800a18)

```
import ultraplot as uplt, numpy as np

fig, axs = uplt.subplots(proj="cyl")
axs.format(
    land=True,
    labels=True,
    lonlines=20,
    latlines=20,
    gridminor=False,
    lonlim=(-140, 60),
    latlim=(-10, 50),
)
uplt.show(block=1)
```

cvanelteren:
Or more succinctly: 

![Image](https://github.com/user-attachments/assets/9e4273a0-f21b-42d7-8683-db3df8f84283)

``` python
import ultraplot as uplt, numpy as np
from itertools import product

fig, axs = uplt.subplots(ncols=2, nrows=2, proj="cyl")

settings = dict(
    land=True,
    labels=True,
    lonlines=20,
    latlines=20,
    lonlim=(-140, 60),
    latlim=(-10, 50),
)
for idx, (gridminor, gridmajor) in enumerate(product([True, False], [True, False])):
    settings["gridminor"] = gridminor
    settings["grid"] = gridmajor
    axs[idx].format(**settings)
uplt.show(block=1)
```

cvanelteren:
Ah wait I read this wrong, you want the ticklens not the grid to change.

664787022:
> Ah wait I read this wrong, you want the ticklens not the grid to change.

Yes, ticks in geographic plots are used more frequently than gridlines, at least in my field. They help keep the map clean and visually appealing by avoiding clutter from multiple gridlines.

![Image](https://github.com/user-attachments/assets/0e669a62-73f3-4c23-9978-97f4f0cad165)

cvanelteren:
![Image](https://github.com/user-attachments/assets/374d1c67-3e37-4dda-a05f-672c0c4123bf) working on it ;-)

cvanelteren:
[...]

PR #126, opened 2025-03-16
Allow ticks for geoaxes

cvanelteren:
This PR addresses #125. It introduces a test that tests specifically three use cases.
It modifies the code to allow for showing ticks in geoaxis. Currently, the tickers are completely controlled through the backend with small modifications to ensure high quality plots. This PR makes it to add ticks. It does this by manually adding a ticker when new keywords `lontick` or `lattick` is given which can take a bool or numeric as an input. When set to `None` the option is effectively ignored.

The defaults should generate a sensible plot with lat and long lines, but without ticks. Then it specifically tests whether leaving the ticks on on one axis does not hamper the other axes.

image

Edit the parameters added would create a 2^4 test plots in terms of the combinations; I opted for specific tests that are a bit more complex but are quicker to generate while still testing all the properties.

### DIFF
--- a/docs/projections.py
+++ b/docs/projections.py
@@ -344,7 +344,7 @@ for globe in (False, True):
 # borders using :ref:`settings <rc_UltraPlot>` like `land`, `landcolor`, `coast`,
 # `coastcolor`, and `coastlinewidth`. Finally, since `ultraplot.axes.GeoAxes.format`
 # calls `ultraplot.axes.Axes.format`, it can be used to add axes titles, a-b-c labels,
-# and figure titles, just like :func:`ultraplot.axes.CartesianAxes.format`.
+# and figure titles, just like :func:`ultraplot.axes.CartesianAxes.format`. UltraPlot also adds the ability to add tick marks for longitude and latitude using the keywords `lontick` and `lattick` for rectilinear projections only. This can enhance contrast and readability under some conditions, e.g. when overlaying contours.
 #
 # For details, see the `ultraplot.axes.GeoAxes.format` documentation.
 
--- a/ultraplot/axes/geo.py
+++ b/ultraplot/axes/geo.py
@@ -4,6 +4,7 @@ Axes filled with cartographic projections.
 """
 import copy
 import inspect
+from functools import partial
 
 import matplotlib.axis as maxis
 import matplotlib.path as mpath
@@ -15,7 +16,14 @@ from .. import constructor
 from .. import proj as pproj
 from ..config import rc
 from ..internals import ic  # noqa: F401
-from ..internals import _not_none, _pop_rc, _version_cartopy, docstring, warnings
+from ..internals import (
+    _not_none,
+    _pop_rc,
+    _version_cartopy,
+    docstring,
+    warnings,
+)
+from ..utils import units
 from . import plot
 
 try:
@@ -65,6 +73,10 @@ longrid, latgrid, grid : bool, default: :rc:`grid`
 longridminor, latgridminor, gridminor : bool, default: :rc:`gridminor`
     Whether to draw "minor" longitude and latitude lines.
     Use the keyword `gridminor` to toggle both at once.
+lonticklen, latticklen, ticklen : unit-spec, default: :rc:`tick.len`
+    Major tick lengths for the longitudinal (x) and latitude (y) axis.
+    %(units.pt)s
+    Use the keyword `ticklen` to set both at once.
 latmax : float, default: 80
     The maximum absolute latitude for gridlines. Longitude gridlines are cut off
     poleward of this value (note this feature does not work in cartopy 0.18).
@@ -562,6 +574,9 @@ class GeoAxes(plot.PlotAxes):
         latgrid=None,
         longridminor=None,
         latgridminor=None,
+        ticklen=None,
+        lonticklen=None,
+        latticklen=None,
         latmax=None,
         nsteps=None,
         lonlocator=None,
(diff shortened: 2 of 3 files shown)


cvanelteren:
Will fail on added test.

cvanelteren:
Note to self, we may want to add this feature to the docs.

cvanelteren:
Removed some scaling around, but should be good now @beckermr

cvanelteren:
We can do the names as fallback and check with direct computation. I will draft something up.

docs/projections.py
```diff
@@ -344,7 +344,7 @@
 # borders using :ref:`settings <rc_UltraPlot>` like `land`, `landcolor`, `coast`,
 # `coastcolor`, and `coastlinewidth`. Finally, since `ultraplot.axes.GeoAxes.format`
 # calls `ultraplot.axes.Axes.format`, it can be used to add axes titles, a-b-c labels,
-# and figure titles, just like :func:`ultraplot.axes.CartesianAxes.format`.
+# and figure titles, just like :func:`ultraplot.axes.CartesianAxes.format`. UltraPlot also adds the ability to add tick marks for longitude and latitude using the keywords `lontick` and `lattick` for flat projections only. This can enhance contrast and readability under some conditions, e.g. when overlaying contours.
```
beckermr, line 347:
```suggestion
# and figure titles, just like :func:`ultraplot.axes.CartesianAxes.format`. UltraPlot also adds the ability to add tick marks for longitude and latitude using the keywords `lontick` and `lattick` for rectilinear projections only. This can enhance contrast and readability under some conditions, e.g. when overlaying contours.
```

ultraplot/axes/geo.py
```diff
@@ -65,6 +71,18 @@
...
+    tick length from rcParams. If `False` or `None`, ticks are not drawn. If a numeric value
+    is provided, ticks are drawn with that length.
+lontick : bool or float, optional
+    Controls tick marks specifically for longitude. Takes precedence over `tick` parameter.
+    If `True`, uses the default tick length. If `False` or `None`, longitude ticks are not drawn.
+    If a numeric value is provided, longitude ticks are drawn with that length in points.
+lattick : bool or float, optional
+    Controls tick marks specifically for latitude. Takes precedence over `tick` parameter.
+    If `True`, uses the default tick length. If `False` or `None`, latitude ticks are not drawn.
+    If a numeric value is provided, latitude ticks are drawn with that length in points.
```
beckermr, line 85: The keyword argument names and semantics for these tick marks needs to follow the same API as done in format for `x` and `y` tick marks. See the doc string here: https://github.com/Ultraplot/ultraplot/blob/main/ultraplot/axes/cartesian.py#L38.

ultraplot/axes/geo.py
```diff
@@ -1673,3 +1768,71 @@ def _update_minor_gridlines(self, longrid=None, latgrid=None, nsteps=None):
...
+        # Check if the transformed points are still collinear
+        # Points are collinear if the slopes between consecutive points are equal
+        x0, y0 = transformed_points[0]
+        x1, y1 = transformed_points[1]
+        x2, y2 = transformed_points[2]
+
+        # Calculate slopes
+        if (x1 - x0) == 0 or (x2 - x1) == 0:  # Avoid division by zero
+            # Check if both are vertical lines
+            return (x1 - x0) == 0 and (x2 - x1) == 0
```
beckermr, line 1807: These tests for 0 need to use `allclose` as well.

ultraplot/axes/geo.py
```diff
@@ -1673,3 +1768,71 @@ def _update_minor_gridlines(self, longrid=None, latgrid=None, nsteps=None):
 # Apply signature obfuscation after storing previous signature
 GeoAxes._format_signatures[GeoAxes] = inspect.signature(GeoAxes.format)
 GeoAxes.format = docstring._obfuscate_kwargs(GeoAxes.format)
+
+
+def _is_rectilinear_projection(ax):
```
beckermr, line 1773: There is some missing coverage in this file. We need to add tests to cover branches.

beckermr (changes requested):
See my comments above.

ultraplot/axes/geo.py
```diff
@@ -65,6 +73,16 @@
 longridminor, latgridminor, gridminor : bool, default: :rc:`gridminor`
     Whether to draw "minor" longitude and latitude lines.
     Use the keyword `gridminor` to toggle both at once.
+lonticks, latticks : locator-spec, optional
+    Used to determine the longitude and latitude axis tick mark positions. Passed
+    to the `~ultraplot.constructor.Locator` constructor. Can be float,
+    list of float, string, or `matplotlib.ticker.Locator` instance.
+    Use `[]`, `'null'`, or `'none'` for no ticks.
```
beckermr, line 80: This needs to lonticklen, latticklen, and ticklen. The extra semantics around Boolean values needs to be removed. Ticklen should set the length for both Lon and lat. The principle here is that the api should be identical.

beckermr (changes requested):

ultraplot/axes/geo.py
```diff
@@ -783,13 +783,15 @@ def format(
...
-                f"Projection is not flat ignoring {lonticks=} and {latticks=} settings"
-            )
+        if lonticks or latticks:
+            # Only add warning when ticks are given
+            if _is_rectilinear_projection(self):
+                self._add_geoticks("x", lonticks, tick)
+                self._add_geoticks("y", latticks, tick)
+            else:
+                warnings._warn_ultraplot(
+                    f"Projection is not flat ignoring {lonticks=} and {latticks=} settings"
```
beckermr, line 793:
```suggestion
                    f"Projection is not rectilinear. Ignoring {lonticks=} and {latticks=} settings."
```

beckermr (changes requested):

ultraplot/axes/geo.py
cvanelteren: I updated the logic to better reflect the logic within the class. Some of the checks were a bit overkill as they are caught by other functions already. This should satisfy the codecov patches (will see when its updated).  The major change is that some of the checks were a bit overkill since the format function itself sets the projection class and as such we could not check for that with the functions available in pytest, although we could  add tests for private functions (but that would be a bit overkill again).

ultraplot/axes/geo.py
cvanelteren: ah I thought you were referring to xtick and ytick

cvanelteren:
todo: rm boolean logic and add unit conversion if string.

cvanelteren:
Looks good now! @beckermr. I will rebase if approved.

cvanelteren:
It is a little tricky to test the untested and partial parts raised by codecov. The main issue is that the code itself does not allow to objects in such a way to these high level function that would create the problems. We could either provide tests for the low level functions, leave them as is, or remove the additional logic that would safeguard against edge cases. I am leaning towards the second option.

cvanelteren:
To add fuel to the fire I also just noticed that there is `_proj_non_rectangular` which also performs a name check. 

I can attempt to modify the projection functions directly using a cartopy class, but the name property is read only -- making testing the latter harder. Basemap has similar saveguards for the names.

ultraplot/axes/geo.py
```diff
@@ -65,6 +73,12 @@
 longridminor, latgridminor, gridminor : bool, default: :rc:`gridminor`
     Whether to draw "minor" longitude and latitude lines.
     Use the keyword `gridminor` to toggle both at once.
+lonticklen, latticklen, ticklen: int, float, bool optional
+
```
beckermr, line 77:
```suggestion
```

ultraplot/axes/geo.py
```diff
@@ -761,10 +778,80 @@ def format(
...
             )
+        # Set tick lengths for flat projections
+        if lonticklen or latticklen:
+            # Only add warning when ticks are given
+            if _is_rectilinear_projection(self):
+                self._add_geoticks("x", lonticklen, ticklen)
+                self._add_geoticks("y", latticklen, ticklen)
+            else:
+                warnings._warn_ultraplot(
+                    f"Projection is not flat ignoring {lonticklen=} and {latticklen=} settings"
```
beckermr, line 789:
```suggestion
                    f"Projection is not rectilinear. Ignoring {lonticklen=} and {latticklen=} settings."
```

ultraplot/axes/geo.py
```diff
@@ -761,10 +778,80 @@ def format(
...
 
+    def _add_geoticks(self, x_or_y, itick, ticklen):
+        """
+        Add tick marks to the geographic axes.
+
+        Parameters
+        ----------
+        x_or_y : {'x', 'y'}
+            The axis to add ticks to ('x' for longitude, 'y' for latitude).
+        itick, tick : unit-spec, default: :rc:`tick.len`
```
beckermr, line 803:
```suggestion
        itick, ticklen : unit-spec, default: :rc:`tick.len`
```

ultraplot/axes/geo.py
```diff
@@ -761,10 +778,80 @@ def format(
...
+            Major tick lengths for the x and y axis.
+            %(units.pt)s
+            Use the keyword `ticklen` to set both at once.
```
beckermr, line 806:
```suggestion
            Use the argument `ticklen` to set both at once.
```

ultraplot/axes/geo.py
```diff
@@ -761,10 +778,80 @@ def format(
...
+        Notes
```
beckermr, line 807:
```suggestion

        Notes
```

ultraplot/axes/geo.py
```diff
@@ -761,10 +778,80 @@ def format(
...
+        # Skip if no tick size specified
+        if size is None:
+            return
+        size = units(size)
+
+        ax = getattr(self, f"{x_or_y}axis")
+
+        # Get the tick positions based on the locator
+        gl = self.gridlines_major
+        # Note: set_xticks points to a different method than self.[x/y]axis.set_ticks from the mpl backend. For basemap we are adding the ticks to the mpl backend and for cartopy we are simple using their functions by showing the axis.
```
beckermr, line 823:
```suggestion
        # Note: set_xticks points to a different method than self.[x/y]axis.set_ticks
        # from the mpl backend. For basemap we are adding the ticks to the mpl backend
        # and for cartopy we are simple using their functions by showing the axis.
```

ultraplot/axes/geo.py
```diff
@@ -761,10 +778,80 @@ def format(
...
+                for object in objects:
+                    # text is wrapped in a list
+                    if isinstance(object, list) and len(object) > 0:
+                        object = object[0]
+                    if isinstance(object, mtext.Text):
+                        object.set_visible(True)
+        else:
+            setattr(gl, f"{x_or_y}padding", 2 * size)
+
+        # Note: set grid_alpha to 0 as it is controlled through the gridlines_major object (which is not the same ticker)
```
beckermr, line 849:
```suggestion
        # Note: set grid_alpha to 0 as it is controlled through the gridlines_major
        # object (which is not the same ticker)
```

beckermr, line 849: Am I understanding correctly that a user cannot use both ticks and grid lines? We should leave that as configurable by the user and not force them. A nice default is good of course. As the code looks now, I think folks would have to add the gridlines back after adding ticks. Instead, they should only have to make a single call to `.format` that adds ticks and sets the gridline alpha to zero.

cvanelteren: The ticks are exaggerated in the tests to allow the users to change the size of the ticks

beckermr: I am confused then. Why is the core setting the grid alpha to zero when adding ticks then?

cvanelteren: otherwise it doubles the grid lines.

cvanelteren: Geoaxes are a bit strange. The set tick functions are also no properly overwritten (issue with cartopy) the grid lines in UltraPlot are set manually in a separate function and set tick Params would add it independently

beckermr (approved):

PR #140, opened 2025-04-06
Add gallery to github page

cvanelteren:
To draw more people in, we need to show by example. This PR creates a dynamic gallery to show on the main page.


cvanelteren:
Well github does not allow for js to run. Halting this for now.

PR #141, opened 2025-04-06
added page for errors

cvanelteren:
Add failed tests to a special page in the docs.


beckermr (changes requested):
Is there a way to debug this before we merge, or will we have to wait for the workflow file to be on main to see if works?

cvanelteren:
Trying to figure this out right now. For some reason the workflow does not show up in actions.

beckermr:
Yeah the workflow file has to be main and be triggered in order to run.

beckermr:
We only want this for failed runs, right?

cvanelteren:
That would be ideal. Having it triggered on all would not be that bad but for now I want to make sure the action works as intended. Pushing a dummy commit to trigger it.

beckermr:
let's merge to main - worst case we put in a bug fix.

PR #146, opened 2025-04-07
Prepare the pytest results in the workflow

cvanelteren:
This PR adds a new workflow that automatically archives and displays matplotlib test comparison failures from CI runs. We discussed some options in #89 

## What does this PR do?

1. Adds a new GitHub workflow (`archive-pytest-results.yml`) that runs when the Matrix Test workflow fails
2. The workflow:
   - Downloads artifacts with failed plot comparisons from the test run
   - Organizes them into a structured archive in the docs directory
   - Creates an index page listing all archived test failures
   - Commits these changes to the repository

## Security considerations

- The workflow uses explicitly scoped permissions (`contents: write`)
- Processing is done on pre-rendered comparison results with no execution of untrusted code
- The workflow maintains a proper audit trail of who triggered the original workflow

### DIFF
--- a/.github/workflows/archive-pytest-results.yml
+++ b/.github/workflows/archive-pytest-results.yml
@@ -12,11 +12,12 @@ jobs:
     steps:
       - name: Print event data
         run: echo '${{ toJson(github.event) }}'
-  
+
   archive-test-results:
     runs-on: ubuntu-latest
     if: ${{ github.event.workflow_run.conclusion == 'failure' }}
-    permissions: {}
+    permissions:
+      contents: write # Need write permission to commit files
 
     steps:
       - uses: actions/checkout@v4
@@ -24,6 +25,8 @@ jobs:
           repository: ${{ github.event.workflow_run.head_repository.full_name }}
           ref: ${{ github.event.workflow_run.head_branch }}
           fetch-depth: 0
+          # Use a token with sufficient permissions
+          token: ${{ secrets.GITHUB_TOKEN }}
 
       - name: Setup Python
         uses: actions/setup-python@v4
--- a/.github/workflows/build-ultraplot.yml
+++ b/.github/workflows/build-ultraplot.yml
@@ -19,7 +19,7 @@ jobs:
     runs-on: ubuntu-latest
     timeout-minutes: 15
     defaults:
-      run: 
+      run:
         shell: bash -el {0}
     steps:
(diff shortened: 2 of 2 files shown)


cvanelteren:
Think this approach still has issues. We could do a few things from here.
We could either 
1. not commit the artifacts and merely read them. Not sure how long the artifacts will be stored on github actions.
2. Manually approve the contents 
3. Use a separate branch

For 3 we would still have the security risk I believe. 

Happy to hear your thoughts @beckermr

beckermr:
Can we embed a run of the pytest-mpl stuff directly in the docs build on read the docs?

cvanelteren:
Pytest-mpl could be added to the doc builds. I would still prefer to have it also run on the PRs since not everything concerns the docs and vice versa. For my comprehension is the main issue the automatic commit or the fact that we could be reading executable malicious data?

beckermr:
It is both. Adding the pytest-mpl tests to the readthedocs build (for PRs only) would solve this since that build happens in readthedocs without access to our secrets. We should still run the tests in the PR too ofc.

cvanelteren:
but we could still potentially open and run malicious code if one wants to commit this an run it through the tests, would it not?

cvanelteren:
Do you have a source on this? I am trying to see what is possible and if this is even possible without over complicating everything.

beckermr:
> How about we just read the artifacts from the workflows on the rst in the docs. 

I don't understand. Can you explain a bit more?

cvanelteren:
> > How about we just read the artifacts from the workflows on the rst in the docs.
> 
> I don't understand. Can you explain a bit more?

What we could do is similar to what we did in `what's new`. We would just read the workflow actions and download their artifacts from the api, extract either the full js and html in these artifacts (as an output of pytest-mpl) or just extract the images and visualize those.

cvanelteren:
My suggestion seems to not be really possible: https://github.com/actions/upload-artifact/issues/51. Or at least not "officially".

cvanelteren:
Was playing around a bit with it, and faced this screen:

image

cvanelteren:
It could be made working by providing a access token without any permissions, but it feels rather hacky than a straightforward approach. The easiest would be to host this somewhere and then just read it. Or, ya know, just leave as is currently and download the artifacts manually ;-)

Issue #149, opened 2025-04-08
Unwanted grid lines on heatmap axes and colorbar


ToryDeng:
In the latest [documentation](https://ultraplot.readthedocs.io/en/latest/2dplots.html#discrete-levels), both the heatmap axes and their associated colorbars display grid lines by default, which visually interfere with the data representation. Here is a simplified example:

```python
import ultraplot as uplt
import numpy as np

state = np.random.RandomState(51423)
data = 10 + state.normal(0, 1, size=(33, 33)).cumsum(axis=0).cumsum(axis=1)

fig, axs = uplt.subplots([1, 2])

axs[0].pcolor(data, discrete=False, cmap="viridis", colorbar="r", colorbar_kw={"grid": False})
axs[0].grid(False)

axs[1].pcolor(data, discrete=False, cmap="viridis", colorbar="r")
```

![Image](https://github.com/user-attachments/assets/d17aeb84-fdcc-4fed-9f68-215c464f28af)

I manually drew the red rectangles. While the heatmap grid can be disabled using `axs[0].grid(False)`, there appears to be no documented method to remove grid lines from the colorbar. I've tried `colorbar_kw={"grid": False}` but it didn't work. Is this default grid-enabled state for colorbars an intentional design choice of `UltraPlot`? If not, would it be possible to either suppress colorbar grids by default or introduce a parameter like `cbar_grid` for explicit control?

cvanelteren:
Hi again,

I agree that the docs are a bit confusing on this. It mirrors the mpl documentation that `drawedges` controls whether to show the grid lines on the colorbar. They can also be controlled with `grid` or `edges`. The toggle does do something as shown by this example

![Image](https://github.com/user-attachments/assets/b9a9a82c-6125-4006-9109-070a9f087a22)

<details> <summary> Code </summary>

```
import ultraplot as uplt, numpy as np, pandas as pd

data = np.random.rand(10, 10)
fig, ax = uplt.subplots(ncols=2)
for axi, drawedges in zip(ax, [True, False]):
    h = axi.pcolor(data, discrete=True)
    cax = fig.colorbar(h, ax=axi, drawedges=drawedges, cmap="viridis")
    axi.set_title(f"{drawedges=}")
uplt.show(block=1)
```

</details>

I agree that the behavior is somewhat odd, particularly when `discrete = False` is given. 

The grid can be turned off for the colorbar by setting

```
cax.ax.grid(False)
```
yielding:

[...]

PR #150, opened 2025-04-09
Fix minor grid showing on cbar

cvanelteren:
When creating a colorbar with `drawedges=False`, the major grid lines are still visible despite the parameter explicitly requesting that edges not be drawn as addressed by #149 . This inconsistency creates visual artifacts in plots where clean colorbars without grid lines are required. Furthermore, the codebase currently produces inconsistent results with for example pcolor. Providing `colorbar` to pcolor will not show the gridlines by default unless `drawedges` is passed. This PR makes the docs behave consistently for the colorbar.


cvanelteren:
Mpl fails on the added test as expected. Ready for review.

beckermr (changes requested):
I think we always want to always turn off the grid for colorbars. 

In matplotlib, the grid at tick marks and the drawedges keyword appear to me to be separate settings.

The current code in ultraplot (from proplot) merges these two things into one setting via

```
grid = _not_none(
            grid=grid, edges=edges, drawedges=drawedges, default=rc["colorbar.grid"]
        ) 
```

So the intent of the original code appears to have been to only ever allow users to control drawedges for colorbars. 

Given that users were not supposed to be allowed to set grid themselves independent of drawedges, we meed a sensible default for grid, which would be False.

cvanelteren:
Let's go with your  suggestion as I don't see a good reason to allow for a grid on a colorbar.

beckermr (approved):

PR #157, opened 2025-04-12
Minor refactor of unittests

cvanelteren:


beckermr (commented):
This looks good but can you add a PR description explaining why we need to do this? Also, a more automatic way to ensure this happens would be nice. Maybe a pytest fixture that some how wraps `uplt.subplots` such that it always calls close?

cvanelteren:
pytest mpl already does this I believe. But I can add this as a fixture too

Issue #158, opened 2025-04-14
Ticks not responding to locator tags


cvanelteren:
When the labels are set to the right, the ticks cannot move there, see img
![Image](https://github.com/user-attachments/assets/3ee61776-48aa-46b2-8210-b90518a739d1)

```python
import ultraplot as uplt, numpy as np

fig, ax = uplt.subplots(
    nrows=2, ncols=2, proj="cyl", share="all"
)  # , backend="basemap")
ax.format(
    lonlim=(0, 10),
    latlim=(0, 20),
    land=True,
    labels=True,
    lonlines=15,
    latlines=10,
    latticklen=True,
    lonticklen=True,
    grid=True,
    gridminor=False,
    lonlabels="t",
    latlabels="r",
)
uplt.show(block=1)

```

cvanelteren:
Fixed in https://github.com/Ultraplot/ultraplot/releases/tag/v1.5

Issue #155, opened 2025-04-12
How to do axis label sharing with geographic axis?


gepcel:
How to do axis label sharing with geographic axis?

```python
import ultraplot as pplt
fig, axs = pplt.subplots(ncols=2, nrows=2, projection='cyl', refwidth='5cm', share='all')
axs.format(lonlim=(0, 60), latlim=(-20, 20),
           lonlines=15, latlines=10, lonlabels='both', latlabels='both')
```

Like the code above, `share='all'` doesn't seems work.

![Image](https://github.com/user-attachments/assets/564071d2-8224-41c1-b5f4-e9c82c58bca4)

cvanelteren:
You're right—this isn't currently supported and would mainly make sense for rectilinear projects. That said, it’s definitely a feature we could consider adding. You can check out the relevant commit here: [2085c2f1](https://github.com/Ultraplot/ultraplot/commit/2085c2f1d559c499cc20f6b51d49ebb253ba118f). If you're interested, feel free to open a PR to kick things off!

cvanelteren:
Thanks for the suggestion! I took a look, and to support this properly, the `GeoAxis` object would need some extended functionality for shared axes. That would involve patching in methods similar to those used by `CartesianAxes`, while also ensuring all projections involved are rectilinear (and the same). The patching would be possible relatively easily but requires proper testing and disproportionate use for a feature of which I think is rather niche (but do correct me when I am wrong). 

Since this is a fairly niche use case, a quick workaround for now would be to manually disable the labels. I agree it's a worthwhile improvement, and I might revisit this when I have a bit more time to work on it -- otherwise I am willing to review pull requests in the mean time.

cvanelteren:
Update: 

Think I figured something out:

![Image](https://github.com/user-attachments/assets/50e8c449-8726-4c39-9c81-19a7a976c901)

cvanelteren:
@gepcel can you checkout #159

cvanelteren:
![Image](https://github.com/user-attachments/assets/fc932c51-8b0f-42b7-93f5-728c10b21fae)
🥳

cvanelteren:
Yes my bad I need to change the typo

cvanelteren:
@gepcel fixed the typos. Should be good now I think.

cvanelteren:
There are still some bugs to figure with the automatic sharing. I will try to limit some.

cvanelteren:
Note to self: A few things need to happen.

[...]

PR #159, opened 2025-04-14
[Feature add] Share Axes in GeoPlot + bug fixes

cvanelteren:
Addresses #155 and #158

### DIFF
--- a/ultraplot/axes/base.py
+++ b/ultraplot/axes/base.py
@@ -2479,9 +2479,6 @@ class Axes(maxes.Axes):
                 self._shared_axes[which].remove(sibling)
                 if which in "xy":
                     setattr(sibling, f"_share{which}", None)
-
-                sibling.tick_params(left=True, right=True, top=True)
-
                 this_ax = getattr(self, f"{which}axis")
                 sib_ax = getattr(sibling, f"{which}axis")
                 # Reset formatters
--- a/ultraplot/axes/base.py
+++ b/ultraplot/axes/base.py
@@ -2482,10 +2482,8 @@ class Axes(maxes.Axes):
                 this_ax = getattr(self, f"{which}axis")
                 sib_ax = getattr(sibling, f"{which}axis")
                 # Reset formatters
-                this_ax.set_major_locator(mticker.AutoLocator())
-                this_ax.set_major_formatter(pticker.AutoFormatter())
-                this_ax.set_minor_locator(mticker.AutoLocator())
-                this_ax.set_minor_formatter(pticker.AutoFormatter())
+                this_ax.major = copy.deepcopy(this_ax.major)
+                this_ax.minor = copy.deepcopy(this_ax.minor)
 
     def _sharex_setup(self, sharex, **kwargs):
         """
--- a/ultraplot/axes/geo.py
+++ b/ultraplot/axes/geo.py
@@ -1393,11 +1393,9 @@ class _CartopyAxes(GeoAxes, _GeoAxes):
         for dir, side in zip(
             "bottom top left right".split(), [bottom, top, left, right]
         ):
-            print(getattr(self.gridlines_major, f"{dir}_labels"))
             if side != True:
                 continue
             sides[dir] = getattr(self.gridlines_major, f"{dir}_label_artists")
-        print(sides)
         return sides
 
     @staticmethod
@@ -1429,7 +1427,6 @@ class _CartopyAxes(GeoAxes, _GeoAxes):
             _CartopyAxes._get_side_labels()
         )
         gl = self.gridlines_major
-        print(gl._labels)
         if left is not None:
             setattr(gl, left_labels, left)
         if right is not None:
--- a/ultraplot/axes/geo.py
+++ b/ultraplot/axes/geo.py
@@ -13,6 +13,8 @@ except:
     # From Python 3.5
     from typing_extensions import override
 
+from collections.abc import MutableMapping
+
 import matplotlib.axis as maxis
 import matplotlib.path as mpath
 import matplotlib.text as mtext
@@ -596,13 +598,9 @@ class GeoAxes(shared._SharedAxes, plot.PlotAxes):
         label = label.replace("left", "l")
         label = label.replace("right", "r")
         match label:
-            case _ if (
-                len(label) == 2 and "t" in label and "b" in label and which == "x"
-            ):
+            case _ if len(label) == 2 and "t" in label and "b" in label:
                 self.xaxis.set_ticks_position("both")
-            case _ if (
-                len(label) == 2 and "l" in label and "r" in label and which == "y"
-            ):
+            case _ if len(label) == 2 and "l" in label and "r" in label:
                 self.yaxis.set_ticks_position("both")
             case "t":
                 self.xaxis.set_ticks_position("top")
@@ -612,7 +610,10 @@ class GeoAxes(shared._SharedAxes, plot.PlotAxes):
                 self.yaxis.set_ticks_position("left")
             case "r":
                 self.yaxis.set_ticks_position("right")
-            case "all" | "both":
+            case "all":
+                self.xaxis.set_ticks_position("both")
+                self.yaxis.set_ticks_position("both")
+            case "both":
                 if which == "x":
                     self.xaxis.set_ticks_position("both")
                 else:
@@ -661,6 +662,15 @@ class GeoAxes(shared._SharedAxes, plot.PlotAxes):
                 target_axis=self._lataxis,
             )
 
+    def _get_gridliner_labels(
+        self,
+        bottom=None,
+        top=None,
+        left=None,
+        right=None,
+    ):
+        assert NotImplementedError("Should be implemente by Cartopy or Basemap Axes")
+
     def _toggle_gridliner_labels(
         self,
         top=None,
@@ -669,39 +679,23 @@ class GeoAxes(shared._SharedAxes, plot.PlotAxes):
         right=None,
         geo=None,
     ):
-        gl = {}
         # For BasemapAxes the gridlines are dicts with key as the coordinate and  keys the line and label
         # We override the dict here assuming the labels are mut excl due to the N S E W extra chars
         if self.gridlines_major is None:
             return
         if any(i is None for i in self.gridlines_major):
             return
-        # Note: the coordinate keys can be the same;
-        # So we need to assign each a unique key.
-        if left is not None or right is not None:
-            gl["left"] = self.gridlines_major[1]
-        if right is not None:
-            gl["right"] = self.gridlines_major[1]
-        if top is not None:
-            gl["top"] = self.gridlines_major[0]
-        if bottom is not None:
-            gl["bottom"] = self.gridlines_major[0]
-        # Toggle the labels
-        # For Basemap, the labels are encapsulated as a
-        # tuple with the first item being a line (the tick) and
-        # the second item being a list of labels. For example,
-        # a key could be '-20' that has a label attached to it
-        # N or S or E or W.
-        for dir, obj in gl.items():
-            for dir, (line, labels) in obj.items():
-                if left is not None:
-                    labels[0].set_visible(left)
-                if right is not None:
-                    labels[-1].set_visible(right)
-                if top is not None:
-                    labels[0].set_visible(top)
-                if bottom is not None:
-                    labels[-1].set_visible(bottom)
+        gridlabels = self._get_gridliner_labels(
+            bottom=bottom, top=top, left=left, right=right
+        )
+        for direction, toggle in zip(
+            "bottom top left right".split(),
+            [bottom, top, left, right],
+        ):
+            if toggle is not None:
+                for label in gridlabels.get(direction, []):
+                    label.set_visible(toggle)
+        self.stale = True
 
     def _handle_axis_sharing(
         self,
@@ -804,6 +798,10 @@ class GeoAxes(shared._SharedAxes, plot.PlotAxes):
             array = [False] * 5
             opts = ("left", "right", "bottom", "top", "geo")
             for string in strings:
+                string = string.replace("left", "l")
+                string = string.replace("right", "r")
+                string = string.replace("bottom", "b")
+                string = string.replace("top", "t")
                 if string == "all":
                     string = "lrbt"
                 elif string == "both":
@@ -1138,7 +1136,6 @@ class GeoAxes(shared._SharedAxes, plot.PlotAxes):
                 grid_alpha=0,
                 **params,
             )
-
         # Apply tick parameters
         # Move the labels outwards if specified
         if hasattr(gl, f"{x_or_y}padding"):
@@ -1147,6 +1144,7 @@ class GeoAxes(shared._SharedAxes, plot.PlotAxes):
             # For basemap backends, emulate the label placement
             # like how cartopy does this
             self._add_gridline_labels(ax, gl, padding=size)
+
         self.stale = True
 
     def _add_gridline_labels(self, ax, gl, padding=8):
@@ -1158,9 +1156,18 @@ class GeoAxes(shared._SharedAxes, plot.PlotAxes):
         for which, formatter in zip("xy", gl):
             for loc, (lines, labels) in formatter.items():
                 for i, label in enumerate(labels):
-                    upper_end = i == 1
+                    upper_end = True
+                    position = label.get_position()
+
+                    if which == "x":
+                        if position[1] < 0:
+                            upper_end = False
+                    elif which == "y":
+                        if position[0] < 0:
+                            upper_end = False
+                    line = lines[0] if upper_end else lines[-1]
+
                     shift_scale = 1 if upper_end else -1
-                    line = lines[0] if shift_scale == 1 else lines[-1]
                     path = line.get_path()
                     vertices = path.vertices
                     label.set_transform(line.get_transform())
@@ -1359,6 +1366,43 @@ class _CartopyAxes(GeoAxes, _GeoAxes):
         gl._draw_gridliner = _draw_gridliner.__get__(gl)
         gl.xlines = gl.ylines = False
         return gl
+
+    @override
+    def _get_gridliner_labels(
+        self,
+        bottom=None,
+        top=None,
+        left=None,
+        right=None,
+    ) -> dict[str, list[mtext.Text]]:
+        sides = {}
+        for label, side in zip(
+            "bottom top left right".split(), [bottom, top, left, right]
+        ):
+            if not side:
+                continue
+            if label in "bottom top".split():
+                labels = self.xlabel_artists
+            else:
+                labels = self.ylabel_artists
+
+            # Labels are aligned in a typical manner
+            # Horizontal alignment is adjusted for left and right.
+            # Similarly vertical alignment is adjusted for top
+            # and bottom
+            for label in labels:
+                target = None
+                if label.get_horizontalalignment() == "right":
+                    target = "left"
+                elif label.get_horizontalalignment() == "left":
+                    target = "right"
+                elif label.get_verticalalignment() == "bottom":
+                    target = top
+                elif label.get_verticalalignment() == "top":
+                    target = bottom
+                if target:
+                    sides[target] = sides.get(target, []) + [label]
+        return sides
 
     @staticmethod
     def _get_side_labels() -> tuple:
@@ -1642,8 +1686,8 @@ class _CartopyAxes(GeoAxes, _GeoAxes):
         )
         gl.xformatter = self._lonaxis.get_major_formatter()
         gl.yformatter = self._lataxis.get_major_formatter()
-        self.xaxis.set_major_formatter(mticker.NullFormatter())
-        self.yaxis.set_major_formatter(mticker.NullFormatter())
+        # self.xaxis.set_major_formatter(mticker.NullFormatter())
+        # self.yaxis.set_major_formatter(mticker.NullFormatter())
 
         # Update gridline label parameters
         # NOTE: Cartopy 0.18 and 0.19 can not draw both edge and inline labels. Instead
@@ -1881,7 +1925,7 @@ class _BasemapAxes(GeoAxes):
                 if isinstance(object, list) and len(object) > 0:
                     object = object[0]
                 if isinstance(object, mtext.Text):
-                    object.set_visible(True)
+                    object.set_visible(False)
 
     def _get_lon0(self):
         """
@@ -2129,6 +2173,67 @@ class _BasemapAxes(GeoAxes):
             axis.isDefault_majfmt = True
             axis.isDefault_majloc = True
             axis.isDefault_minloc = True
+
+    @override
+    def _get_gridliner_labels(
+        self,
+        bottom=None,
+        top=None,
+        left=None,
+        right=None,
+    ):
+        # For basemap object, the text is organized
+        # as a dictionary. The keys are the numerical
+        # location values, and the values are a list
+        # where the version item is the tick and the
+        # the rest are mtext.Text objects. The labels
+        # are clustereed on the location per axis.
+        # This means that top and bottom labels are assigned
+        # to the same numerical loc.
+        # We therefore create a mapping per direction to make
+        # it more semantically logical.
+        def group_labels(
+            labels: list[mtext.Text],
+            which: str,
+            bottom=None,
+            top=None,
+            left=None,
+            right=None,
+        ) -> dict[str, list[mtext.Text]]:
+            group = {}
+            # We take zero here as a baseline
+            for label in labels:
+                position = label.get_position()
+                target = None
+                if which == "x":
+                    if bottom is not None and position[1] < 0:
+                        target = "bottom"
+                    elif top is not None and position[1] >= 0:
+                        target = "top"
+                else:
+                    if left is not None and position[0] < 0:
+                        target = "left"
+                    elif right is not None and position[0] >= 0:
+                        target = "right"
+                if target is not None:
+                    group[target] = group.get(target, []) + [label]
+            return group
+
+        # Group the text object based on their location
+        grouped = {}
+        for which, gl in zip("xy", self.gridlines_major):
+            for loc, (line, labels) in gl.items():
+                tmp = group_labels(
+                    labels=labels,
+                    which=which,
+                    bottom=bottom,
+                    top=top,
+                    left=left,
+                    right=right,
+                )
+                for key, values in tmp.items():
+                    grouped[key] = grouped.get(key, []) + values
+        return grouped
 
 
 # Apply signature obfuscation after storing previous signature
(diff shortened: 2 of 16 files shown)


cvanelteren:
Note: Tests will fail on added tests.

cvanelteren:
Will add more tests tomorrow for codecov but it's looking good locally with a few plots showing different sizes but otherwise have high visual fidelity. Opening this up for review.

cvanelteren:
Need to add:
- unittest for the unsharing of axes
- sort the newly added functions by name
- document changes

cvanelteren:
PR looking good will add the remaining points in the next days.

cvanelteren:
Needs rebase befor merging(!)

cvanelteren:
This one is also ready for review @beckermr

ultraplot/axes/base.py
```diff
@@ -2336,6 +2336,40 @@ def _range_tightbbox(self, s):
...
+        # I think it is fine to leave it as otherwise we would need to determine the scale, which may get messy.
+
+        grouper = self._shared_axes[which]
+        siblings = list(grouper.get_siblings(self))
+        for sibling in siblings:
+            if sibling is not self:
+                # Unshare by removing them from the grouper
+                grouper.remove(sibling)
+                sibling._shared_axes[which].remove(self)
+                # To be save let's remove this
```
beckermr, line 2358:
```suggestion
                # To be safe let's remove this
```

ultraplot/axes/shared.py
```diff
@@ -210,7 +210,9 @@ def _share_axis(self, which, other):
         other_axis = getattr(other, f"{which}axis")
 
         # Set minor ticker
-        this_axis.minor = other_axis.minor
+        this_axis.set_minor_locator(other_axis.get_minor_locator())
+        this_axis.set_minor_formatter(other_axis.get_minor_formatter())
+        # this_axis.minor = other_axis.minor
```
beckermr, line 215: Let's remove dead code.

ultraplot/figure.py
```diff
@@ -1131,11 +1126,83 @@ def _add_subplot(self, *args, **kwargs):
...
+        self._WARN_SHARING_GEOAXIS = True
+        # Allow sharing for GeoAxes if rectilinear
+        if self._sharex or self._sharey:
+            if (
+                len(self.axes) > 1
+                and isinstance(ax, paxes.GeoAxes)
+                and not ax._is_rectilinear()
+            ):
+                self._unshare_axes()
+                if self._WARN_SHARING_GEOAXIS:
```
beckermr, line 1140: We should be using the [warnings filter `once`](https://docs.python.org/3/library/warnings.html#the-warnings-filter) here instead of rolling our own global. 

The code as written warns once per object, not once per class as well.

ultraplot/figure.py
```diff
@@ -1131,11 +1126,83 @@ def _add_subplot(self, *args, **kwargs):
...
+        children=False,
+        hidden=False,
+    ):
+        """
+        Share or unshare axes in the figure along a given direction.
+
+        Parameters:
+        - which: 'x', 'y' or 'view'.
+        - share: int indicating the levels (see above)
+        - panels: Whether to include panel axes.
```
beckermr, line 1178: Not all of the parameters are documented.

ultraplot/tests/test_axes.py
```diff
@@ -187,3 +187,22 @@ def test_subset_format():
...
         axs.format(title=["a"])
+
+
+def test_unsharing():
+    """
+    Test some basic properties of unsharing axes.
+    """
+    fig, ax = uplt.subplots(ncols=2)
+    # Does nothing since key is not an axis or a view
+    ax[0]._unshare(which="key does not exist")
```
beckermr, line 198: Should thjis raise on error instead of silently failing?

ultraplot/tests/test_geographic.py
```diff
@@ -162,13 +162,54 @@ def test_geoticks():
...
+        grid=True,
+        gridminor=False,
+        lonlabels=lonlabels,
+        latlabels=latlabels,
+    )
+    return fig
+
+
+def test_geoticks_shared_non_rectilinear():
+    with pytest.warns(plt.warnings.UltraplotWarning):
```
beckermr, line 206: Is it `UltraPlotWarning` or `UltraplotWarning`? 

I see both in the code in various spots (code, doc strings, etc.) and we should stick to one or the other.

These differences caused a bug in https://github.com/Ultraplot/ultraplot/pull/190 as well.

beckermr (changes requested):
I left some comments. The test failures look like they are not supposed to happen and need addressing.

ultraplot/figure.py
cvanelteren: Ah thanks was not aware of this.

ultraplot/tests/test_geographic.py
cvanelteren: It should be UltraPlot warning but the changes in #69  are not merged yet.

ultraplot/tests/test_axes.py
cvanelteren: We can raise a warning

ultraplot/figure.py
cvanelteren: Wouldn't this filter all warnings once?

cvanelteren: This does not happen. Removing the variable

ultraplot/figure.py
cvanelteren: Thanks

ultraplot/tests/test_axes.py
cvanelteren: You think we need a test?

beckermr: Yes we should test.

ultraplot/tests/test_geographic.py
beckermr: Ah great. Thanks for clarifying.

cvanelteren: It indeed interferes with pytest-xdist I noticed.

cvanelteren: #69 is a bit messy right now but I will either split them or prioritize it now since the other PRs are depending on this one too.

ultraplot/figure.py
```diff
@@ -1201,11 +1202,82 @@ def _add_subplot(self, *args, **kwargs):
...
+
         if ax.number:
             self._subplot_dict[ax.number] = ax
         return ax
 
+    def _unshare_axes(self):
+        for which in "x y z".split():
+            self._toggle_axis_sharing(which=which, share=False)
+        # Force setting extent
+        # This is necessary to ensure that the axes are properly aligned and we don't get weird scaling issues for geographic axes. This action is expensive for GeoAxes
```
beckermr, line 1231:
```suggestion
        # This is necessary to ensure that the axes are properly aligned
        # and we don't get weird scaling issues for geographic axes.
        # This action is expensive for GeoAxes
```

beckermr (commented):
one more long comment

beckermr:
Merge when you are ready!

beckermr:
The test failure here results in an axis not receiving tick marks. Is that correct?

Screenshot 2025-04-30 at 6 04 49 AM

cvanelteren:
No. I also mean to fix the missing tests. This is not ready yet.

cvanelteren:
Added the labels back in but the offset calculation is a bit funky. Currently, it is removed for both backends. The basemap backend has no direct access to the coordinate reference system so I cannot make the plots look equal for both cartopy and basemap in a trivial manner. 

Luke initially deprecated the basemap backend. I thought that since development picked for basemap it would be nice to support it. However, the development over there has stalled -- so now we are left at the same question: do we maintain two backends or not?

beckermr:
What features does basemap provide that the other backend does not?

cvanelteren:
Good question. Not too sure actually. Will get back to you on this.

cvanelteren:
The way I understand it as a non-geo scientist, is that cartopy is the future. I think it has more legacy support supposedly having easier ticker manager. Basemap is supposed to be native to matplotlib, but the development over the years have stalled and is less well supported in modern matplotlib. It is currently also halting our ability to upgrade to matplotlib >=3.10 as their numpy upgrades are not live yet on basemap.

beckermr:
I'm fine with removing it. We should stop the back and forth on whether or not we are going to remove it and just go with something.

cvanelteren:
I'll put a deprecation warming up. I will keep it available in this pr.

cvanelteren:
Also addresses #204 point 1 now.

ultraplot/constructor.py
```diff
@@ -1107,6 +1107,8 @@ def Locator(locator, *args, discrete=False, **kwargs):
     ultraplot.axes.Axes.colorbar
     ultraplot.constructor.Formatter
     """  # noqa: E501
+
+    lon0 = kwargs.pop("lon0", None)
```
beckermr, line 1111: Is it true that the geo axes never use this factor function for building locators?

beckermr: I see calls here 

https://github.com/Ultraplot/ultraplot/blob/ae0029761b8e0160acfc38277c611c21fd8d84dd/ultraplot/axes/geo.py#L335

and so am confused as to how this is going to result in correct results.

cvanelteren: Yes. Just removed it.  It is correct since I am setting the lon0 attribute in 1641. Passing it along it not necessary. Not very happy with the implementation but it is not clear to me exactly how the formatters are generated.

cvanelteren: I would rather push it to the `_LonAxis` class but the that would require some tooling on how `lon0` is retrieved.

cvanelteren:
Don't really understand the artifact created now in one of the tests. It should be tackled by the logic in 315 of geo.py. What's also funky is that when I put an assert on __call__ from the formatter it is never executed (!) but when I change the logic it does propagate...confusing

cvanelteren:
Found the issue  I wasn't setting the formatters -_-

cvanelteren:
Will add some tests in the next few days an reping for a review then we can proceed to merge.

cvanelteren:
I think this PR is ready for merge. I think I did my due diligence providing the proper test for testing axis sharing and unsharing. This PR took a lot more work than I thought. It adds features such as axis sharing and unsharing. Please note that this feature is discouraged by matplotlib. The PR in addition fixes numerous bugs for basemap.

ultraplot/tests/test_projections.py
```diff
@@ -134,3 +134,49 @@ def test_polar_projections():
...
+            assert np.allclose(L, 0)
+
+
+def test_sharing_axes_different_projections():
+    """
+    Test sharing axes for GeoAxes
+    """
+
+    # For rectilinear plots all axes can be shared
+    #
```
beckermr, line 165:
```suggestion
```

ultraplot/tests/test_projections.py
```diff
@@ -134,3 +134,49 @@ def test_polar_projections():
         rlines=0.25,
     )
     return fig
+
+
+def test_sharing_axes():
```
beckermr, line 139: We should assert that we get no warnings here.

ultraplot/axes/base.py
```diff
@@ -2454,6 +2454,39 @@ def _range_tightbbox(self, s):
...
+                self._shared_axes[which].remove(sibling)
+                if which in "xy":
+                    setattr(sibling, f"_share{which}", None)
+                this_ax = getattr(self, f"{which}axis")
+                sib_ax = getattr(sibling, f"{which}axis")
+                # Reset formatters
+                this_ax.set_major_locator(mticker.AutoLocator())
+                this_ax.set_major_formatter(pticker.AutoFormatter())
+                this_ax.set_minor_locator(mticker.AutoLocator())
+                this_ax.set_minor_formatter(pticker.AutoFormatter())
```
beckermr, line 2488: We should make a deepcopy of the current formatter for the axis here instead? 

IIUIC, that way any custom changes to the formatting would be preserved?

ultraplot/figure.py
```diff
@@ -1201,11 +1202,116 @@ def _add_subplot(self, *args, **kwargs):
...
+                warnings._warn_ultraplot(
+                    f"GeoAxes can only be shared for rectilinear projections, {ax.projection=} is not a rectilinear projection."
+                )
+
         if ax.number:
             self._subplot_dict[ax.number] = ax
         return ax
 
+    def _unshare_axes(self):
+        for which in "x y z".split():
```
beckermr, line 1228:
```suggestion
        for which in "xyz":
```

This works fine:

```python
In [2]: for b in "xyz":
   ...:     print(b)
   ...: 
x
y
z
```

ultraplot/figure.py
```diff
@@ -1201,11 +1202,116 @@ def _add_subplot(self, *args, **kwargs):
...
+        share=True,
+        panels=False,
+        children=False,
+        hidden=False,
+    ):
+        """
+        Share or unshare axes in the figure along a given direction.
+
+        Parameters:
+        - which: 'x', 'y' or 'view'.
```
beckermr, line 1251:
```suggestion
        - which: 'x', 'y', 'z', or 'view'.
```

ultraplot/tests/test_figure.py
```diff
@@ -0,0 +1,115 @@
...
+        for ref, axi in zip(ref_ax, ax):
+            for which in "xy":
+                ref_index = getattr(ref, f"_share{which}")
+                axi_index = getattr(ref, f"_share{which}")
+                if ref_index is None:
+                    assert ref_index == axi_index
+                else:
+                    assert ref_index.number == axi_index.number
+                    ref_axis = getattr(ref, f"{which}axis")
+                    axis = getattr(axi, f"{which}axis")
```
beckermr, line 92: This else block is missing actual tests if the axes' properties.

ultraplot/tests/test_geographic.py
```diff
@@ -164,9 +164,181 @@ def test_geoticks_input_handling(recwarn):
     # Should warn that about non-rectilinear projection.
     with pytest.warns(uplt.warnings.UltraPlotWarning):
         ax.format(lonticklen=True)
-    # When set to None the latticks are not added.
-    # No warnings should be raised.
     ax.format(lonticklen=None)
     assert len(recwarn) == 0
+    # When set to None the latticks are not added.
+    # No warnings should be raised.
```
beckermr, line 170: This comment is out of place and needs to be moved back.

ultraplot/tests/test_geographic.py
```diff
@@ -164,9 +164,181 @@ def test_geoticks_input_handling(recwarn):
...
+
+    # All label are off
+    for gli in gl:
+        for _, (line, labels) in gli.items():
+            for label in labels:
+                assert label.get_visible() == False
+
+    # Should be off
+    ax[0]._toggle_gridliner_labels(top=True)
+    # Gridliner labels are not added for the top (and I guess right for GeoAxes). Need to figure out how this is set is set in matplotlib
```
beckermr, line 334:
```suggestion
    # Gridliner labels are not added for the top (and I guess right for GeoAxes).
    Need to figure out how this is set in matplotlib
```

ultraplot/tests/test_geographic.py
```diff
@@ -164,9 +164,181 @@ def test_geoticks_input_handling(recwarn):
...
+    dir_labels = ax[0]._get_gridliner_labels(
+        left=True, right=True, top=True, bottom=True
+    )
+    for dir, labels in dir_labels.items():
+        expectation = False
+        if dir == "top":
+            expectation = True
```
beckermr, line 341: This line of code is missed in code coverage which means I think that this test is not quite working correctly.

ultraplot/axes/geo.py
```diff
@@ -480,6 +501,255 @@ def __init__(self, *args, **kwargs):
         """
         super().__init__(*args, **kwargs)
 
+    @override
+    def _sharey_limits(self, sharey: "GeoAxes"):
+        return self._share_limits(sharey, which="y")
```
beckermr, line 506: The lack of code coverage here means we need an additional test or there is a bug.

ultraplot/axes/geo.py
```diff
@@ -480,6 +501,255 @@ def __init__(self, *args, **kwargs):
...
+
+    @override
+    def _sharex_limits(self, sharex: "GeoAxes"):
+        return self._share_limits(sharex, which="x")
```
beckermr, line 510: The lack of code coverage here means we need an additional test or there is a bug.

ultraplot/axes/geo.py
```diff
@@ -480,6 +501,255 @@ def __init__(self, *args, **kwargs):
...
+            "major_locator",
+            "minor_locator",
+            "major_formatter",
+        ]
+        for prop, func in zip(props, funcs):
+            if getattr(other_ax, prop) and not getattr(this_ax, prop):
+                setter = getattr(other_ax, f"set_{func}")
+                getter = getattr(this_ax, f"get_{func}")
+                if setter and getter:
+                    setter(getter())
```
beckermr, line 540: Again missing coverage in these new functions points to incomplete tests or a bug.

ultraplot/axes/geo.py
```diff
@@ -1290,8 +1686,8 @@ def _update_major_gridlines(
         )
         gl.xformatter = self._lonaxis.get_major_formatter()
         gl.yformatter = self._lataxis.get_major_formatter()
-        self.xaxis.set_major_formatter(mticker.NullFormatter())
-        self.yaxis.set_major_formatter(mticker.NullFormatter())
+        # self.xaxis.set_major_formatter(mticker.NullFormatter())
+        # self.yaxis.set_major_formatter(mticker.NullFormatter())
```
beckermr, line 1690: If this code is not needed, it should be removed. Otherwise, we need to add it back.

beckermr (changes requested):
This PR is looking great!

The test coverage in `geo.py` seems to not cover key bits of the logic, certain branches, etc. We should improve on that before merging.

beckermr:
Screenshot 2025-05-08 at 6 48 42 AM

Is the image comparison failure above correct or is there another bug that needs fixing?

ultraplot/figure.py
cvanelteren: Habbit of mine ;-) thanks

ultraplot/axes/base.py
cvanelteren: Ah right. I believe I had this originally but moved this to auto. I think I ran into some snag but will reset it to see if it works.

cvanelteren: Your suggestion is the way to go I think.

ultraplot/tests/test_geographic.py
cvanelteren: Thanks! Fixed. Apparently the labels need to be toggled first before they can be turned off. Will investigate this as it is not intended. Keeping it open for now, will fix the other stuff first.

beckermr:
Let's not mark comments as resolved until the commits that resolve them are pushed to the PR.

ultraplot/axes/base.py
cvanelteren: Fixed in a latest commit.

cvanelteren:
Putting this in draft until I can figure the bug out.

ultraplot/axes/geo.py
cvanelteren: Fixed!

ultraplot/axes/geo.py
cvanelteren: Fixed!

ultraplot/axes/geo.py
cvanelteren: Fixed!

ultraplot/tests/test_geographic.py
```diff
@@ -164,9 +165,252 @@ def test_geoticks_input_handling(recwarn):
     # Should warn that about non-rectilinear projection.
     with pytest.warns(uplt.warnings.UltraPlotWarning):
         ax.format(lonticklen=True)
-    # When set to None the latticks are not added.
-    # No warnings should be raised.
     ax.format(lonticklen=None)
     assert len(recwarn) == 0
+    # When set to None the latticks are not added.
+    # No warnings should be raised.
```
beckermr, line 171:
```suggestion
    # When set to None the latticks are not added.
    # No warnings should be raised.
    ax.format(lonticklen=None)
    assert len(recwarn) == 0
```

beckermr (changes requested):
There are still some sections of code that look pretty critical but have no test coverage at all. 

Do we have a test that tries different sharing levels besides no sharing and sharing everything?

That test case seems pertinent and might help address some of the coverage issues.

beckermr:
Also the tests still fail, with the middle panel of the geoticks test having a different y range.

Screenshot 2025-05-08 at 7 33 19 PM

cvanelteren:
> There are still some sections of code that look pretty critical but have no test coverage at all.
> 
> Do we have a test that tries different sharing levels besides no sharing and sharing everything?
> 
> That test case seems pertinent and might help address some of the coverage issues.

Ok will strive for 100 percent coverage. Thought this was "good enough".

beckermr:
> > There are still some sections of code that look pretty critical but have no test coverage at all.
> > Do we have a test that tries different sharing levels besides no sharing and sharing everything?
> > That test case seems pertinent and might help address some of the coverage issues.
> 
> Ok will strive for 100 percent coverage. Thought this was "good enough".

Ah I just realized that the missing coverage is due to some tests not being run with basemap. We probably do not care about this. 

Let's add a test with different sharing levels and call it good.

cvanelteren:
There are still some tests I need to add, will ping you when ready.

beckermr:
I took another close look at the image comparison test failure and it is definitely a bug. 

Here is a zoom in on the middle panels, with the result of this PR on the bottom and the baseline on the top:

Screenshot 2025-05-10 at 6 20 16 AM

The y-axis labels are indeed the same, but notice that the **continent outlines differ**. The same location on the earth cannot be at two different latitudes. 

Somehow with this PR, the content of the middle panel does not match the applied axis labels and this is a bug for sure.

beckermr (changes requested):
I took a closer look at the image test failure:

Here is the code:

```
@pytest.mark.mpl_image_compare
def test_geoticks():

    lonlim = (-140, 60)
    latlim = (-10, 50)
    basemap_projection = uplt.Proj(
        "cyl",
        lonlim=lonlim,
        latlim=latlim,
        backend="basemap",
    )
    fig, ax = uplt.subplots(
        ncols=3,
        proj=(
            "cyl",  # cartopy
            "cyl",  # cartopy
            basemap_projection,  # basemap
        ),
        share=0,
    )
    settings = dict(land=True, labels=True, lonlines=20, latlines=20)
    # Shows sensible "default"; uses cartopy backend to show the grid lines with ticks
    ax[0].format(
        lonlim=lonlim,
        latlim=latlim,
        **settings,
    )

    # Add lateral ticks only
    ax[1].format(
        latticklen=True,
        gridminor=True,
        lonlim=lonlim,
        latlim=latlim,
        **settings,
    )

    ax[2].format(
        latticklen=5.0,
        lonticklen=2.0,
        grid=False,
        gridminor=False,
        **settings,
    )
    return fig
```

Here is a zoom of the middle panel (top is baseline, bottom is this PR):

Screenshot 2025-05-10 at 6 20 16 AM

It appears that the call to set `latlim` for the middle panel is now being ignored in the new version of the code in this PR. This is definitely a bug. I don't see why latlim should now be ignored.

cvanelteren:
Not sure why the ticks are not using the full range but I cut off the edges here and it looks the same.

cvanelteren:
weirdly enough when I do this for the basemap the yticks behave strangely. Only one appears.

ultraplot/axes/geo.py
```diff
@@ -825,35 +1122,109 @@ def _add_geoticks(self, x_or_y, itick, ticklen):
...
+            if x_or_y == "x":
+                lim = self._lonaxis.get_view_interval()
+                locator = gl.xlocator
+                tick_positions = self._lonaxis._get_ticklocs(locator)[1:-1]
+            else:
+                lim = self._lataxis.get_view_interval()
+                locator = gl.ylocator
+                tick_positions = self._lataxis._get_ticklocs(locator)[
+                    1:-1
+                ]  # cutoff edges
```
beckermr, line 1139: Why are we truncating the edges by hand here? This is a hack to get a test to pass and not something that generalizes.

beckermr (changes requested):

ultraplot/axes/geo.py
cvanelteren: Not sure will do some more digging as to why this is caused. I am setting the tick locs specifically, which has the range in the pic, but the first subplot (without the tics) clearly has a reduced range.

cvanelteren: please note that the reduced range is the target range that needs to be set so it is indeed a bug.

cvanelteren: Think I figured it out. The update extent method is called before the ticks are set. If I call it again if we add the ticks, it looks good for cartopy. For some reason the x labels are treated differently for basemap but not sure if that is how it should look. It now looks like:

cvanelteren: This way we don't have to manually trunk it

ultraplot/tests/test_geographic.py
```diff
@@ -170,3 +171,411 @@ def test_geoticks_input_handling(recwarn):
...
+    with pytest.warns(uplt.warnings.UltraPlotWarning):
+        fig, ax = uplt.subplots(ncols=2, proj="aeqd", share="all")
+        ax.format(
+            land=True,
+            labels=True,
+            lonlabels="all",
+            latlabels="all",
+        )
+        fig.canvas.draw()  # draw is necessary to invoke the warning
+    return fig
```
beckermr, line 214: This test needs to assert the warning was raise. It does not need to return the fig instance and it probably should close it.

beckermr (changes requested):
Two more items. Thank you so much for the good work on this PR!

ultraplot/tests/test_geographic.py
cvanelteren: You are right about the return although we do have a fixture that closes all figures after each test. I do not follow your first comment though, the `with` block fails if the warning is not given.

cvanelteren: I removed the closing, in the latest commit. 

I also removed the dead code in the first issue raised. I was first thinking of comparing that the axes are the same but the axes are different objects. Otherwise the tests needs to be expanded. I think checking that that the references of the axes are good enough for now. That is, we now check that the axes being shared matches the same structure as the reference, but we do not check if the properties change with it. This should be covered by other tests however so  I think it is fine as is, but we could add this if that is better.

cvanelteren: Added a view test in the latest commit

beckermr (approved):
LGTM! Great work on this PR!

cvanelteren:
Will draft an appropriate release to 1.5. Lots of new features :smile:

Issue #160, opened 2025-04-16
Rotated x-axis tick labels are misaligned with xticks


ToryDeng:
When rotating x-axis tick labels using `axs[1].format(xrotation=angle)`, the text labels are not properly aligned with their corresponding tick marks. To illustrate this issue, I set `xrotation=89` in the first subplot and set `xrotation=90` in the second subplot:

```python
import ultraplot as uplt
import numpy as np
import pandas as pd

state = np.random.RandomState(51423)
data = state.normal(size=(3, 3)).cumsum(axis=0)
data = pd.DataFrame(data, columns=["aaa", "bbb", "ccc"])

fig, axs = uplt.subplots(nrows=1, ncols=2, refwidth=4.5)
axs[0].heatmap(data,cmap="ColdHot")
axs[0].format(xrotation=89)
axs[1].heatmap(data,cmap="ColdHot")
axs[1].format(xrotation=90)
```

![Image](https://github.com/user-attachments/assets/1611de94-a3f2-4f73-b68d-b17f50e64673)

cvanelteren:
Not sure if a bug or a feature. The logic can be found in `cartesian.py` in line 802. The issue is that the anchor is not shifted. Which works well for say 45 degree shifted but not well for these edge cases where you have close to 90.

```python
   def _update_rotation(self, s, *, rotation=None):
        """
        Rotate the tick labels. Rotate 90 degrees by default for datetime *x* axes.
        """
        # Apply rotation for datetime axes.
        # NOTE: Rotation is done *before* horizontal/vertical alignment. Cannot
        # change alignment with set_tick_params so we must apply to text objects.
        # Note fig.autofmt_date calls subplots_adjust, so we cannot use it.
        current = f"_{s}axis_current_rotation"
        default = f"_{s}axis_isdefault_rotation"
        axis = getattr(self, f"{s}axis")
        if rotation is not None:
            setattr(self, default, False)
        elif not getattr(self, default):
            return  # do not rotate
        elif s == "x" and isinstance(axis.converter, DATE_CONVERTERS):
            rotation = rc["formatter.timerotation"]
        else:
            rotation = "horizontal"

        # Apply tick label rotation if necessary
        if rotation != getattr(self, current):
            rotation = {"horizontal": 0, "vertical": 90}.get(rotation, rotation)
            kw = {"rotation": rotation}
            if rotation not in (0, 90, -90):
                kw["ha"] = "right" if rotation > 0 else "left"
[...]

PR #161, opened 2025-04-16
Make anchor_mode default

cvanelteren:
Addresses #160


beckermr (changes requested):
We should add a test that makes a plot and ensures the ticks are anchored properly.

beckermr (approved):

Issue #162, opened 2025-04-17
trying non-existent color appears to break Ultraplot until kernel restart?


Thomas-Moore-Creative:
**WHAT:** very unusual behaviour.  Working Geographic code cell no long works after getting an error because of wrong color string?  Code will run again without error after a kernel restart.
**WHERE**: interactive Jupyter lab on a MacOS laptop

Hi, thanks for picking up where ProPlot left off.  The open source development is appreciated!

I'm just trialing UltraPlot for the first time and was making some VERY simple Geographic plots.

This code cell works fine:
```
fig = uplt.figure(refwidth=6)
axs = fig.subplots(nrows=2, proj="robin", proj_kw={"lon0": 150})
axs.format(
    suptitle="Testing UltraPlot Cartopy",
    land=True,
    landcolor='mushroom',
    coast=True,
    coastcolor='black',
    coastlinewidth=0.6,
    latlines=30,
    lonlines=30
)
```
![Image](https://github.com/user-attachments/assets/c0d62fc3-e684-4f8e-a03c-cd6085f7a8bd)

If I change a color to something erroneous, like `coastcolor='BLAH',` I get an expected `ValueError: 'BLAH' is not a valid color arg.`

HOWEVER, if I change the color back to the same 'black' string the code now throws a `KeyError`.  This can only be fixed if I restart the kernel.

![Image](https://github.com/user-attachments/assets/a405421e-079e-46e0-ab17-2fb8e166af1c)

This is probably some kind of user error on my part but it has me stumped!

[ full error dump below ]
```
fig = uplt.figure(refwidth=6)
axs = fig.subplots(nrows=2, proj="robin", proj_kw={"lon0": 150})
axs.format(
    suptitle="Testing UltraPlot Cartopy",
    land=True,
    landcolor='mushroom',
    coast=True,
    coastcolor='black',
    coastlinewidth=0.6,
    latlines=30,
    lonlines=30
)

---------------------------------------------------------------------------
KeyError                                  Traceback (most recent call last)
Cell In[12], line 2
      1 fig = uplt.figure(refwidth=6)
----> 2 axs = fig.subplots(nrows=2, proj="robin", proj_kw={"lon0": 150})
      3 axs.format(
      4     suptitle="Testing UltraPlot Cartopy",
      5     land=True,
   (...)
     11     lonlines=30
     12 )

[...]

PR #164, opened 2025-04-18
Ipy rc kernel reset

cvanelteren:
Addresses #162


cvanelteren:
Will fix later today

cvanelteren:
Seeing an issue where the catch does not yield expected results in jupyter lab

Issue #166, opened 2025-04-19
Error when setting titles for a subset of axes using a list of titles


ToryDeng:
Like the issue title said. Here is a reproducible example:

```python
fig, axs = uplt.subplots(nrows=1, ncols=3)
axs[1:].format(title=["a", "b"])
```

```text
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
Cell In[73], line 2
      1 fig, axs = uplt.subplots(nrows=1, ncols=3)
----> 2 axs[1:].format(title=["a", "b"])

File /usr/local/miniforge3/envs/experiment/lib/python3.12/site-packages/ultraplot/gridspec.py:1630, in SubplotGrid.format(self, **kwargs)
   1600 @docstring._snippet_manager
   1601 def format(self, **kwargs):
   1602     """
   1603     Call the ``format`` command for the `~SubplotGrid.figure`
   1604     and every axes in the grid.
   (...)   1628     ultraplot.config.Configurator.context
   1629     """
-> 1630     self.figure.format(axs=self, **kwargs)

File /usr/local/miniforge3/envs/experiment/lib/python3.12/site-packages/ultraplot/internals/warnings.py:118, in _rename_kwargs.<locals>._decorator.<locals>._deprecate_kwargs_wrapper(*args, **kwargs)
    113         key_new = key_new.format(value)
    114     _warn_ultraplot(
    115         f"Keyword {key_old!r} was deprecated in version {version} and may "
    116         f"be removed in {next_release()}. Please use {key_new!r} instead."
    117     )
--> 118 return func_orig(*args, **kwargs)

File /usr/local/miniforge3/envs/experiment/lib/python3.12/site-packages/ultraplot/figure.py:1694, in Figure.format(self, axs, figtitle, suptitle, suptitle_kw, llabels, leftlabels, leftlabels_kw, rlabels, rightlabels, rightlabels_kw, blabels, bottomlabels, bottomlabels_kw, tlabels, toplabels, toplabels_kw, rowlabels, collabels, includepanels, **kwargs)
   1687 for ax in axs:
   1688     kw = {
   1689         key: value
   1690         for cls, kw in kws.items()
   1691         for key, value in kw.items()
   1692         if isinstance(ax, cls) and not classes.add(cls)
   1693     }
-> 1694     ax.format(rc_kw=rc_kw, rc_mode=rc_mode, skip_figure=True, **kw, **kwargs)
   1696 # Warn unused keyword argument(s)
[...]

PR #165, opened 2025-04-18
make cbar labelloc possible for all direction

cvanelteren:
Addresses #163

PR #168, opened 2025-04-22
Refactor colorbar handling into class

cvanelteren:
# Why this PR?

The codebase currently contains mainy functions that have grown naturally over time. They are getting to a point where maintaining them because cumbersome. We start by moving the colorbar handling into a proper class and removing the monkey patches.

This is following #165 and #166


cvanelteren:
Closing this PR in favor of a new one.

PR #169, opened 2025-04-22
Add network plotting to UltraPlot

cvanelteren:
# Why this PR?

Networks are the base of many disciplines, and visualizing them is an important part of understanding how "intuitively" graphs differ from one another. A pet peeve of mine is that network plotting is verbose or too opaque with poor chosen defaults, preventing clear communication of the intended purpose: highlighting connections.

This PR presents an interface for plotting networkx graphs. It provides sensible defaults and makes beautiful default plots while still allowing access to changes when needed. Networkx has grown to be the often de facto standard for visualizing and base analysis of small to medium networks in python. It has a nice drawing interface that is unfortunately a bit verbose. By drawing on the power of networkx, we can enhance the plotting experience by providing a tight interface with networkx and matplotlib, while preventing the verbose of the former.  Other backends could be added (such as networkit, graphtool and others) but for now networkx is supported.

### DIFF
--- a/docs/networks.py
+++ b/docs/networks.py
@@ -1,0 +1,79 @@
+# -*- coding: utf-8 -*-
+# ---
+# jupyter:
+#   jupytext:
+#     text_representation:
+#       extension: .py
+#       format_name: percent
+#       format_version: '1.3'
+#       jupytext_version: 1.11.4
+#   kernelspec:
+#     display_name: Python 3
+#     language: python
+#     name: python3
+# ---
+# %% [raw] raw_mimetype="text/restructuredtext"
+# Networks
+# ========
+
+# Visualizing Connections: Graphs!
+# ================================
+# Networks form a core aspect of any disciplines of science from engineering to biology. UltraPlot supports plotting networks using `networkx <https://networkx.org/>`__.  It provides an intuitive interface for plotting networks using :func:`~ultraplot.axes.PlotAxes.graph` to plot networks. Plot customization can be passed to the networkx backend, see for full details their documentation.
+# Layouts can be passed using strings, dicts or functions as long as they are compatible with networkx's layout functions.
+
+# Minimal Example
+# ===============
+# To plot a graph, use :func:`~ultraplot.axes.PlotAxes.graph`. You need to merely provide a graph and ultraplot will take care of the rest. UltraPlot will automatically style the layout to give sensible default. Every setting can be overidden if the user wants to customize the nodes, edges, or layout.
+# %%
+import networkx as nx, ultraplot as uplt, numpy as np
+
+g = nx.path_graph(10)
+A = nx.to_numpy_array(g)
+el = np.array(g.edges())
+
+fig, ax = uplt.subplots(ncols=3, refheight="3cm")
+ax[0].graph(g)  # We can plot graphs
+ax[1].graph(A)  # Or adjacency matrices and edgeslists
+ax[2].graph(el)
+ax.format(title=["From Graph", "From Adjacency Matrix", "From Edgelist"])
+uplt.show()
+
+
+# %% [raw] raw_mimetype="text/restructuredtext"
+# More Advanced Customization
+# ===========================
+# To customize a network plot, you can pass a dictionary of parameters to the :func:`~ultraplot.axes.PlotAxes.graph` function. These parameters are passed to the networkx backend, so you can refer to their documentation for more details. A more complicated example is shown below.
+# %%
+import networkx as nx, ultraplot as uplt, numpy as np
+
+# Generate some mock data
+g = nx.gn_graph(n=100).to_undirected()
+x = np.linspace(0, 10, 300)
+y = np.sin(x) + np.cos(10 * x) + np.random.randn(*x.shape) * 0.3
+layout = [[1, 2, 3], [1, 4, 5]]
+fig, ax = uplt.subplots(layout, share=0, figsize=(10, 4))
+
+# Plot network on an inset
+inax = ax[0].inset_axes([0.75, 0.75, 0.23, 0.23], zoom=False)
+ax[0].plot(x, y)
+ax[0].plot(x, y - np.random.rand(*x.shape))
+ax[0].format(xlabel="time $(t)$", ylabel="Amplitude", title="Inset example")
+inax.graph(
+    g,
+    layout="forceatlas2",
+    spines=True,
+    facecolor="white",
+    node_kw=dict(node_size=0.2),
+)
+
+# Show off different way of parsing inputs. When None is set it defaults to a Kamada Kawai
+circular = nx.circular_layout(g)
+layouts = [None, nx.arf_layout, circular, "random"]
+names = ["Kamada Kawai", "Arf layout", "Circular", "Random"]
+cmaps = ["viko", "bamo", "roma", "fes"]
+for axi, layout, name, cmap in zip(ax[1:], layouts, names, cmaps):
+    cmap = uplt.colormaps.get_cmap(cmap)
+    colors = cmap(np.linspace(0, 1, g.number_of_nodes(), 0))
+    axi.graph(g, layout=layout, node_kw=dict(node_color=colors))
+    axi.set_title(name)
+uplt.show()
--- a/docs/networks.py
+++ b/docs/networks.py
@@ -23,7 +23,7 @@
 
 # Minimal Example
 # ===============
-# To plot a graph, use :func:`~ultraplot.axes.PlotAxes.graph`. You need to merely provide a graph and ultraplot will take care of the rest. UltraPlot will automatically style the layout to give sensible default. Every setting can be overidden if the user wants to customize the nodes, edges, or layout.
+# To plot a graph, use :func:`~ultraplot.axes.PlotAxes.graph`. You need to merely provide a graph and UltraPlot will take care of the rest. UltraPlot will automatically style the layout to give sensible default. Every setting can be overidden if the user wants to customize the nodes, edges, or layout.
 # %%
 import networkx as nx, ultraplot as uplt, numpy as np
 
--- a/docs/networks.py
+++ b/docs/networks.py
@@ -24,6 +24,8 @@
 # Minimal Example
 # ===============
 # To plot a graph, use :func:`~ultraplot.axes.PlotAxes.graph`. You need to merely provide a graph and UltraPlot will take care of the rest. UltraPlot will automatically style the layout to give sensible default. Every setting can be overidden if the user wants to customize the nodes, edges, or layout.
+#
+# By default, UltraPlot automatically styles the plot by removing the background and spines, and adjusting the layout to fit within a normalized [0,1] coordinate box. It also applies an equal aspect ratio to ensure square dimensions, which is ideal for the default circular markers. Additional customization—such as modifying labels, legends, axis limits, and more—can be done using :func:~ultraplot.axes.Cartesian.format to override the default styling.
 # %%
 import networkx as nx, ultraplot as uplt, numpy as np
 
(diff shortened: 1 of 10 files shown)


beckermr (changes requested):
Can you add some tests that make images to compare via pytest-mpl? I'd like to look at them before I review.

cvanelteren:
Made a doc page with them 

https://ultraplot--169.org.readthedocs.build/en/169/networks.html

cvanelteren:
Will add a few with IMG comp too

cvanelteren:
We should maybe remove the failure on a missing test. This PR is ready for 2nd review.

beckermr:
The background being blue does not match the default styles for the rest of ultraplot. Let's make it white.

cvanelteren:
Background is off by default. This is also how networkx handles it.

cvanelteren:
Or are you referring to the test itself @beckermr ?

beckermr:
In the doc link you sent above the background was blue when I looked at it, but it is not now. I suspect one of the recent commits changed this.

cvanelteren:
Yeah I reverted a commit where facecolor was not matching the docstring. I intended it to be clear by default.

ultraplot/axes/plot.py
```diff
@@ -3666,6 +3726,134 @@ def fill_betweenx(self, *args, **kwargs):
...
+            width = max_x - min_x
+            height = max_y - min_y
+            pos = {
+                k: (
+                    (x - min_x) / width if width else 0.5,
+                    (y - min_y) / height if height else 0.5,
+                )
+                for k, (x, y) in pos.items()
+            }
+        # Set a sensible default if not given
```
beckermr, line 3798: This block of code is impressive. Very nice!

ultraplot/axes/plot.py
```diff
@@ -3666,6 +3726,134 @@ def fill_betweenx(self, *args, **kwargs):
...
+
+            # Convert to `node_size` in pt² (as required by nx.draw)
+            # 1 point = 1/72 inch → diameter in points = (desired_px / dpi) * 72
+            diameter_inch = desired_px_diameter / dpi
+            diameter_pt = diameter_inch * 72
+            radius_pt = diameter_pt / 2
+            node_size = np.pi * radius_pt**2  # area in pt²
+            node_kw["node_size"] = node_size
+
+        # By default soften the edge alpha to prevent "hairball " effect
```
beckermr, line 3826:
```suggestion
        # By default soften the edge alpha to prevent "hairball" effect
```

ultraplot/axes/plot.py
```diff
@@ -1067,6 +1068,65 @@
...
+grid : bool, default: False
+    Whether to show a background grid.
+aspect : {'equal', 'auto'} or float, default: 'equal'
+    The aspect ratio of the plot. `'equal'` ensures that the units are the same in every direction.
+    `'auto'` lets the axes scale independently. A float can specify a custom aspect ratio.
+facecolor : str or None, default: 'none'
+    The background color of the plot area. If ``None``, the default facecolor is used.
+spines : bool, iterable or str, default: False
+    Whether to show axis spines (borders around the plot). If `True`, all spines are shown.
+    If an iterable is given, only the specified spines (e.g., ``['left', 'bottom']``) are displayed.
```
beckermr, line 1111: Why are these keyword arguments here? Shouldn't the user use `ax.format` instead? I think we should remove them.

ultraplot/axes/plot.py
```diff
@@ -1067,6 +1068,65 @@
...
+    Whether to show node labels. If `True`, labels are drawn using node names. If an
+    iterable is given, only those nodes are labeled.
+nodes : bool or iterable, default: True
+    Which nodes to draw. If `True`, all nodes are drawn. If an iterable is provided, only
+    the specified nodes are included.
+edges : bool or iterable, default: True
+    Which edges to draw. If `True`, all edges are drawn. If an iterable of edge tuples is
+    provided, only those edges are included.
+grid : bool, default: False
+    Whether to show a background grid.
```
beckermr, line 1103: We should remove this in favor of people using `ax.format`.

ultraplot/axes/plot.py
```diff
@@ -1067,6 +1068,65 @@
...
+aspect : {'equal', 'auto'} or float, default: 'equal'
+    The aspect ratio of the plot. `'equal'` ensures that the units are the same in every direction.
+    `'auto'` lets the axes scale independently. A float can specify a custom aspect ratio.
```
beckermr, line 1106: Same with this one. People should be using `ax.format` and we should not provide more than one obvious way to do things.

ultraplot/axes/plot.py
```diff
@@ -3666,6 +3726,134 @@ def fill_betweenx(self, *args, **kwargs):
...
+        if labels:
+            if np.iterable(labels):
+                label_kw["labels"] = labels
+            labels = nx.draw_networkx_labels(g, pos=pos, ax=self, **label_kw)
+
+        # Apply styling
+        self.set_aspect(aspect)
+        self.grid(grid)
+        self.set_facecolor(facecolor)
+        self._toggle_spines(spines)
```
beckermr, line 3854: We should apply the default styling not via keywords to the function but rather through the standard way ultraplot gathers defaults and possible overrides. We should also provide rcparams.

ultraplot/tests/test_plot.py
```diff
@@ -0,0 +1,98 @@
...
+    nodes, edges, labels = ax.graph(g, nodes=[0, 1], labels=labels_in)
+
+    # Expecting 2 nodes 1 edge
+    assert len(edges.get_offsets()) == 1
+    assert len(nodes.get_offsets()) == 2
+    assert len(labels) == len(labels_in)
+
+
+def test_graph_edges_kw():
+    """Test the graph method by setting keywords for nodes"""
```
beckermr, line 27: This test doesn't set the edges_kw. Is that how it is supposed to be?

ultraplot/tests/test_plot.py
```diff
@@ -0,0 +1,98 @@
...
+
+    # Expecting 2 nodes 1 edge
+    assert len(edges.get_offsets()) == 1
+    assert len(nodes.get_offsets()) == 2
+    assert labels == False
+
+
+def test_graph_input():
+    """
+    Graph function should show labels when asked
```
beckermr, line 44: This doc string doesn't match the test content.

ultraplot/tests/test_plot.py
```diff
@@ -0,0 +1,98 @@
...
+    ax.graph(g)  # Graphs
+    ax.graph(A)  # Adjcency matrices
+    ax.graph(el)  # edgelists
+    with pytest.raises(TypeError):
+        ax.graph("invalid_input")
+
+
+def test_graph_layout_input():
+    """
+    Graph function should show labels when asked
```
beckermr, line 62: Doc string doesn't appear to match again.

ultraplot/tests/test_axes.py
```diff
@@ -187,3 +187,25 @@ def test_subset_format():
...
+    # Need to get the actual ax not the SubplotGridspec
+    ax[0]._toggle_spines(True)
+    assert ax.spines["bottom"].get_visible()
+    assert ax.spines["top"].get_visible()
+    assert ax.spines["left"].get_visible()
+    assert ax.spines["right"].get_visible()
+    ax[0]._toggle_spines(False)
+    assert not ax.spines["bottom"].get_visible()
+    assert not ax.spines["top"].get_visible()
+    assert not ax.spines["left"].get_visible()
```
beckermr, line 204: Is "right" supposed to be turned off here? If yes we should test that.

beckermr (changes requested):
Very nice PR! I had a few comments on code cleanup and how we handle defaults. In particular, the default styling needs to follow the normal way defaults are set and overridden in ultraplot as opposed to being keywords on the function.

ultraplot/axes/plot.py
```diff
@@ -1067,6 +1068,65 @@
...
+aspect : {'equal', 'auto'} or float, default: 'equal'
+    The aspect ratio of the plot. `'equal'` ensures that the units are the same in every direction.
+    `'auto'` lets the axes scale independently. A float can specify a custom aspect ratio.
+facecolor : str or None, default: 'none'
+    The background color of the plot area. If ``None``, the default facecolor is used.
+spines : bool, iterable or str, default: False
+    Whether to show axis spines (borders around the plot). If `True`, all spines are shown.
+    If an iterable is given, only the specified spines (e.g., ``['left', 'bottom']``) are displayed.
+rescale : bool, default: False
+    Whether to rescale the plot to fit the data. See networkx docs for more information
```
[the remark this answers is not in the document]
cvanelteren: Good catch I changed it such that the sizes of the nodes are consistent regardless of screen size. Need to remove this.

cvanelteren:
Thanks for the comments. Adjusted the PR @beckermr

docs/networks.py
```diff
@@ -0,0 +1,79 @@
...
+ax[1].graph(A)  # Or adjacency matrices and edgeslists
+ax[2].graph(el)
+ax.format(title=["From Graph", "From Adjacency Matrix", "From Edgelist"])
+uplt.show()
+
+
+# %% [raw] raw_mimetype="text/restructuredtext"
+# More Advanced Customization
+# ===========================
+# To customize a network plot, you can pass a dictionary of parameters to the :func:`~ultraplot.axes.PlotAxes.graph` function. These parameters are passed to the networkx backend, so you can refer to their documentation for more details. A more complicated example is shown below.
```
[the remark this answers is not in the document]
beckermr: Why was this marked as resolved without a link being added?

ultraplot/axes/plot.py
```diff
@@ -3666,6 +3724,143 @@ def fill_betweenx(self, *args, **kwargs):
...
+        edge_kw={},
+        label_kw={},
+        labels: Union[None, bool, Iterable] = None,
+        nodes: Union[None, bool, Iterable] = None,
+        edges: Union[None, bool, Iterable] = None,
+        grid: Union[None, str] = None,
+        aspect: Union[None, str] = None,
+        facecolor: Union[None, str] = None,
+        spines: Union[None, bool, Iterable, str] = None,
+        rescale: Union[None, bool] = None,
```
beckermr, line 3743: These keywords should be removed.

cvanelteren: Are you referring to the typehinting? I personally like them.

cvanelteren:
> Not all comments have been addressed.
> 
> I'm happy to discuss them of course, but we should do that before we merge.

which are missing?

ultraplot/axes/plot.py
beckermr: No the actual function keywords. It's odd to supply this small subset of apis here when many other functions do not.

cvanelteren: I see what you mean. However, most functions override matplotlib functionality and so keywords are passed on. When I compare UltraPlot native fuctions (e.g. `format` there are keywords given. In this case there is no matplotlib equivalent.  I would lean to having them in over having `*args, **kwargs`; or is there a way to sanitize the input with allowed keywords?

beckermr: My point was that people should be using format as the primary way to format the plot. The keywords to the graph function should only concern the graph nodes and edges themselves. Right now that function has keywords that style the entire plot.

cvanelteren: Hmm. Ok we could be move the settings  to format and then have some defaults through `rc`, would that work? Then it would be more inline with "nice defaults but full control".

beckermr: Yeah that's totally fine by me.

cvanelteren: Ok let's go with that 😬

cvanelteren: Do the latest changes resolve this?

docs/networks.py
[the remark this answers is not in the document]
cvanelteren: Do the latest changes resolve this?

cvanelteren:
I'll fix the inputs and continue to merge it. Thanks!

cvanelteren:
- [x] Docs look good
- [x] Tests pass

PR #177, opened 2025-04-25
Ensure that shared labels are consistently updated.

### LINKED ISSUE #176: Inconistent formatting when setting labelloc for shared axes
> Ah I see... that is a different bug. 
> 
> To explain what is happening. The axes are shared which makes the left most and bottom most subplot the references. If you set the ticks to be top and right, it updates them for the references. What you want, I think, is set the top labels on for the top and right labels for the right most axes. 
> 
>  

 _Originally posted by @cvanelteren in [#155](https://github.com/Ultraplot/ultraplot/issues/155#issuecomment-2829791466)_

```python
import ultraplot as pplt
fig, axs = pplt.subplots(ncols=2, nrows=2,share='all',refwidth='2cm')
axs.format(xlim=(0,60),ylim=(-20,20), xticklabelloc='t', yticklabelloc='r')
pplt.show()
```

cvanelteren:
Fixes #176 

<details> </summary> code </summary> 

```python

layout = [
    [1, 2, 0],
    [1, 2, 5],
    [3, 4, 5],
    [3, 4, 0],
]
fig, ax = uplt.subplots(layout)

ax.format(
    xticklabelloc="t",
    yticklabelloc="r",
)

layout = [
    [1, 0, 2],
    [0, 3, 0],
    [4, 0, 5],
]

fig, ax = uplt.subplots(layout)
ax.format(
    xticklabelloc="t",
    yticklabelloc="r",
)
uplt.show(block=True)

```
</details>

will now be correct:

image

### DIFF
--- a/ultraplot/tests/test_axes.py
+++ b/ultraplot/tests/test_axes.py
@@ -229,6 +229,8 @@ def test_sharing_labels_top_right_odd_layout():
             for label in getattr(ax[number], f"get_{which}ticklabels")():
                 assert label.get_visible() == state
 
+    # these correspond to the indices of the axis
+    # in the axes array (so the grid number minus 1)
     check_state([0, 2], False, which="y")
     check_state([1, 3, 4], True, which="y")
     check_state([2, 3], False, which="x")
@@ -246,6 +248,8 @@ def test_sharing_labels_top_right_odd_layout():
         xticklabelloc="t",
         yticklabelloc="r",
     )
+    # these correspond to the indices of the axis
+    # in the axes array (so the grid number minus 1)
     check_state([0, 3], False, which="y")
     check_state([1, 2, 4], True, which="y")
     check_state([0, 1, 2], True, which="x")
--- a/ultraplot/tests/test_axes.py
+++ b/ultraplot/tests/test_axes.py
@@ -187,3 +187,67 @@ def test_subset_format():
     # Shorter than number of axs
     with pytest.raises(ValueError):
         axs.format(title=["a"])
+
+
+def test_sharing_labels_top_right():
+    fig, ax = uplt.subplots(ncols=3, nrows=3, share="all")
+    # On the first format sharexy is modified
+    ax.format(
+        xticklabelloc="t",
+        yticklabelloc="r",
+    )
+    # If we format again, we expect all the limits to be changed
+    # Plot on one shared axis a non-trivial point
+    # and check whether the limits are correctly adjusted
+    # for all other plots
+    ax[0].scatter([30, 40], [30, 40])
+    xlim = ax[0].get_xlim()
+    ylim = ax[0].get_ylim()
+
+    for axi in ax:
+        for i, j in zip(axi.get_xlim(), xlim):
+            assert i == j
+        for i, j in zip(axi.get_ylim(), ylim):
+            assert i == j
+
+
+def test_sharing_labels_top_right_odd_layout():
+    layout = [
+        [1, 2, 0],
+        [1, 2, 5],
+        [3, 4, 5],
+        [3, 4, 0],
+    ]
+    fig, ax = uplt.subplots(layout)
+    ax.format(
+        xticklabelloc="t",
+        yticklabelloc="r",
+    )
+
+    def check_state(numbers: list, state: bool, which: str):
+        for number in numbers:
+            for label in getattr(ax[number], f"get_{which}ticklabels")():
+                assert label.get_visible() == state
+
+    check_state([0, 2], False, which="y")
+    check_state([1, 3, 4], True, which="y")
+    check_state([2, 3], False, which="x")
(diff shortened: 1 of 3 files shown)


cvanelteren:
What I don't like about this fix is that it will fool mpl by turning the `sharex/y` to `None`.  I need to add a way to check for this to turn this back on automatically to prevent a potential edge case to be hit (not sure which).

cvanelteren:
Need to handle non trivial layout

beckermr (changes requested):
The tests here should be used to generate an image comparison plot.

ultraplot/tests/test_axes.py
```diff
@@ -187,3 +187,67 @@ def test_subset_format():
...
+
+    def check_state(numbers: list, state: bool, which: str):
+        for number in numbers:
+            for label in getattr(ax[number], f"get_{which}ticklabels")():
+                assert label.get_visible() == state
+
+    check_state([0, 2], False, which="y")
+    check_state([1, 3, 4], True, which="y")
+    check_state([2, 3], False, which="x")
+    check_state([0, 1, 4], True, which="x")
```
beckermr, line 235: You didn't test 5.

ultraplot/tests/test_axes.py
```diff
@@ -187,3 +187,67 @@ def test_subset_format():
...
+
+    fig, ax = uplt.subplots(layout)
+    ax.format(
+        xticklabelloc="t",
+        yticklabelloc="r",
+    )
+    check_state([0, 3], False, which="y")
+    check_state([1, 2, 4], True, which="y")
+    check_state([0, 1, 2], True, which="x")
+    check_state([3, 4], False, which="x")
```
beckermr, line 252: You didn't test 5.

ultraplot/tests/test_axes.py
```diff
@@ -187,3 +187,67 @@ def test_subset_format():
...
+        [4, 0, 5],
+    ]
+
+    fig, ax = uplt.subplots(layout)
+    ax.format(
+        xticklabelloc="t",
+        yticklabelloc="r",
+    )
+    check_state([0, 3], False, which="y")
+    check_state([1, 2, 4], True, which="y")
```
beckermr, line 250: I don't understand this test case at all.

beckermr: I would assume we'd have 2, 0, 5 as on 1, 3, 4 as off for y. What does the plot look like?

beckermr: NVM I figured it out sorry.

beckermr, line 250:
```suggestion
    check_state([0, 1, 4], False, which="y")
    check_state([2, 3, 5], True, which="y")
```

Looking at the example, I would think this test should read like above.

ultraplot/tests/test_axes.py
```diff
@@ -187,3 +187,67 @@ def test_subset_format():
...
+        [0, 3, 0],
+        [4, 0, 5],
+    ]
+
+    fig, ax = uplt.subplots(layout)
+    ax.format(
+        xticklabelloc="t",
+        yticklabelloc="r",
+    )
+    check_state([0, 3], False, which="y")
```
beckermr, line 249:
```suggestion
    # these correspond to the indices of the axis
    # in the axes array (so the grid number minus 1)
    check_state([0, 3], False, which="y")
```

ultraplot/tests/test_axes.py
```diff
@@ -187,3 +187,67 @@ def test_subset_format():
...
+        xticklabelloc="t",
+        yticklabelloc="r",
+    )
+
+    def check_state(numbers: list, state: bool, which: str):
+        for number in numbers:
+            for label in getattr(ax[number], f"get_{which}ticklabels")():
+                assert label.get_visible() == state
+
+    check_state([0, 2], False, which="y")
```
beckermr, line 232:
```suggestion
    # these correspond to the indices of the axis
    # in the axes array (so the grid number minus 1)
    check_state([0, 2], False, which="y")
```

beckermr:
An image comparison test is failing, but the artifact has no images in it to look at. I think this is a bug.

cvanelteren:
I am actually tending to use less of the pytest-mpl if possible. These tests take much longer too complete than regular pytests. Currently, the test suite runs over 12 minutes I believe, which is rather long. I think we should move to more selectivel visual tests that benchmark the visual output to reduce the test running times.

cvanelteren:
Need to fix  the edgecase where it is not doing this for alt/twin axes.

cvanelteren:
Note that for twinned axes if set manually this would be the output:

image

This has `twinx` and `twiny` set for `ax[1]` (top right). Not sure if bug or feature.

beckermr (approved):

PR #186, opened 2025-04-25
skip tests properly

cvanelteren:
Fixes skipping test when baseline is not run. There is something funky going on with classes not having access to std stream. This simplifies the previous commit.

### DIFF
--- a/ultraplot/tests/conftest.py
+++ b/ultraplot/tests/conftest.py
@@ -60,7 +60,7 @@ class StoreFailedMplPlugin:
     def pytest_runtest_logreport(self, report):
         """Hook that processes each test report."""
         # Delete successfull tests
-        if not report.failed:
+        if report.when == "call" and report.failed == False:
             if self._has_mpl_marker(report):
                 self._remove_success(report)
 
--- a/ultraplot/tests/conftest.py
+++ b/ultraplot/tests/conftest.py
@@ -1,6 +1,6 @@
 import os, shutil, pytest, re, numpy as np, ultraplot as uplt
 from pathlib import Path
-import warnings, logging, sys
+import warnings, logging
 
 
 @pytest.fixture(autouse=True)
--- a/ultraplot/tests/conftest.py
+++ b/ultraplot/tests/conftest.py
@@ -1,6 +1,6 @@
 import os, shutil, pytest, re, numpy as np, ultraplot as uplt
 from pathlib import Path
-import warnings, logging
+import warnings, logging, sys
 
 
 @pytest.fixture(autouse=True)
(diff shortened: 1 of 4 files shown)


cvanelteren:
@beckermr it's picking up the dummy tests but ignoring the missing compare. See https://github.com/Ultraplot/ultraplot/actions/runs/14672582270/job/41182369596?pr=186

cvanelteren:
The net effect is the same, but I will add another run just in case.

beckermr:
We need to check that the images for the comparison failure are in the test artifact.

cvanelteren:
Ok this looks good. It is detecting
- differences in existing tests
- ignoring added tests on comp, but still runs them on the branch.
Rebasing and merging

beckermr:
Did you check that the artifact with the comparison plots had the failure?

beckermr (changes requested):
This CI run (https://github.com/Ultraplot/ultraplot/actions/runs/14673011691) has a failed test but the image comparison doesn't have the actual images in it.

Something is still broken.

beckermr:
Here is a screen shot:

Screenshot 2025-04-25 at 4 01 14 PM

cvanelteren:
Yes I did. Due to an earlier commit we have sometimes the issue that the layout algorithm will rescale the image and change the dimensions. The error is therefore not what I would expect but the point is that it did run the comparison for an existing test.

beckermr:
This other PR which has the code from main works fine: https://github.com/Ultraplot/ultraplot/actions/runs/14672477335?pr=182. So something in this PR is definitely not working.

beckermr:
> Yes I did. Due to an earlier commit we have sometimes the issue that the layout algorithm will rescale the image and change the dimensions. The error is therefore not what I would expect but the point is that it did run the comparison for an existing test.

The actual plots are not in the artifact. Here is the file tree:


Screenshot 2025-04-25 at 4 05 20 PM


There should be a subdirectory with the plots.

beckermr:
This is a working example:

Screenshot 2025-04-25 at 4 06 27 PM

beckermr:
My point is that with this PR, the plots for failures are not saved in the artifact and they should be.

cvanelteren:
Hm that's strange. Perhaps I messed up the folder locally. Not sure what could cause this.

beckermr:
OK. I am putting this one as draft until we figure that bit out.

ultraplot/tests/conftest.py
```diff
@@ -54,36 +55,35 @@ def _remove_success(self, report: pytest.TestReport):
...
-            if report.failed == False:
-                if self._has_mpl_marker(report):
-                    self._remove_success(report)
-            else:
-                print(f"{report.failed=}")
-                print(f"Test {report.nodeid} failed!")
+        # Delete successfull tests
+        if not report.failed:
+            if self._has_mpl_marker(report):
+                self._remove_success(report)
```
beckermr: I am guess it is this block somehow.

cvanelteren:
should be good now.

beckermr (dismissed):

Issue #181, opened 2025-04-25
Colorbar params from docs yield errors


ToryDeng:
From the [example](https://github.com/Ultraplot/ultraplot/issues/163#issuecomment-2811919762) you provided, I know I can modify the rotation and alignment of the colorbar's label and remove the spines of the colorbar's axes by manually creating a colorbar. However, I still want to know if there is a more convenient way to modify these properties in `UltraPlot`.

An example is:
```python
import ultraplot as uplt

cmap = uplt.Colormap("plasma_r")
fig, axs = uplt.subplots(nrows=1, ncols=2, wspace=10)
axs[0].colorbar(cmap, loc="bottom", label="My Label", labelloc="left", length=0.5)

cax = axs[1].colorbar(cmap, loc="bottom", length=0.5)
cax.ax.set_ylabel("My Label", rotation=0, ha="right", va="center")
for spine in cax.ax.spines.values():
    spine.set_visible(False)
```

![Image](https://github.com/user-attachments/assets/8f0cac8a-c0ac-4de9-8297-32e3004a6327)

The title of the second color bar is easier to read and will not overlap with the xticklabels.

cvanelteren:
The docs would suggest that you could do 

```python
cmap = uplt.Colormap("plasma_r")
fig, axs = uplt.subplots(nrows=1, ncols=2)
cbar = fig.colorbar(
    cmap,
    loc="bottom",
    label="My Label",
    labelloc="right",
    frameon=False,
)
cbar.set_label("My Label", rotation=45)
uplt.show(block=1)
```

But I see that that produces an error.  In the codebase I can see that it indeed does not handle the frame on -- even though it is listed  (see https://ultraplot.readthedocs.io/en/latest/api/ultraplot.figure.Figure.html#ultraplot.figure.Figure.colorbar). I have another commit where you can toggle the spines with one command but it is currently private (and not merged yet) I could make it public to handle this and then incorporate it within the colorbar. 

All the issues raised are valid and should be incorporated. I do have, however, another 6 ish PRs that need merging and reviewing. In short, this will go on the todos and should not be a big deal to implement.

cvanelteren:
Btw the code you are looking for is (for now)

```python

import ultraplot as uplt

cmap = uplt.Colormap("plasma_r")
fig, axs = uplt.subplots(nrows=1, ncols=2)
cbar = fig.colorbar(
    cmap,
    loc="bottom",
    label="My Label",
    labelloc="right",
)
[...]

PR #189, opened 2025-04-26
sensible defaults and unittest

cvanelteren:
The community indicated (@ToryDeng) that colorbar label placing may need some TLC. We recently introduced placing colorbar labels along the outside of the colorbar. This may cause unnecessary post processing as currently the placingo f the labels is not stylized. This PR adds sensible label placement for colorbars. It adds a new keyword `labelrotation` which is set to default to "auto" or can be set to any arbitrary number; please note the the alignment is then null and void. 

Furthermore, it adds the keyword `outline` which allows for turning on and off the outline frame of the colorbar.

Addresses #181 

@TorryDeng do you want to include an rc param for the frame or is this sufficient?

### DIFF
--- a/ultraplot/axes/base.py
+++ b/ultraplot/axes/base.py
@@ -544,8 +544,8 @@ rasterize : bool, default: :rc:`colorbar.rasterize`
     Whether to rasterize the colorbar solids. The matplotlib default was ``True``
     but ultraplot changes this to ``False`` since rasterization can cause misalignment
     between the color patches and the colorbar outline.
-outline : bool, default : True
-    Controls the visibility of the frame. When set to False, the spines of the colorbar are hidden.
+outline : bool, None default : None
+    Controls the visibility of the frame. When set to False, the spines of the colorbar are hidden. If set to `None` it uses the `rc['colorbar.outline']` value.
 labelrotation : str, float, default: "auto"
     Controls the rotation of the colorbar label. When set to auto it produces a sensible default where the rotation is adjusted to where the colorbar is located. For example, a horizontal colorbar with a label to the left or right will match the horizontal alignment and rotate the label to 0 degrees. Users can provide a float to rotate to any arbitrary angle.
 
@@ -1042,7 +1042,7 @@ class Axes(maxes.Axes):
         linewidth=None,
         edgefix=None,
         rasterized=None,
-        outline: bool = True,
+        outline: Union[bool, None] = None,
         labelrotation: Union[str, float] = "auto",
         **kwargs,
     ):
@@ -1236,6 +1236,7 @@ class Axes(maxes.Axes):
             orientation=orientation,
             **kwargs,
         )
+        outline = _not_none(outline, rc["colorbar.outline"])
         obj.outline.set_visible(outline)
         obj.ax.grid(False)
         # obj.minorlocator = minorlocator  # backwards compatibility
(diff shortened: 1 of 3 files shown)


ToryDeng:
> @TorryDeng do you want to include an rc param for the frame or is this sufficient?

Thanks for the prompt response! Adding an rc param would be helpful, especially when a figure contains multiple colorbars and users want to remove their frames.

cvanelteren:
Test fails as expected due to changes.

cvanelteren:
Tests fail as expected. Here is the change:

2025-04-26T12:53:40,074016588+02:00

cvanelteren:
Doubting on making the label on the side default hor, but otherwise fine with the PR.

cvanelteren:
You could perhaps look at this: https://github.com/matplotlib/pytest-mpl/issues/136#issuecomment-2781567975. Ended up not completing it, but it is doable.

ultraplot/axes/base.py
```diff
@@ -1033,6 +1042,8 @@ def _add_colorbar(
         linewidth=None,
         edgefix=None,
         rasterized=None,
+        outline: Union[bool, None] = None,
+        labelrotation: Union[str, float] = "auto",
```
beckermr, line 1046: Defaults needs to be handled via the rc mechanisms.

beckermr (changes requested):
One comment here.

ultraplot/axes/base.py
cvanelteren: You are right!

cvanelteren:
Any preferences for the horizontal labels @ToryDeng not sure what to do here.

ToryDeng:
> Any preferences for the horizontal labels @ToryDeng not sure what to do here.

Yes horizontal labels may not be always good for horizontal colorbars with labels to the left or right. I think ​​vertical labels​​ could be set as ​​the default in this case. Just provide users with options to adjust if needed.

ultraplot/axes/base.py
```diff
@@ -1294,6 +1312,39 @@ def _add_colorbar(
...
+                    elif labelloc == "right":
+                        kw_label["va"] = "center"
+                        labelrotation = 270
+                    else:
+                        labelrotation = 0
+                case Number():
+                    pass
+                case _:
+                    labelrotation = 0
+            print(labelrotation)
```
beckermr, line 1345: This needs to be removed.

ultraplot/tests/test_colorbar.py
```diff
@@ -290,6 +290,57 @@ def test_label_placement_colorbar():
...
+    """
+    cmap = uplt.colormaps.get_cmap("viridis")
+    mylabel = "My Label"
+    fig, ax = uplt.subplots()
+    cbar = ax.colorbar(cmap, labelloc="top", loc="right", vert=False, labelrotation=23)
+    # Get the label Text object
+    for which in "xy":
+        tmp = getattr(cbar.ax, f"{which}axis").label
+        if tmp.get_text() == mylabel:
+            label = tmp
```
beckermr, line 306: It looks like you didn't finish the test here.

beckermr (changes requested):
A few more items.

ultraplot/axes/base.py
cvanelteren: Thanks!

ultraplot/tests/test_colorbar.py
cvanelteren: Thanks!

PR #190, opened 2025-04-26
perf: faster test suite with xdist

beckermr:
As it says...

### DIFF
--- a/.github/workflows/build-ultraplot.yml
+++ b/.github/workflows/build-ultraplot.yml
@@ -58,7 +58,6 @@ jobs:
           slug: Ultraplot/ultraplot
 
   compare-baseline:
-    needs: build-ultraplot
     runs-on: ubuntu-latest
     defaults:
       run:
@@ -84,7 +83,6 @@ jobs:
           git checkout ${{ github.event.pull_request.base.sha }}
           python -c "import ultraplot as plt; plt.config.Configurator()._save_yaml('ultraplot.yml')"
           pytest \
-            -n auto \
             -W ignore \
             --mpl-generate-path=baseline \
             --mpl-default-style="./ultraplot.yml"
--- a/.github/workflows/build-ultraplot.yml
+++ b/.github/workflows/build-ultraplot.yml
@@ -43,7 +43,13 @@ jobs:
 
       - name: Test Ultraplot
         run: |
-          pytest --cov=ultraplot --cov-branch --cov-report term-missing --cov-report=xml ultraplot
+          pytest \
+            -n auto \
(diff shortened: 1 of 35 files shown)


cvanelteren:
The problem is python xdist does not play well with pytest-mpl out of the box.

cvanelteren:
You could perhaps look at this: https://github.com/matplotlib/pytest-mpl/issues/136#issuecomment-2781567975. Ended up not completing it, but it is doable.

beckermr:
> The problem is python xdist does not play well with pytest-mpl out of the box.

What happens? This is something I am happy to look into.

cvanelteren:
The issue is that the tests need to be collected on a singular thread. The tests are reported correctly, but not put in the correct folder. I was playing around with it in the link above. Was able to get it working but didn't pursue it further to clean it up. Ideally we would merge it with them for native support (and force their hand).

beckermr:
So there are multiple bugs related to using xdist here, both due to removing non-failed comparisons and due to writing the report. This will take me longer than I had guessed. :/

cvanelteren:
I ended up writing s custom html page and just moved the failed test over. But dunno if I still have the code. Will check when I have a computer.

beckermr:
I tried the code in your issue and it doesn't work with the latest mpl pytest. I plan to look later I am just very busy in the next two weeks.

cvanelteren:
Couldn't find the code unfortunately :( Thinks I torched the branch.

.github/workflows/build-ultraplot.yml
```diff
@@ -77,14 +83,27 @@ jobs:
...
+            --mpl-default-style="./ultraplot.yml"
           git checkout ${{ github.sha }}  # Return to PR branch
 
       - name: Image Comparison Ultraplot
         run: |
           mkdir -p results
           python -c "import ultraplot as plt; plt.config.Configurator()._save_yaml('ultraplot.yml')"
-          pytest -W ignore --mpl --mpl-baseline-path=baseline --mpl-generate-summary=html --mpl-results-path=./results/ --mpl-default-style="./ultraplot.yml" --store-failed-only ultraplot/tests
+          pytest \
+            -n auto \
```
beckermr, line 98:
```suggestion
```

beckermr:
I am seeing non-reproducible failures in the main test suite when using xdist in addition to failures in the image comparisons. I think this points to something more fundamentally wrong. So I am going to abandon this PR for now in favor of https://github.com/Ultraplot/ultraplot/pull/213 which gives a 25% reduction in the runtime of the tests without the tricky failures.

Issue #199, opened 2025-04-30
format of ticklabels of log scale axis


syrte:
It may be more natural to use 'log' format for tick labels by default when log scale is used.
Or at least, it would be nice to allow users to configure this behavior in rc files. Such configuration is not available yet.

Luke said that this behavior was made intentional (see our previous conversation https://github.com/proplot-dev/proplot/issues/332), and it indeed looks nice if the numbers are close to 1, (e.g., from 0.01 to 1000). I assume the typical dynamical range may depend on the specific discipline. 
Nevertheless, it is more than common to meet axes spanning several orders of magnitude, which is basically why we need log scale.

I believe that a meaningful default is better than a nice-looking default that may fail badly in some(typical) cases.

See examples below:

```python
from matplotlib import pyplot as plt
import ultraplot as uplt
import numpy as np

x = np.geomspace(1e-5, 1e10, 100)
y = x**2
fig, ax = uplt.subplot()
ax.loglog(x, y)
```
Similar problem for symlog scale.

![Image](https://github.com/user-attachments/assets/7ca784c1-bdc0-4141-a8fd-495e1ec3c04f)

This can be fixed by 
```python
ax.format(
    xticklabels='log',
    yticklabels='log'
)
```

![Image](https://github.com/user-attachments/assets/7f74d6a5-ec69-4e71-b61d-6184597e492a)

cvanelteren:
I can add the `formatter.log` flag but I don't understand the `symlog` request. Do you want to have `formatter.log` to have strings e.g. setting `symlog` or `log` for `loglog`?

syrte:
Not sure yet what's the best way to show `symlog` (do we want minor ticks in the linear range and how to set them, etc). 
But setting it to `log` scale is probably at least better than linear scale?

```
from matplotlib import pyplot as plt
import ultraplot as uplt
import numpy as np

x = np.geomspace(1e-5, 1e4, 100)
x = np.r_[-x[::-1], x]
y = x**2
fig, ax = uplt.subplot()
ax.semilogy(x, y)

ax.format(
    xscale='symlog',
    xticklabels='log',
    yticklabels='log'
)
```

![Image](https://github.com/user-attachments/assets/0c936993-d4cb-49e5-a467-56114605473e)


```
from matplotlib import pyplot as plt
import ultraplot as uplt
import numpy as np

x = np.geomspace(1e-5, 1e4, 100)
x = np.r_[-x[::-1], x]
y = x**2
[...]

PR #203, opened 2025-04-30
override logx/y/log with updated docstring

cvanelteren:
This PR introduces a new rc parameters `formatter.log` which is set to `False` by default. When set to true it automatically applies a `log` scale for x, y or both depending on the plotting function used (`semilogx, semilogy`, and `loglog` respectively).  

Addresses #199

### DIFF
--- a/ultraplot/internals/rcsetup.py
+++ b/ultraplot/internals/rcsetup.py
@@ -1200,7 +1200,8 @@ _rc_ultraplot_table = {
     "formatter.log": (
         False,
         _validate_bool,
-        "Whether to use log scale for semilog or loglog plots.",
+        "Whether to use log formatting (e.g., $10^{4}$) for "
+        "logarithmically scaled axis tick labels.",
     ),
     "formatter.limits": (
         [-5, 6],  # must be list or else validated
--- a/ultraplot/internals/rcsetup.py
+++ b/ultraplot/internals/rcsetup.py
@@ -1197,6 +1197,11 @@ _rc_ultraplot_table = {
         _validate_bool,
         "Whether to trim trailing decimal zeros on tick labels.",
     ),
+    "formatter.log": (
+        False,
+        _validate_bool,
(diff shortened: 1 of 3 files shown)


cvanelteren:
Will add some tests later. Think I may also need to add some rc params test in a separate PR.

ultraplot/internals/rcsetup.py
```diff
@@ -1197,6 +1197,11 @@ def copy(self):
         _validate_bool,
         "Whether to trim trailing decimal zeros on tick labels.",
     ),
+    "formatter.log": (
+        False,
+        _validate_bool,
+        "Whether to use log scale for semilog or loglog plots.",
```
beckermr, line 1203:
```suggestion
        "Whether to use log formatting (e.g., $10^{4}$) for "
        "logarithmically scaled axis tick labels.",
```

This doc string is confusing. The term "scale" is used to refer to how a unit of distance on the axis translate into a change in the data. Here you mean how the axis labels are formatted.

ultraplot/tests/test_plot.py
```diff
@@ -225,3 +225,41 @@ def test_quiver_discrete_colors():
     ax.quiver(X - 2, Y, U, V, C)
     ax.quiver(X - 3, Y, U, V, color="red", infer_rgb=True)
     return fig
+
+
+def test_setting_log_with_rc():
```
beckermr, line 230: This test is testing the wrong thing. You need to check the tick label formatters, not the scale of the axis itself.

beckermr (changes requested):
The doc strings and tests in this PR have confused how tick labels are formatted with how the axis itself is scaled. 

This will need fixing before it can be merged.

ultraplot/tests/test_plot.py
cvanelteren: Thanks I updated the test. Have a look

beckermr (approved):

PR #210, opened 2025-05-11
rm dep warning

cvanelteren:
I'm personally a strong proponent of calling plot methods directly on a series of axes. With that in mind, I believe we should remove this method, as I intend to continue supporting that approach. That said, I'm open to discussion if there's an important consideration I'm overlooking. One of UltraPlot’s strengths is its ability to simplify the management of subplot properties while still allowing fine-grained control. For that reason, I suggest we remove this statement.


beckermr (approved):
LGTM. If we decide the effort is too much to support, then we can revisit.

PR #212, opened 2025-05-12
fix: specify import exception type and add typing-extensions to deps

beckermr:
As it says...


beckermr:
The fact that this didn't cause a testing failure is very odd to say the least.

PR #213, opened 2025-05-12
perf: run comparison tests at the same time as the main tests

beckermr:
This PR makes the test suite about 6 minutes faster by running the image comparison tests in parallel with the main test suite.


cvanelteren:
Can you check if you included all the files that needed to be changed? I am only seeing a singular edit that is not related to the PR I suppose.

beckermr:
Yes all of the files are included. The only change is in the github workflow files. This PR does not use xdist or anything like that.

cvanelteren:
I see. Yeah let's merge this.

Issue #216, opened 2025-05-16
Bug/Broken: Color cycle is reset during successive plot calls


Ockenfuss:
According to the docs, is is possible to specify a cycle argument in successive plot calls, which should not reset the color cycle every time.
See [here](https://ultraplot.readthedocs.io/en/latest/cycles.html#changing-the-color-cycle).
However, currently this seems broken, as we can see by the example plot on the same page (copied it here from the docs):

![Image](https://github.com/user-attachments/assets/4f4c6c6c-1597-4eba-91e3-fa6df0b14c54)

cvanelteren:
Thanks.  Think I found the issue. The last case hit an edge case where the cycler object was initialized at the first draw call which set the cycler to be `qual1` but only have 1 color item. This was caused by the singular draw inferring the size of the cycler based on the columns. The previous examples worked fine as the data being passed contained all the colors. Tracking this now on #218

cvanelteren:
Fixed on main. Will bundle this with the GeoAxes fix and update it on pypi accordingly.

PR #218, opened 2025-05-16
fix cycler setting to 1 when only 1 column is parsed

cvanelteren:
Addresses #216


beckermr (changes requested):
Can we add unit test for this?

cvanelteren:
Yes on it

cvanelteren:
The issue was that it sets `N=1` from a singular column plot which creates a cycler of `N = 1` item. This PR checks that we are not setting this parameter for singular column plots.

beckermr:
There is a testing bug that looks legit. :/

cvanelteren:
Ah I see. Will fix. Think this may be a versioning issue for mocks

beckermr (approved):

Issue #217, opened 2025-05-16
gridliner label attribute error in ultraplot 1.50


sinan0310:
Hello,

I would like to report an error related to newer versions of ultraplot, below you may find the error. I am using python version 3.11.8 and when I downgrade the ultraplot to 1.10.0 in terminal by "conda install ultraplot=1.10.0" Everything works fine

```python
Traceback (most recent call last):
  File "/home/sahinos/GRIDDED_DATA/WSPD_SST/trend_ciz.py", line 42, in <module>
    fig.savefig("deneme.png",bbox_inches="tight",dpi=400)
  File "/home/sahinos/miniconda3/envs/sinan/lib/python3.11/site-packages/ultraplot/figure.py", line 2066, in savefig
    self.canvas.draw()
  File "/home/sahinos/miniconda3/envs/sinan/lib/python3.11/site-packages/ultraplot/figure.py", line 472, in _canvas_preprocess
    fig.auto_layout()
  File "/home/sahinos/miniconda3/envs/sinan/lib/python3.11/site-packages/ultraplot/figure.py", line 1747, in auto_layout
    gs._auto_layout_tight(renderer)
  File "/home/sahinos/miniconda3/envs/sinan/lib/python3.11/site-packages/ultraplot/gridspec.py", line 848, in _auto_layout_tight
    bbox = fig.get_tightbbox(renderer)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/sahinos/miniconda3/envs/sinan/lib/python3.11/site-packages/matplotlib/_api/deprecation.py", line 457, in wrapper
    return func(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^
  File "/home/sahinos/miniconda3/envs/sinan/lib/python3.11/site-packages/matplotlib/figure.py", line 1787, in get_tightbbox
    bbox = ax.get_tightbbox(
           ^^^^^^^^^^^^^^^^^
  File "/home/sahinos/miniconda3/envs/sinan/lib/python3.11/site-packages/ultraplot/axes/geo.py", line 1802, in get_tightbbox
    self._apply_axis_sharing()
  File "/home/sahinos/miniconda3/envs/sinan/lib/python3.11/site-packages/ultraplot/axes/geo.py", line 664, in _apply_axis_sharing
    self._handle_axis_sharing(
[...]

PR #219, opened 2025-05-16
Skip sharing logic when colorbar is added to GeoPlots.

cvanelteren:
This PR addresses #217. The recent added feature of sharing GeoAxes did not test when `PlotAxes` methods are called on `GeoAxes`. We don't have to apply the GeoSharing there; alternatively we could add the private function to the CartesianPlots. In this PR we just skip these axes.


cvanelteren:
Actually the issue was caused by the colorbar being added.

beckermr (approved):

PR #220, opened 2025-05-16
Allow rasterization on GeoFeatures.

cvanelteren:
This issue was discussed in #217. When exporting to SVGs, high-resolution coastline attributes cause significant file size inflation. To address this, FeatureArtists can be rasterized to reduce file bloat. Cartopy supports passing a rasterized boolean flag to features, which helps prevent this problem.

In this PR, we introduce two new options for land and coast that allow toggling rasterization. These options can be controlled directly via the format, for example, by passing format(coastrasterized=True).

### DIFF
--- a/.github/workflows/build-ultraplot.yml
+++ b/.github/workflows/build-ultraplot.yml
@@ -52,7 +52,6 @@ jobs:
           slug: Ultraplot/ultraplot
 
   compare-baseline:
-    needs: build-ultraplot
     runs-on: ubuntu-latest
     defaults:
       run:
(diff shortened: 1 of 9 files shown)


sinan0310:
Hi, adding the rasterization feature is also necessary for other features like rivers, borders and oceans.

cvanelteren:
Can you check this PR out @sinan0310 for the generated svg sizes?

cvanelteren:
Cool :).

cvanelteren:
You can also set the default to be true through the rc params directly (if you have a custom styling mpl). Then it will apply to all plots.

beckermr (approved):

beckermr:
pre-commit.ci autofix

Issue #214, opened 2025-05-13
Weird behavior of pcolormesh compared to matplotlib


lkugler:
Hi, ultraplot has really smart features.

However, I noticed a confusing behavior.
When I plot zeros using a diverging colormap "bwr" or "RdBu", I expect to get a white plot, however in ultraplot with pcolormesh it's not white.

Thanks!
ultraplot version '1.50.1', matplotlib version '3.10.1'

```python
import numpy as np
import matplotlib as mpl
import matplotlib.pyplot as plt
import matplotlib.colors as mcolors

cmap = plt.get_cmap('RdBu')
norm = plt.Normalize(vmin=-1, vmax=1)
a = np.zeros((10,10))

plt.pcolormesh(a, cmap=cmap, norm=norm)  # gives white

import ultraplot as uplt
fig, ax = uplt.subplots()
ax.pcolormesh(a, cmap=cmap, norm=norm)  # gives red-ish color

``` 

![Image](https://github.com/user-attachments/assets/c8372456-4bd3-47c8-8235-0d08f0557b2b)

cvanelteren:
UltraPlot infers if the data is discrete from the input. You can override this by providing `discrete=False`. You can see this by plotting the colorbar. On the right discrete is off and you see that the center is 0. On the left is the code you provided and the colorbar shows a color as it puts zero in one of the two bins.  More of the behavior can be seen in https://ultraplot.readthedocs.io/en/stable/api/ultraplot.axes.PlotAxes.html#ultraplot.axes.PlotAxes.pcolormesh 

![Image](https://github.com/user-attachments/assets/2e8418de-6b6a-4861-81af-e1bde06cab78)

<details> <summary> code </summary>

```python
import ultraplot as uplt
import numpy as np

a = np.zeros((10, 10))
fig, ax = uplt.subplots(ncols=2)
for discrete, axi in zip([True, False], ax):
    cbar = axi.pcolormesh(a, cmap="rdbu", discrete=discrete, colorbar="r")
fig.show()
uplt.show(block=1)
```

</details>

cvanelteren:
oh and hi and welcome 👋

cvanelteren:
@beckermr not sure if we should allow for the white bin to be the center as I do think this is kinda confusing.

cvanelteren:
The first is not necessary is it? 

I do agree that this is a none issue for any real data and that we are already covering the majority, but to ensure the edgecase a fix would be nice I think.

[...]

PR #228, opened 2025-05-21
Adjust the ticks to center on 'nice' values

cvanelteren:
With UltraPlot, our goal is to provide high-quality scientific visualizations with sensible defaults. However, these defaults don’t always meet the needs of every use case and occasionally require adjustment.

A recent issue (#214) highlighted unexpected behavior in `pcolormesh`. By default, the associated colorbars use `MaxNLocator` to generate a "nice" range based on the data. While this works well in many cases, it can lead to counterintuitive results with diverging colormaps—specifically, values around zero may be split across positive and negative bins, which is visually misleading.

This PR introduces a new keyword argument, center_values, which adjusts the colorbar range so that the "nice" limits are centered on bin edges. This ensures that zero lies at the center of a bin, providing a more intuitive representation for diverging colormaps.

```python

a = np.random.rand(10, 10) * 2 - 1
fig, ax = uplt.subplots(ncols=2)
for axi, centered in zip(ax, [False, True]):
    axi.pcolormesh(
        a,
        colorbar="r",
        N=10,
        center_values=centered,
    )
    ax.set_title("Default" if not centered else f"{centered=}")
uplt.show(block=True)
```

Will not produce:

image

### DIFF
--- a/ultraplot/axes/plot.py
+++ b/ultraplot/axes/plot.py
@@ -2853,6 +2853,7 @@ class PlotAxes(base.Axes):
         norm_kw=None,
         extend=None,
         symmetric=None,
+        center_values=None,
         **kwargs,
     ):
         """
@@ -2891,6 +2892,7 @@ class PlotAxes(base.Axes):
         locator_kw = locator_kw or {}
         extend = _not_none(extend, "neither")
         levels = _not_none(levels, rc["cmap.levels"])
+        center_values = _not_none(center_values, False)
         vmin = _not_none(vmin=vmin, norm_kw_vmin=norm_kw.pop("vmin", None))
         vmax = _not_none(vmax=vmax, norm_kw_vmax=norm_kw.pop("vmax", None))
         norm = constructor.Norm(norm or "linear", **norm_kw)
@@ -2963,6 +2965,15 @@ class PlotAxes(base.Axes):
             nlevels.append(olevels[-1])
             levels = norm.inverse(nlevels)
 
+        # Center the bin edges around the center of the bin
+        # rather than its edges
+        if center_values:
+            # Shift the entire range but correct the range
+            # later
+            width = np.diff(levels)[0]
+            levels -= width * 0.5
+            # Add another bin edge at the width
+            levels = np.append(levels, levels[-1] + width * np.sign(levels[-1]))
         return levels, kwargs
 
     def _parse_level_vals(
@@ -4778,7 +4789,10 @@ class PlotAxes(base.Axes):
             to_centers = edges = False
         x, y, z, kw = self._parse_2d_args(x, y, z, edges=edges, **kwargs)
         kw.update(_pop_props(kw, "collection"))
-        kw = self._parse_cmap(x, y, z, to_centers=to_centers, **kw)
+        center_values = kwargs.pop("center_values", None)
+        kw = self._parse_cmap(
+            x, y, z, to_centers=to_centers, center_values=center_values, **kw
+        )
         edgefix_kw = _pop_params(kw, self._fix_patch_edges)
         labels_kw = _pop_params(kw, self._add_auto_labels)
         guide_kw = _pop_params(kw, self._update_guide)


cvanelteren:
I am not too sure yet on leaving the minor ticks in in the centered plot, otherwise looks good I think.

cvanelteren:
Need to address the errors.

ultraplot/axes/plot.py
```diff
@@ -396,7 +396,7 @@
     If the latter, levels are inferred using `~ultraplot.utils.edges`.
     This will override any `levels` input.
 center_levels : bool, default False
-    If set to true, the bin edges will be centered. This option can be used for divergin colormaps to ensure will be allocated a separate bin.
+    If set to true, the bin edges will be centered. This option can be used for diverging colormaps to ensure values are allocated to separate bins.
```
beckermr, line 399: This doc string needs an edit. Here is a suggestion:

```suggestion
    If set to true, the discrete color bar bins will be centered on the level values 
    instead of using the level values as the edges of the discrete bins. This option can be used for diverging, discrete color bars with both positive and negative data to ensure data near zero is properly represented.
```

cvanelteren:
> I don't get why we don't simply force the number of segments to be odd in some cases, instead of shifting the labels. I think the way UltraPlot labels the edges is better in all cases?

There are a few reasons why we don’t just force an odd number of segments:

1. **The default is already odd**: In most cases, the default number of bins is already odd, so there’s not much to enforce.
2. **Tick generation via `MaxNLocator`**: When ticks are generated using `MaxNLocator`, the number of bins is at most `N + 1` and can be fewer. This approach doesn’t guarantee centered bins—especially around zero—and avoids introducing a hard dependency on a fixed bin count. Using something like `linspace` would require similar post-processing to get visually acceptable results, essentially duplicating what `MaxNLocator` already does well.
3. **UltraPlot's design choices**: UltraPlot explicitly aligns ticks with the centers of colorbar segments, which Matplotlib does not do. This ensures that the tick positions correspond to actual data values within each bin. By aligning ticks this way, UltraPlot effectively achieves the benefits of odd binning—centered bins around key values—while remaining compatible with Matplotlib’s default behavior.

cvanelteren:
Also regarding the moving of the labels. I tried initially to just move the range without centering the labels. This places the ticks outside the range we are looking at (e.g. 1.1 to -1.1 in the case above). Additionally, this removes the 0 label and instead adds the -0.2 0.2 bin. It looked visually not pleasing to me and there were some artifacts with what was on screen and what the code told me that the range was. For  example when showing the figure it would show 1.1 but not -1.1. When the image was fullscreened all the ticks were visible, but exporting it resulting in the first draw without fullscreen.

beckermr (changes requested):
OK. Agreed that we need to turn off minor ticks. Was this done?

cvanelteren:
Just pushed turning the ticks off. Not sure which is preferred

image

cvanelteren:
Oh and we wanted the grid off right?

beckermr:
Yes we wanted the grid off too! Good catch!

beckermr:
I like the minor ticks off for the discrete case for sure. We should check continuous color bars are not changed.

ultraplot/axes/plot.py
```diff
@@ -396,7 +396,7 @@
     If the latter, levels are inferred using `~ultraplot.utils.edges`.
     This will override any `levels` input.
 center_levels : bool, default False
-    If set to true, the discrete color bar bins will be centered on the level values 
+    If set to true, the discrete color bar bins will be centered on the level values
     instead of using the level values as the edges of the discrete bins. This option can be used for diverging, discrete color bars with both positive and negative data to ensure data near zero is properly represented.
```
beckermr, line 400:
```suggestion
    If set to true, the discrete color bar bins will be centered on the level values
    instead of using the level values as the edges of the discrete bins. This option
    can be used for diverging, discrete color bars with both positive and negative
    data to ensure data near zero is properly represented.
```

ultraplot/internals/rcsetup.py
```diff
@@ -1575,6 +1575,11 @@ def copy(self):
         "Color for negative bars and shaded areas when using ``negpos=True``. "
         "See also :rcraw:`poscolor`.",
     ),
+    "pcolormesh.grid": (
```
beckermr, line 1578: There are several ways to plot grids of data, including pcolor and imshow. Do we need to generalize this option to cover all of them? IDK what the semantics of ultrplot is around these redundancies.

cvanelteren: Good question. Not sure. There is `axes.grid` being set to `True` by default. I can also see that there is a `colorbar.grid` but nothing else. So I assume the rationale is: by default on, only in special cases off.

cvanelteren: In terms of the other functions would need to check but I am already over my allotted time for this -- so will have to check back in in a few days.

ultraplot/internals/rcsetup.py
```diff
@@ -32,6 +32,7 @@
 # NOTE: These are set to True inside __init__.py
 # NOTE: We really cannot delay creation of 'rc' until after registration because
 # colormap creation depends on rc['cmap.lut'] and rc['cmap.listedthresh'].
+# center_levels = _not_none(center_levels, rc["c
```
beckermr, line 35: This comment needs to be fixed. IDK what it does or if we should just remove it.

beckermr (changes requested):
OK. This looks good to go. One more item I found and then we can merge.

ultraplot/internals/rcsetup.py
cvanelteren: ah yes thanks!

Issue #234, opened 2025-05-26
Issue with shared axis labels on geographic axis


lwq-star:
Referencing this conversation #155 , everything works normally, but looping through each subplot to plot causes a bug.

```
import ultraplot as uplt
import numpy as np

fig, axs = uplt.subplots(ncols=2, nrows=3, span=False, share='all', proj="cyl",
                         journal='nat2')
axs.format(abc=True, abcloc="ul", lonlabels='both', latlabels='both')
```

![Image](https://github.com/user-attachments/assets/203601f8-eeab-49c7-b292-121d49d207d5)

This is my example.

```
import ultraplot as uplt
import numpy as np

fig, axs = uplt.subplots(ncols=2, nrows=3, span=False, share='all', proj="cyl",
                         journal='nat2')
axs.format(abc=True, abcloc="ul", lonlabels='both', latlabels='both')

a = np.random.rand(10, 10) * 2 - 1

for idx in range(6):
    ax = axs[idx]
    m = ax.pcolormesh(a, cmap="viridis")
fig.colorbar(m, label="ac1", ticks=4, loc="r")

```

![Image](https://github.com/user-attachments/assets/7dd20c19-cd54-41f0-a5a0-f22332c0dd16)
There might be some conflicts that have arisen.

cvanelteren:
Need some minor fixes but this will be fixed soon. For users now turn off the sharing and it will work as it worked in the past

cvanelteren:
Fixed on main/

PR #236, opened 2025-05-27
Hotfix get_border_axes

cvanelteren:
The colorbar (apparently) fundamentally changes the gridspec in a non-trivial way, see #234. This causes an issue with the introduced `GeoAxes` sharing feature where the borders were determined based on the position in the gridpec. We can sidestep this issue by using the gridspec of the figure to reconstruct the position.

Current implementation now produces

image

### DIFF
--- a/ultraplot/axes/geo.py
+++ b/ultraplot/axes/geo.py
@@ -666,6 +666,9 @@ class GeoAxes(shared._SharedAxes, plot.PlotAxes):
                 target_axis=self._lataxis,
             )
 
+        # If we are not stale just return
+        if not self.stale:
+            return
         # Turn all labels off
         # Note: this action performs it for all the axes in
         # the figure. We use the stale here to only perform
@@ -701,6 +704,7 @@ class GeoAxes(shared._SharedAxes, plot.PlotAxes):
             for side in sides:
                 tmp[side] = True
             axi._toggle_gridliner_labels(**tmp)
+        self.stale = True
 
     def _get_gridliner_labels(
         self,
--- a/ultraplot/axes/geo.py
+++ b/ultraplot/axes/geo.py
@@ -666,6 +666,42 @@ class GeoAxes(shared._SharedAxes, plot.PlotAxes):
                 target_axis=self._lataxis,
             )
 
+        # Turn all labels off
+        # Note: this action performs it for all the axes in
+        # the figure. We use the stale here to only perform
+        # it once as it is an expensive action.
+        border_axes = self.figure._get_border_axes()
+        # Recode:
+        recoded = {}
+        for direction, axes in border_axes.items():
+            for axi in axes:
+                recoded[axi] = recoded.get(axi, []) + [direction]
+
+        # We turn off the tick labels when the scale and
+        # ticks are shared (level >= 3)
+        are_ticks_on = True
+        if self.figure._get_sharing_level() >= 3:
+            are_ticks_on = False
+
+        default = dict(
+            left=are_ticks_on,
+            right=are_ticks_on,
+            top=are_ticks_on,
+            bottom=are_ticks_on,
+        )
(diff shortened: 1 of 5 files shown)


cvanelteren:
For context, when a colorbar is added, for example, to the right, the gridspec will change the coordinate for the 4th plot to (1,2),(1,2) (row and col span) I really don't understand why this happens and a similar cut happens for plots 6. Furthermore, the gridspec is changed from 3 x 3 to a 4x3 which does not make sense -- 3x4 would make sense but then still the gridspec spans are odd. Dunno why this happens but this PR sidesteps that issue.

cvanelteren:
The border checker algorithm is replaced with a recursive algorithm based on the grid space. It uses the internal functions to recreate a gridspace rather than to rely on the subplotspec directly. We do need to put it on the docket why the subplotspec behaves so strangely. I would suggest we add this to the todos when we cover the gridspec.py in tests.

cvanelteren:
Need to still fix that after unsharing the default plots will look different than when the axes were not shared.

ultraplot/axes/geo.py
```diff
@@ -722,41 +772,6 @@ def _handle_axis_sharing(
...
-            # sharing that is specific for the GeoAxes.
-            if not isinstance(axi, GeoAxes):
-                continue
-
-            sides = recoded.get(axi, [])
-            tmp = default.copy()
-            for side in sides:
-                tmp[side] = True
-            axi._toggle_gridliner_labels(**tmp)
         self.stale = False
```
beckermr, line 776: I am confused about why this bit of code was left here. In the top of `_apply_axis_sharing` the code reads

```
  # Handle X axis sharing
        if self._sharex:
            self._handle_axis_sharing(
                source_axis=self._sharex._lonaxis,
                target_axis=self._lonaxis,
            )

        # Handle Y axis sharing
        if self._sharey:
            self._handle_axis_sharing(
                source_axis=self._sharey._lataxis,
                target_axis=self._lataxis,
            )
```

So how does the code block you moved ever get called then?

beckermr: Specifically, it appears the code always sets stale = False and then you early return with 

```
# If we are not stale just return
        if not self.stale:
            return
```

cvanelteren: The code gets called through `GeoAxes.draw`. I may have reversed the logic. But to my understanding when an object is stale, we don't have to update. So I think the code is correct, but we can move it outside of draw function and move it to the setup, and add appropriate calls when the labels are changed.

cvanelteren: To add to this -- the reason I am skipping it as the draw is called repeatedly and it is unnecessary to handle this (and rather expensive)

cvanelteren: I am not really happy with the answer I gave you. I see your point. I think the labels should be handled on setup rather than on the sharing step. I will move it there and update it when the label setters are called. One moment please.

cvanelteren: hmm there is some issue I have to solve.

cvanelteren: Ok I tried moving it to the setup but the gridmajors a only modified on draw which creates an issue when we attempt to set it up on the figure class for example. I am reverting the commits

cvanelteren: Ok so the issue is that for the normal cartesian plots we can rely on the internal mechanisms from matplotlib where sharex and sharey exists. For GeoAxes the ticks are behaving differently depending on the backend used. The gridliner object is created relatively late in its life cycle and here we are patching it in such a way where we are modifying the labels after all the figures are created. When subplots are added iteratively, the sharing may be incorrect on creation -- on draw we know that the plot should have all the plots we want to see (at least for that draw). This logic relies on the stale mechanism to not draw this on every frame but only when changes are necessary.

Issue #237, opened 2025-05-28
panel for  geo axis


lwq-star:
I want to add panels to the geo axis, but this error occurred.

```
import ultraplot as uplt
import numpy as np

# Dummy data
data = np.random.rand(10, 10)

fig, ax = uplt.subplots(proj="cyl")
c1 = ax.imshow(data, cmap="viridis")

ax.colorbar(c1, loc="r", label="Some Label")
px = ax.panel_axes("b")

uplt.show()
```

`AttributeError: '_CartopyAxes' object has no attribute 'is_panel_group_member'`

I want an effect similar to this. The main plot uses a geographic axis, with additional panels using regular axes.

![Image](https://github.com/user-attachments/assets/7a518f9b-0e07-487d-ad8c-b2988b555838)

lwq-star:
Will this feature be possible in the future?

cvanelteren:
We meet again!  

The GeoPlot needs some TLC and we are looking for active maintainers that focus on these areas. I am not working inside this field so for me it is rather difficult to see the bugs on this part of the codebase directly. In the end, I would like to see that the GeoAxes behave the same as the cartesian plots (or 3d plots). In short, yes this should be a features, but I am also limited on time and space so contributions are more than welcome. I will put this on the docket too.

lwq-star:
Thank you for your contributions. After I finish my paper, I will also do what I can within my limited capacity to see how to implement this feature.

> We meet again!
> 
> The GeoPlot needs some TLC and we are looking for active maintainers that focus on these areas. I am not working inside this field so for me it is rather difficult to see the bugs on this part of the codebase directly. In the end, I would like to see that the GeoAxes behave the same as the cartesian plots (or 3d plots). In short, yes this should be a features, but I am also limited on time and space so contributions are more than welcome. I will put this on the docket too.

cvanelteren:
Just curious how helpful would panels be for GeoAxes? By  default the panels share the coordinate references as such won't the plot not be difficult to maintain? I implemented a fix in #238 but need to think a bit more about the tests and or if this feature would be helpful and I would love your input.

lwq-star:
I have an example here, but I don't know how to adjust the size of the right-side plot frame. I want to set refaspect=0.5, but the fig.subplot does not have this parameter.
```
import ultraplot as uplt

fig = uplt.figure(span=False, share='all', journal='nat2')
gs = uplt.GridSpec(ncols=2, nrows=2)

[...]

PR #238, opened 2025-05-29
Hotfix panel

cvanelteren:
Added some functions that were missing. This would address #237

### DIFF
--- a/ultraplot/axes/geo.py
+++ b/ultraplot/axes/geo.py
@@ -9,7 +9,7 @@ from functools import partial
 try:
     # From python 3.12
     from typing import override
-except:
+except ImportError:
     # From Python 3.5
     from typing_extensions import override
 
--- a/ultraplot/axes/geo.py
+++ b/ultraplot/axes/geo.py
@@ -6,10 +6,20 @@ import copy
 import inspect
 from functools import partial
 
+try:
+    # From python 3.12
+    from typing import override
+except:
+    # From Python 3.5
+    from typing_extensions import override
+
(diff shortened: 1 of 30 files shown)


cvanelteren:
Not sure how helpful panels would be for `GeoAxes` but the api should be consistent or raise a warning.

cvanelteren:
The test will produce this when saved but not when viewing. This is not related to the PR I believe, but should be marked as suspicious for the future.

test

cvanelteren:
The PR now contains the HEAD of #236

ultraplot/axes/geo.py
```diff
@@ -557,6 +557,18 @@ def _share_limits_with(self, other: "GeoAxes", which: str):
...
+    def _is_panel_group_member(self, other):
+        """
+        Return whether the axes belong in a panel sharing stack..
+        """
+        return (
+            self._panel_parent is other  # other is child panel
+            or other._panel_parent is self  # other is main subplot
+            or other._panel_parent
+            and self._panel_parent  # ...
+            and other._panel_parent is self._panel_parent  # other is sibling panel
```
beckermr: Can you add parentheses `()` to clarify the logic here?

cvanelteren: I will actually refactor this. I copied this from the other function in base or cartesian. Looks indeed very confusing like this.

cvanelteren: Actually we should move this to `base.py` since the logic is the same now.

cvanelteren: done

PR #242, opened 2025-05-31
Fix links for 1d plots in docs

cvanelteren:


cvanelteren:
Need to fix some issues. It's a bit weird how a method is sometimes a func and vice versa

cvanelteren:
Will merge this after visual inspection. For some reason `set_x/ylim` does not show for :func: or :method: not sure why

PR #243, opened 2025-05-31
Deprecate basemap

cvanelteren:
This PR will introduce a deprecation warning for basemap. The rationale behind deprecation is as follows: basemap was deprecated in the past, but it seemed like development was picking  up steam again. It has been a few months since I submitted a few PRs that would update basemap with modern tooling (in particular numpy, see https://github.com/matplotlib/basemap/issues/604, https://github.com/matplotlib/basemap/pull/623, https://github.com/matplotlib/basemap/pull/622). Not sure what the status is currently but little activity has been seen on that repo for a while. Furthermore, cartopy has more powerful features with good matplotlib support. Lastly, it would streamline our development as we only have to focus on a singular backend and `geo.py` needs a refactor in my opinion. 

I open this PR up for community members to chip in, but I don't see a good reason to not do this.

### DIFF
--- a/ultraplot/figure.py
+++ b/ultraplot/figure.py
@@ -826,7 +826,7 @@ class Figure(mfigure.Figure):
         # Basemap is currently being developed again so are removing the deprecation warning
         if backend == "basemap":
             warnings._warn_ultraplot(
-                f"{backend=} will be deprecated in next major release (v1.6)"
+                f"{backend=} will be deprecated in next major release (v2.0). See https://github.com/Ultraplot/ultraplot/pull/243"
             )
         return backend
 


ultraplot/figure.py
```diff
@@ -824,6 +824,10 @@ def _parse_backend(backend=None, basemap=None):
         Handle deprecation of basemap and cartopy package.
         """
         # Basemap is currently being developed again so are removing the deprecation warning
+        if backend == "basemap":
+            warnings._warn_ultraplot(
+                f"{backend=} will be deprecated in next major release (v1.6)"
```
beckermr: Please be careful about version numbers. Version 1.6 < 1.50 which is where we are now. So this needs to be 1.60. Also, a "major release" would be `2.*`. What do you mean here?

beckermr (changes requested):

ultraplot/figure.py
cvanelteren: Good catch! 2.0 seems a bit  far away to me. I have the feeling that people mainly use cartopy and do not use basemap that much -- after all we default to cartopy. We can set it to 2.0 for now and slowly approach the number. It wou

Issue #233, opened 2025-05-25
Sharing axes labels goes wrong for twinned axes.


cvanelteren:
First addresses in  #232. 

For non-twinned axes the labeling is shared; meaning for a grid we get an expected label on the left for sharing labels above or equal to one. For shared labels this does not occur.

For sharing level "all", the last plot get the ylabel but none are spawned.
For sharing level "labels" the label is copied from the axis it is shared from.

# Expectation

The labels should behave similarly to how the non-twinned subplots behave, with the axes being shared. According to the docs, the properties are copied from the sharing axis. I can confirm that this is correctly done. Not sure how trivial this would be to implement since the sharing will make the determination more complex. We could use the `_get_border_axes` function to determine where to put the labels.

<details> <summary> snippet </summary>

```python
import ultraplot as uplt, numpy as np, pandas as pd

df = pd.DataFrame(np.random.rand(10, 2), columns=["a", "b"])
layout = [[1, 2], [3, 4]]
fig, axs = uplt.subplots(layout, share="all")
for axi in axs:
    alt = axi.alty()
    axi.plot(x="a", y="b", data=df)
    alt.plot(x="a", y="b", data=df)
uplt.show(block=1)
```
</details>

# Problem plots
## share labels
![Image](https://github.com/user-attachments/assets/bca06f99-850e-478b-9bcc-f8c6485698cf)

## share all

![Image](https://github.com/user-attachments/assets/6d8283d7-2b3a-4217-a7c2-9163b2f916bf)

cvanelteren:
Need to confirm:
- How are y labels shared across subplots
- How are twinned axes spawned. 
They are held within `_twinned_axes` and the current yaxis is being shared.

cvanelteren:
For example, this:

![Image](https://github.com/user-attachments/assets/4c2ac9ef-4b6f-4602-b0ae-8e234d02150e)
Looks great. 

<details> <summary> snippet </summary> 

```python
import ultraplot as uplt, numpy as np, pandas as pd

df = pd.DataFrame(np.random.rand(10, 2), columns=["a", "b"])
layout = [[1, 2], [3, 4]]
fig, ax = uplt.subplots(layout, share=1)

for axi in ax:
    alt = axi.alty()
    axi.plot(x="a", y="b", data=df)
    alt.plot(x="a", y="b", data=df)

ax.format(xlabel="Main x", ylabel="Main y")

fig.save("test.png")
uplt.show(block=1)
```
</details>

[...]

PR #244, opened 2025-06-02
Feat more advanced sharing of axes.

cvanelteren:
Fixed issue where twinned axes reversed axis sharing: https://github.com/Ultraplot/ultraplot/issues/233

### DIFF
--- a/ultraplot/axes/base.py
+++ b/ultraplot/axes/base.py
@@ -3185,6 +3185,23 @@ class Axes(maxes.Axes):
             return True
 
         # Not in the same panel group
+
+    def _is_ticklabel_on(self, side: str) -> bool:
+        """
+        Check if tick labels are on for the specified sides.
+        """
+        # NOTE: This is a helper function to check if tick labels are on
+        # for the specified sides. It returns True if any of the specified
+        # sides have tick labels turned on.
+        axis = self.xaxis
+        if side in ["labelleft", "labelright"]:
+            axis = self.yaxis
+        label = "label1"
+        if side in ["labelright", "labeltop"]:
+            label = "label2"
+        for tick in axis.get_major_ticks():
+            if getattr(tick, label).get_visible():
+                return True
         return False
 
     @docstring._snippet_manager
--- a/ultraplot/axes/base.py
+++ b/ultraplot/axes/base.py
@@ -1589,7 +1589,7 @@ class Axes(maxes.Axes):
             return [pax for pax in paxs if not pax._panel_hidden and pax._panel_share]
 
         # Internal axis sharing, share stacks of panels and main axes with each other
-        # NOTE: This is called on the main axes whenver a panel is created.
+        # NOTE: This is called on the main axes whenever a panel is created.
         # NOTE: This block is why, even though we have figure-wide share[xy], we
         # still need the axes-specific _share[xy]_override attribute.
         if not self._panel_side:  # this is a main axes
--- a/ultraplot/axes/cartesian.py
+++ b/ultraplot/axes/cartesian.py
@@ -387,10 +387,6 @@ class CartesianAxes(shared._SharedAxes, plot.PlotAxes):
             if level > 0:
                 labels._transfer_label(axis.label, self._sharex.xaxis.label)
                 axis.label.set_visible(False)
-            if level > 2:
-                # WARNING: Cannot set NullFormatter because shared axes share the
-                # same Ticker(). Instead use approach copied from mpl subplots().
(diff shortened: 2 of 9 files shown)


cvanelteren:
~~Need to address some of the failing tests but otherwise looks good.~~

cvanelteren:
I think the issue is that the mapping from sharing and sharee is exactly inverted in UltraPlot compared to matplotlib. This is not nessarily an issue, except that we cannot toggle the tick_params individually. This is what needs to be addressed here to properly fix it. Not sure yet how tho.

cvanelteren:
I’m trying to track down a behavioral difference between UltraPlot and Matplotlib regarding axis sharing and tick label display.

In Matplotlib, the shared and sharee roles appear to be the inverse of how UltraPlot handles them—but this doesn't seem to be the root cause of the issue I'm seeing.

Here’s what I’ve found: in Matplotlib, we can control tick label visibility per subplot. For example, setting
```
ax[0, 0].tick_params(labeltop=True, labelright=True)
```
results in the following output:

Matplotlib example
image

However, doing the same in UltraPlot does not show the tick labels:

UltraPlot example
image

Now that I think about it, this might actually be due to how the shared axes are set up. In Matplotlib, the top-left subplot (ax[0, 0]) does not share its axis with any other subplot (i.e., its .sharex and .sharey are None). In UltraPlot, however, those attributes are not None—which might suppress the label rendering.

When I manually set the shared axes for the top-left subplot in UltraPlot to None, the tick labels appear as expected.

Could use your thoughts on this, @beckermr—does this sound like a difference in how the initial sharing is applied or propagated?

beckermr:
TBH I don't know at all. :/

cvanelteren:
hmmm ok. We could revert the commit for now, but it is also broken in that state. I cannot directly see the difference that are apparent. I can at least state:
- we are not overriding tick_params
- labels become visible when _sharex/y is set to None
- Reversing sharex or sharey will merely move the labels to the "other" set than is active currently but  will not solve the issue (see comment above for reverse order)
- somehow UltraPlot is forcing that labels are off and cannot be controlled by tick_params which is not the same as for matplotlib. 
Will look more into this issue, but progress may be slow.

cvanelteren:
Found the issue in `_apply_axis_share`

cvanelteren:
I need to move some logic there to the setup, and check whether it will affect downstream processing. But little later ;-)!

cvanelteren:
I believe I got it working :partying_face:.

The function `_apply_axis_sharing` was forcing the ticks to be off for all axes top and bottom for x and left an right for y. I moved this logic to the setup functions and I am using the same logic as is used for the `GeoAxes` to determine which labels should be on or off. This creates an additional benefit that the geoplots and cartesian plots end up having a similar feel (or at least is going to). 

Note that with this change, we can now modify the axes post creation which prevents a weird disconnect with behavior under matplotlib and under ultraplot.

cvanelteren:
Need to move some code around, but this is starting to look good: the problem is identified and we can solve this.

cvanelteren:
I refactored the sharing and moved it into figure. I have in addition introduced some new functionality to utils for internal use. I further expanded the axes sharing to allow for subsharing depending on the axis type (geo, cartesian etc). It is in a rough state now and it drastically needs some tests, but I intend to move the sharing and turning on and off on label on figure creation, and or when all axes are formatted.  This differs from the current implementation that forces everything on draw and limits the kinds of sharing that is allowed. It additionally creates a side effect where modifying singular subplots is not allowed.

cvanelteren:
Some tests are still failing, but most of the pieces are in place now. Just need to check that on creation the sharing is called at an appropriate time.

cvanelteren:
Need to fix an edge case where some plots are spanning another (or vice versa). For these cases the labels should be turned on.

cvanelteren:
I guess the best approach would be to convert the crawler into something that check ranges. Then if the range is exactly the same to the left or right, we drop the label, otherwise we keep them on (same for top and bottom).

cvanelteren:
Need to finalize this before I forget and become not merged in main. 


the main issue is that ultraplot currently works flawlessly for sharing bottom and left labels but fails to work for top right labels on regular grids and non regular layouts. 

I see the major issue can be encapsulated as: 

- in the apply sharing function, we force the labels for x or y to be off if we are sharing x or y, this needs to be undone if the axis is a border and potentially removing the labels from the not shared axis.


I initially tries to solve the issue by inverting the sharing pattern, but this does not solve the issue as it will cause strange interactions with twinned axes and furthermore, will not work if all the labels are on. This would imply that we copy the logic for sharing and introduce `sharex_top` and `sharey_right` or something similar.

PR #251, opened 2025-06-06
Feature: Beeswarm plot

cvanelteren:
With AI and machine learninig all the rage I hereby introduce a wanted feature: the beeswarm plot. This plot is used to indicate feature importance using shapely values. The plot type is more general and is effectively a scatter plot that removes overlap between the points to indicate the distribution of a feature. It would allow for visualizing data in both directions and adding or removing colorbars if necessary.

It looks like:

image

### DIFF
--- a/ultraplot/axes/plot.py
+++ b/ultraplot/axes/plot.py
@@ -733,6 +733,8 @@ matplotlib.axes.Axes.scatter
 docstring._snippet_manager["plot.scatter"] = _scatter_docstring.format(y="y")
 docstring._snippet_manager["plot.scatterx"] = _scatter_docstring.format(y="x")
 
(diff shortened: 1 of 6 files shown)


ultraplot/axes/plot.py
```diff
@@ -733,6 +733,20 @@
 docstring._snippet_manager["plot.scatter"] = _scatter_docstring.format(y="y")
 docstring._snippet_manager["plot.scatterx"] = _scatter_docstring.format(y="x")
 
+_beeswarm_docstring = """
+Beeswarm plot with SHAP-style feature value coloring.
```
beckermr: Let's make this a link to their docs.

cvanelteren: Done, and thanks again for the watchful eye.

beckermr (approved):

Issue #252, opened 2025-06-08
I have already set lonlabels="b" and latlabels="l", but the top and right labels still appear. How can I fix this?"


nickbsb:
ax1 = fig.subplot(gs[0, 0], proj=fplt.PLATE_CARREE)
ax1.format(lonlim=(118, 136), latlim=(38, 54), 
           grid=False,
           lonlabels="b", latlabels="l",
           lonticklen=1,latticklen=1,labelsize=8)

![Image](https://github.com/user-attachments/assets/94dda0c7-23ac-4482-9c7b-d64a708452b5)

cvanelteren:
what version are you on?

cvanelteren:
This should be fixed on main and or latest release.

cvanelteren:
For context, v1.50 erroneously turned all the tick on for geo by default, this is fixed in recent releases.

nickbsb:
My version of matplotlib is 3.9.4, and ultraplot is 1.55. It turned out the issue was on my side — adding sharex=False and sharey=False solved it: fig = uplt.figure(dpi=400, space=3, sharex=False, sharey=False). Also, I noticed that uplt.GridSpec and uplt.subplots provide similar functionality. In what specific scenarios should each be used?

cvanelteren:
I will do some investigation. Using `subplots` is a much more succinct way to create layouts. It will create a gridspec from a complex layout or regular grid. Using a gridspec is a manual approach that is more cumbersome In my opinion but should operate in the same manner (which it clearly didn't do as with subplots the above does not happen)

cvanelteren:
Can confirm the bug. Just pushed a fix. Apologies for the inconvenience.

nickbsb:
Thank you very much for your reply. I have one more small question.

I defined the colorbar like this:

![Image](https://github.com/user-attachments/assets/ab05c7a7-d5b0-4678-aa8b-509a498f82e4)
My requirement is not to display the ticks on the colorbar, and to adjust the spacing between the tick labels and the colorbar.

It seems that the only way to achieve this is by using:

![Image](https://github.com/user-attachments/assets/3cb0feb3-d507-4a8b-b82d-7557a4d86181)
I couldn't find a corresponding parameter for this in ultraplot's colorbar interface. Is that correct?

cvanelteren:
[...]

PR #253, opened 2025-06-09
GeoTicks not responsive

cvanelteren:
Addresses #252 

image

```python
# %%
import ultraplot as uplt, numpy as np, pandas as pd
import numpy as np
import matplotlib.patheffects as pe

fig, ax = uplt.subplots(proj="cyl")
ax.format(
    latlabels="l",
    lonlabels="b",
)
uplt.show(block=1)
```

### DIFF
--- a/ultraplot/figure.py
+++ b/ultraplot/figure.py
@@ -913,14 +913,16 @@ class Figure(mfigure.Figure):
         containing a list of axes on that border.
         """
 
+        border_axes = dict(top=[], bottom=[], left=[], right=[])
         gs = self.gridspec
+        if gs is None:
+            return border_axes
 
         # Skip colorbars or panels etc
         all_axes = [axi for axi in self.axes if axi.number is not None]
 
         # Handle empty cases
         nrows, ncols = gs.nrows, gs.ncols
-        border_axes = dict(top=[], bottom=[], left=[], right=[])
         if nrows == 0 or ncols == 0 or not all_axes:
             return border_axes
         # We cannot use the gridspec on the axes as it
--- a/ultraplot/figure.py
+++ b/ultraplot/figure.py
@@ -1315,7 +1315,8 @@ class Figure(mfigure.Figure):
                 print(axi.number)
                 axi._toggle_gridliner_labels(**turn_on_or_off)
             else:
-                axi.tick_params(which=which, **turn_on_or_off)
+                continue
+                # axi.tick_params(which=which, **turn_on_or_off)
 
     def _toggle_axis_sharing(
         self,
--- a/ultraplot/figure.py
+++ b/ultraplot/figure.py
@@ -1276,6 +1276,10 @@ class Figure(mfigure.Figure):
         Helpers function to ensure the labels
         are shared for rectilinear GeoAxes.
         """
+        # Only apply sharing of labels when we are
+        # actually sharing labels.
+        if self._get_sharing_level() == 0:
+            return
         # Turn all labels off
         # Note: this action performs it for all the axes in
         # the figure. We use the stale here to only perform
@@ -1287,8 +1291,6 @@ class Figure(mfigure.Figure):
             for axi in axes:
                 recoded[axi] = recoded.get(axi, []) + [direction]
 
-        # We turn off the tick labels when the scale and
-        # ticks are shared (level >= 3)
         are_ticks_on = False
         default = dict(
             labelleft=are_ticks_on,
@@ -1969,7 +1971,7 @@ class Figure(mfigure.Figure):
             ax.number = store_old_number
         # When we apply formatting to all axes, we need
         # to potentially adjust the labels.
-        if len(axs) == len(self.axes) and self._get_sharing_level() > 0:
+        if len(axs) == len(self.axes):
             self._share_labels_with_others()
 
         # Warn unused keyword argument(s)
--- a/ultraplot/figure.py
+++ b/ultraplot/figure.py
@@ -1312,6 +1312,7 @@ class Figure(mfigure.Figure):
                     turn_on_or_off[sidelabel] = True
 
             if isinstance(axi, paxes.GeoAxes):
+                print(axi.number)
                 axi._toggle_gridliner_labels(**turn_on_or_off)
             else:
                 axi.tick_params(which=which, **turn_on_or_off)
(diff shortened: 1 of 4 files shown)


cvanelteren:
This PR combines partly the changes from #244. The difference between these tow PRs is that this just targets GEO whereas #244 also combines this with more profound axis sharing which I will target at a later date.

cvanelteren:
Looks good. The failure of the test is expected given the changes.

ultraplot/tests/test_geographic.py
```diff
@@ -758,3 +760,68 @@ def test_inset_axes_geographic():
...
+        assert ax[0]._is_ticklabel_on(label) != False
+        ax[0]._toggle_gridliner_labels(**{label: False})
+        assert ax[0]._is_ticklabel_on(label) != True
+    uplt.close(fig)
+
+
+@pytest.mark.mpl_image_compare
+def test_sharing_cartopy_with_colorbar():
+
+    def are_labels_on(ax, which=["top", "bottom", "right", "left"]) -> tuple[bool]:
```
beckermr, line 786:
```suggestion
    def are_labels_on(ax, which=("top", "bottom", "right", "left")) -> tuple[bool]:
```

Using mutable objects as function defaults is a big anti-pattern in python and will causes unexpected side effects.

beckermr:
```suggestion
    def are_labels_on(ax, which=("top", "bottom", "right", "left")) -> tuple[bool]:
```

Using mutable objects as function defaults is a big anti-pattern in python and will causes unexpected side effects.

ultraplot/tests/test_geographic.py
```diff
@@ -758,3 +760,68 @@ def test_inset_axes_geographic():
...
+    n = 3
+    settings = dict(land=True, ocean=True, labels="both")
+    fig, ax = uplt.subplots(ncols=n, nrows=n, share="all", proj="cyl")
+    ax.format(**settings)
+    fig, ax = uplt.subplots(
+        ncols=3,
+        nrows=3,
+        proj="cyl",
+        share="all",
+    )
```
beckermr: This block of code can't be right as it makes fig and ax twice. I think there is some leftover code from another test or something?

ultraplot/tests/test_geographic.py
```diff
@@ -253,14 +255,7 @@ def are_labels_on(ax, which=["top", "bottom", "right", "left"]) -> tuple[bool]:
...
     settings = dict(land=True, ocean=True, labels="both")
     fig, ax = uplt.subplots(ncols=n, nrows=n, share="all", proj="cyl")
-    # Add data and ensure the tests still hold
-    # Adding a colorbar will change the underlying gridspec, the
-    # labels should still be correctly treated.
-    data = np.random.rand(10, 10)
-    h = ax.imshow(data)[0]
-    fig.colorbar(h, loc="r")
     ax.format(**settings)
-    fig.canvas.draw()  # need a draw to trigger ax.draw for  sharing
```
beckermr, line 263: Why was this code removed? The comment indicates that it is important to the test.

ultraplot/tests/test_geographic.py
```diff
@@ -554,12 +554,14 @@ def assert_views_are_sharing(ax):
...
         fig, ax = uplt.subplots(ncols=2, nrows=2, proj="cyl", share=level)
-        ax.format(labels="both")
+
         for axi in ax:
             axi.format(
                 lonlim=lonlim * axi.number,
                 latlim=latlim * axi.number,
             )
+        ax.format(labels="both")
+
```
beckermr: Please add a comment to the code as to why this call to format was moved.

ultraplot/figure.py
```diff
@@ -1268,6 +1271,56 @@ def _unshare_axes(self):
...
+                    assert sidelabel in turn_on_or_off
+                    turn_on_or_off[sidelabel] = True
+
+            if isinstance(axi, paxes.GeoAxes):
+                print(axi.number)
+                axi._toggle_gridliner_labels(**turn_on_or_off)
+            else:
+                # TODO: implement
+                continue
+                # axi.tick_params(which=which, **turn_on_or_off)
```
beckermr: This section of code appears to have a debug statement. It also has a big TODO item and I am concerned that if we merge it will leave some bits of non-geographic axes not working. Is that true? If not, can you add a comment as to why?

ultraplot/figure.py
```diff
@@ -1268,6 +1271,56 @@ def _unshare_axes(self):
...
+        # it once as it is an expensive action.
+        border_axes = self._get_border_axes()
+        # Recode:
+        recoded = {}
+        for direction, axes in border_axes.items():
+            for axi in axes:
+                recoded[axi] = recoded.get(axi, []) + [direction]
+
+        # We turn off the tick labels when the scale and
+        # ticks are shared (level >= 3)
```
beckermr: Please add a comment indicating where in the code the sharing level is checked. If it is not checked, it should be.

ultraplot/tests/test_geographic.py
cvanelteren: Thanks!

ultraplot/tests/test_geographic.py
cvanelteren: It is moved to a separate test `test_sharing_cartopy_with_colorbar`

ultraplot/tests/test_geographic.py
cvanelteren: Should be placed back. There was an issue I ended up fixing. Thanks!

ultraplot/figure.py
cvanelteren: Done.

ultraplot/figure.py
cvanelteren: It is checked  in line 1972, but will move it internal

ultraplot/tests/test_geographic.py
cvanelteren: Think this happened when I separated the two tests.

ultraplot/figure.py
cvanelteren: Done!

ultraplot/figure.py
cvanelteren: For extra motivation, I haven't had time to fully flush out #244. As GeoAxes and Cartesian behave differently in terms of how the labels are shared, this function should not affect how Cartesian plots share their label. My aim is to in the end unify these approaches to have a singular interface that would work for all axes rather than how it is handled currently. This is just a step into this direction rather than tackling everything in one go.

beckermr (approved):

beckermr:
testing error is expected and new plot looks better

PR #254, opened 2025-06-11
Refactor beeswarm

cvanelteren:
This PR is a major rework of the `beeswarm` functionality that was recently merged to main. While testing the newly added implementation with examples from the SHAP library, I discovered that our current approach doesn't integrate well with their plotting workflows and becomes computationally expensive for larger datasets. The former implementation processed each point individually to reduce overlap, which scales poorly as dataset size increases.

This rework improves performance by vectorizing the overlap reduction function, replacing the previous O(n²) per-point approach with vectorized operations that scale better with large datasets. Our implementation remains intentionally barebones compared to the SHAP library's more sophisticated feature clustering capabilities. For users requiring advanced clustering features, we recommend the specialized SHAP library, which can achieve similar visual results with additional functionality.

The docstring is also corrected to be in line with the signature of the function. Lastly, the tests are updated to reflect the changes in the api and rely more on internal functions.

image

```python
import xgboost, shap

# # # train XGBoost model
X, y = shap.datasets.adult()
model = xgboost.XGBClassifier().fit(X, y)

# # # compute SHAP values
explainer = shap.Explainer(model, X)
shap_values = explainer(X)
# %%
print(shap_values.data.shape)
c = shap_values.data
c = (c - c.min(0)[None]) / (c.max(0) - c.min(0))[None]

fig, ax = uplt.subplots()
ax.beeswarm(shap_values.values, feature_values=c, cmap="burd", ss=1)
ax.format(
    xlabel="SHAP values",
    yticks=np.arange(c.shape[1]),
    yticklabels=shap_values.feature_names,
)

uplt.show(block=1)
```
</details>

### DIFF
--- a/ultraplot/tests/test_1dplots.py
+++ b/ultraplot/tests/test_1dplots.py
@@ -685,9 +685,8 @@ def test_beeswarm():
         data[:, cat] = np.random.normal(cat * 1.5, 0.6, n_points)
         feature_values[:, cat] = np.random.randn(n_points)
 
-    fig, (ax1, ax2, ax3, ax4) = uplt.subplots(
-        nrows=2,
-        ncols=2,
+    fig, (ax1, ax2, ax3) = uplt.subplots(
+        ncols=3,
         share=0,
     )
 
@@ -726,22 +725,8 @@ def test_beeswarm():
         yticks=categories,
         yticklabels=["Group A", "Group B", "Group C", "Group D"],
     )
-
-    # A test to ensure that code using a singular x is also working
-    ax3.beeswarm(
-        data,
-        levels[0],
-        feature_values=feature_values,
-        colorbar="ur",
-        colorbar_kw=dict(
-            title="Feature Score",
-        ),
-    )
+    ax3.beeswarm(data[:, 0], levels[:, 0])
     ax3.format(
-        title="Singular X with groups",
-    )
-    ax4.beeswarm(data[:, 0], levels[:, 0])
-    ax4.format(
         title="Singular Beeswarm",
     )
     return fig


cvanelteren:
For reference this is the plot generated from the same dataset in the shap library:

image

cvanelteren:
Note to self I may want to change the max lineage to be slightly less than half the level width.

cvanelteren:
Unittests are a bit different visually but act similarly.

cvanelteren:
That is, I was not expecting them to behave the same as on main I computed the colours based on the height, which is now handled implicitly through scatter functionality.

beckermr:
No there is an actual error

```
FAILED ultraplot/tests/test_1dplots.py::test_beeswarm - AttributeError: PathCollection.set() got an unexpected keyword argument 'size'
```

cvanelteren:
Ah ok will fix.

cvanelteren:
Should be good now. When merged I will draft a release.

cvanelteren:
I changed the test. The colors were first inferred and now delegated to scatter so it follow the prop cycle.

cvanelteren:
Will need to change the bottom left but the rest looks fine to me.

beckermr:
Did you check that the new and old code give consistent results? Changing the code and the test means that we cannot tell.

cvanelteren:
I can check it; the reason why i changed is to ensure that the prop cycle is used as it's more consistent with the rest of the api. I can temp change it to confirm.

cvanelteren:
image
The test is visualized next to the mocked top left plot using this PR.  The original has slightly more jitter since we were modifying the x and y value randomly from the current position, here we just modify the x or y value (depending on the orientation, for this example just x). 

I will adjust the test to mimic the original and call it a day.

cvanelteren:
Drafting a release now introducing the new feature.

PR #256, opened 2025-06-12
feat: advanced axis sharing refactor + enhancements

cvanelteren:
Rename of #244 

One of UltraPlot's most powerful features is the ability to do advanced axis sharing. The current implementation works well for bottom and left label sharing, but failed in sharing labels for top right labels. This PR addresses this problem while at the same time allowing for sharing axes with more complex layouts.

### DIFF
--- a/ultraplot/axes/plot.py
+++ b/ultraplot/axes/plot.py
@@ -2509,7 +2509,7 @@ class PlotAxes(base.Axes):
         # functions and helper functions.
         parsers = (self._parse_cmap, *self._level_parsers)
         if c is None or mcolors.is_color_like(c):
-            if infer_rgb and c is not None:
+            if infer_rgb and c is not None and c != "none":
                 c = pcolors.to_hex(c)  # avoid scatter() ambiguous color warning
             if apply_cycle:  # False for scatter() so we can wait to get correct 'N'
                 kwargs = self._parse_cycle(**kwargs)
--- a/ultraplot/tests/test_plot.py
+++ b/ultraplot/tests/test_plot.py
@@ -420,3 +420,16 @@ def test_pie_labeled_series_in_dataframes():
     for text, index in zip(texts, data.index):
         assert text.get_text() == index
     uplt.close(fig)
+
+
+def test_color_parsing_for_none():
+    """
+    Ensure that none is not parsed to white
+    """
+    fig, ax = uplt.subplots()
+    ax.scatter(0.4, 0.5, 100, fc="none", ec="k", alpha=0.2)
+    ax.scatter(0.5, 0.5, 100, fc="none", ec="k")
+    ax.scatter(0.6, 0.5, 100, fc="none", ec="k", alpha=1)
+    for artist in ax[0].collections:
+        assert artist.get_facecolor().shape[0] == 0
+    uplt.close(fig)


cvanelteren:
~~Need to fix the gridspec position by adding a new function to subplotspec or gridspec~~

cvanelteren:
Ok this PR is shaping up to something. There are some minor bugs to hash out but I think I am close to finalizing this.

cvanelteren:
~~Still need to  "copy" the axis that are on for the side panels.~~ Done.

for example

```python

# need to fix the grid not being parse correctly (again)
# Use property cycle for columns of 2D input data
fig, ax = uplt.subplots()
ax.format(xtickloc="top", ytickloc="right")
for i in range(2):
    ax.panel("t")

for i in range(2):
    ax.panel("r")

uplt.show()

```

Needs to have the x tick on t he right panels to be moved to the top.

cvanelteren:
I think this is ready for review. We have to be careful when merging this as it profoundly changes how the labels are treated compared to default matplotlib and previous ultraplot versions. 

The PR generalizes the visibility of the ticklabels such that sharing is correct when the labels are top or right, e.g.


test

```python
fig, ax = uplt.subplots(ncols=2)
ax.format(xtickloc="t", ytickloc="r")
```
</details>



It also handles panel but with a caveat. The labels on the main axis are "pushed outwards". By default ultraplot did this only for bottom and left labels, but removed the right labels. Now these are shown by default. Take for example

```python
fig, ax = uplt.subplots()
ax.panel("t")
ax.panel("r")
```

The labels for "x" are pushed outwards to the top and for y to the right and they are turned on. **They can be toggled off post-hoc which was not possible for ultraplot right now**. This is a change to how it is currently handled, but I _feel_ like this approach makes more sense -- open for feedback however. The reason why I think this makes more sense that panels usually are concerning the plot it is attached to. This mean that sharing the tick labels make little sense otherwise you could achieve the same with an extra plot next  to it.


image

cvanelteren:
Oh minor note -- in case it wasn't clear from the last comment, this is why some tests now differ in terms of the output.

cvanelteren:
Failing tests look good -- there is still a minor issue with the seed no passing through correctly to the network functions.

~~Need to check why `test_contour_legend_without_label` is not showing the proper difference.~~

cvanelteren:
Problem was not with cleaning the directory. 

I am leaving the errors in to indicate in the review what the changes are but we should change them on agreement on how the visuals of the new behavior. If accepted we need to change the docs as well to remain visual fidelity in those images or leave them to signal new behavior.

cvanelteren:
image
this works now properly

cvanelteren:
When I compare with main locally I am not getting this error:



https://github.com/user-attachments/assets/af959fb3-0a4d-4d24-a693-49c0f404060b

cvanelteren:
what I did was delete `baseline` and `results`; checkout main; generate baselines and then go to this PR and compare with baseline.

beckermr:
Well something is different in the CI which indicates a bug or some environment effect. The code needs to work everywhere, not just locally on your machine.

cvanelteren:
At least the network error is gone now :partying_face:

cvanelteren:
Figured out the problem see #274

beckermr:
OK I can proceed on review of this PR.

cvanelteren:
Think the tests that fail are looking good last I checked (prior to latest merge). The issue addresses in #233  is not completely fixed since the partial reversal that happened.

ultraplot/tests/test_axes.py
```diff
@@ -259,51 +259,104 @@ def test_sharing_labels_top_right():
...
+@pytest.mark.mpl_image_compare
+def test_alt_axes_x_shared():
+    layout = [[1, 2], [3, 4]]
+    fig, ax = uplt.subplots(ncols=2, nrows=2)
+
+    for axi in ax:
+        alt = axi.altx()
+        alt.set_xlabel("Alt X")
+        axi.set_xlabel("X")
+    return fig
```
beckermr: For these tests it'd be good to add assert statements that the labels are set correctly maybe?

ultraplot/tests/test_subplots.py
```diff
@@ -215,3 +241,33 @@ def test_axis_sharing(share):
...
+    return fig
+
+
+@pytest.mark.parametrize("layout", [[1, 2], [3, 4]])
+@pytest.mark.mpl_image_compare
+def test_panel_sharing_top_right(layout):
+    fig, ax = uplt.subplots(layout)
+    for dir in "left right top bottom".split():
+        ax[0].panel(dir)
+    return fig
```
beckermr: These tests could also use assert statements that the labels and ticks are set correctly?

ultraplot/figure.py
```diff
@@ -1985,6 +1948,53 @@ def format(
...
+            labelbottom=are_ticks_on,
+        )
+        for axi in self._iter_axes(hidden=False, panels=False, children=False):
+            # Turn the ticks on or off depending on the position
+            sides = recoded.get(axi, [])
+            turn_on_or_off = default.copy()
+            # The axis will be a border if it is either
+            # (a) on the edge
+            # (b) not next to a subplot
+            # (c) not next to a subplot of the same kind
```
beckermr, line 1983: This comment seems out of place as none of these items are checked here.

ultraplot/figure.py
```diff
@@ -1985,6 +1948,53 @@ def format(
...
+        # it once as it is an expensive action.
+        border_axes = self._get_border_axes(same_type=False)
+        # Recode:
+        recoded = {}
+        for direction, axes in border_axes.items():
+            for axi in axes:
+                recoded[axi] = recoded.get(axi, []) + [direction]
+
+        # We turn off the tick labels when the scale and
+        # ticks are shared (level >= 3)
```
beckermr: I don't see the level checked here. Is that checked somewhere else?

ultraplot/figure.py
beckermr, line 1984: This comment seems out of place as none of these items are checked here.

ultraplot/tests/test_subplots.py
cvanelteren: I will add it to be more explicit but it will be caught by the image compare, thanks.

cvanelteren: Just found a bug which I need to address, but added the checks

ultraplot/tests/test_axes.py
cvanelteren: Forgot these momentito

cvanelteren: done

cvanelteren:
Will need to look at `layout1` for `test_label_sharing_top_right` but not no as its getting late ;-)

ultraplot/utils.py
```diff
@@ -904,6 +907,185 @@ def units(
     return result[0] if singleton else result
 
 
+def _get_subplot_layout(
+    gs: "GridSpec",
+    all_axes: list["paxes.Axes"],
+    same_type=True,
+) -> tuple[np.ndarray[int, int], np.ndarray[int, int], dict[type, int]]:
+    """
+    Helper function to determine the grid layout of axes in a GridSpec. It returns a grid of axis numbers and a grid of axis types. This function is used internally to determine the layout of axes in a GridSpec.
```
beckermr: Doc string is too long and needs wrapping.

ultraplot/utils.py
```diff
@@ -904,6 +907,185 @@ def units(
     return result[0] if singleton else result
 
 
+def _get_subplot_layout(
```
beckermr, line 910: This whole function needs many more comments. What is axis_type = 1 and why is it hard coded?

ultraplot/utils.py
```diff
@@ -904,6 +907,185 @@ def units(
...
+
+@dataclass
+class _Crawler:
+    """
+    A crawler is used to find edges of axes in a grid layout.
+    This is useful for determining whether to turn shared labels on or depending on the position of an axis in the gridspec.
+    It crawls over the grid in all four cardinal directions and checks whether it reaches a border of the grid or an axis of a different type. It was created as adding colorbars will
+    change the underlying gridspec and therefore we cannot rely
+    on the original gridspec to determine whether an axis is a border or not.
+    """
```
beckermr: Doc string needs to be wrapped.

beckermr:
@cvanelteren the test failures here need looking at

cvanelteren:
Will have a closer look tomorrow. For me these tests need TLC:
- test_inset_basic
- test_share_all_basic

beckermr:
I see these as well

```
FAILED ultraplot/tests/test_subplots.py::test_label_sharing_top_right[layout0] - AssertionError: assert False == True
 +  where False = _is_ticklabel_on('labeltop')
 +    where _is_ticklabel_on = CartesianAxes(index=(0, 0), number=1)._is_ticklabel_on
FAILED ultraplot/tests/test_subplots.py::test_label_sharing_top_right[layout1] - AssertionError: assert False == True
 +  where False = _is_ticklabel_on('labelright')
 +    where _is_ticklabel_on = CartesianAxes(index=(0, 0), number=1)._is_ticklabel_on
```

cvanelteren:
> I see these as well
> 
> ```
> FAILED ultraplot/tests/test_subplots.py::test_label_sharing_top_right[layout0] - AssertionError: assert False == True
>  +  where False = _is_ticklabel_on('labeltop')
>  +    where _is_ticklabel_on = CartesianAxes(index=(0, 0), number=1)._is_ticklabel_on
> FAILED ultraplot/tests/test_subplots.py::test_label_sharing_top_right[layout1] - AssertionError: assert False == True
>  +  where False = _is_ticklabel_on('labelright')
>  +    where _is_ticklabel_on = CartesianAxes(index=(0, 0), number=1)._is_ticklabel_on
> ```

For that particular plot, I need some input. The test produces: 
image
The issue is in how the axis sharing is applied. For the middle plot it is not sharing any axis so it turns the top and right label on. For the bottom left and right, it is sharing x with the top plots and hence is not showing the labels on the top. Similarly the plot on the top left is sharing with the top right.

The current logic is looking for border that are either the edges of the subplot grid or when a plot is not adjacent to another plot (in that direction). The new border detection logic is correctly identifying the borders of each subplot, but it is overriden by the `apply_axis_sharing` logic.  

This particular layout, lends itself most to just turning the axis sharing off (I think), but we should allow users to still perform the sharing. This would mean that we should adjust the test accordingly, or adjust the apply axis sharing logic. Not sure where to proceed here.

beckermr:
OK I lost track of this PR. Sorry. 

What is the share level for this test? The docs currently say


> Supporting five sharing “levels”. These values can be passed to sharex, sharey, or share, or assigned to rc['subplots.share']. The levels are defined as follows:
> 
> - False or 0: Axis sharing is disabled.
> 
> - 'labels', 'labs', or 1: Axis labels are shared, but nothing else. Labels will appear on the leftmost and bottommost subplots.
> 
> - 'limits', 'lims', or 2: Same as 1, but axis limits, axis scales, and major and minor tick locations and formatting are also shared.
> 
> - True or 3 (default): Same as 2, but axis tick labels are also shared. Tick labels will appear on the leftmost and bottommost subplots.
> 
> - 'all' or 4: Same as 3, but axis limits, axis scales, and axis ticks are shared even between subplots not in the same row or column.

For the middle panel in each sharing level we should have

- level 0: display limits+ticks+labels
- level 1: display limits+ticks, no axis label
- level 2: display limits+ticks, no axis label
- level 3: display limits+ticks because it is the outer most panel for both vertical and horizontal direction for its row+column, no axis label
- level 4: same as level 3

cvanelteren:
Thanks will have a careful look tomorrow and update you on the details to finalize this PR

cvanelteren:
So  I had to some time to settle back in and ready to pick this up again. 

I think for the label sharing level, we should change the docs to state that it will share with the outermost labels. For left and bottom labels that would be the leftmost and bottomost, and for top and right it will be topmost and rightmost. 

Now there are the edgecases where a plot is not flanked directly by another plot. In this case, I think it makes most sense to also add the labels there -- do you agree @beckermr? 

This would mean that the code and docs would need to reflect this change, which won't affect dense grids but will more complex layouts.

beckermr:
I agree in general unless the labels are duplicated like in the image above. In that case, we should keep only the one that applies to the full plot. This should be what I wrote above.

cvanelteren:
Ok great will focus on fixing the PR and will ping you again for review.

cvanelteren:
image
These tests fails but require some input on what we allow. We are seeing that the outermost panel now has the ticks turned on. I think this makes sense as the panel belongs to a singular plot, but we may want to return to the old behavior.

Let me know what you think @beckermr

beckermr:
I think for panels we should return to the old behavior.

cvanelteren:
All green ;-)!

docs/subplots.py
```diff
@@ -377,7 +377,12 @@
 #
 #    * ``False`` or ``0``: Axis sharing is disabled.
 #    * ``'labels'``, ``'labs'``, or ``1``: Axis labels are shared, but
-#      nothing else. Labels will appear on the leftmost and bottommost subplots.
+#      nothing else. Labels will appear on the outermost plots. This implies that for left, and bottom labels (default) the labels will appear on the leftmost and bottommost subplots
```
beckermr: Let's split this comment across multiple lines.

Issue #259, opened 2025-06-14
unexpected keyword argument 'legend_kw' of pie plot if DataFrame has index name.


gepcel:
ultraplot: 1.56

When plotting a pie chart using a DataFrame with a named index, the error unexpected keyword argument 'legend_kw' occurs (other plot types untested). Examples:

1. Works (no index name):
```python
d1 = pd.DataFrame(index=list('abcd'),data=dict(y=range(1,5)))
fig, ax = pplt.subplot()
ax.pie('y', data=d1)
```

2. Fails (with index name):
```python
d2 = pd.DataFrame(index=pd.Index(list('abcd'), name='x'),data=dict(y=range(1,5)))
fig, ax = pplt.subplot()
ax.pie('y', data=d2)  # Raises: TypeError: unexpected keyword argument 'legend_kw'
```

error msg:
```
TypeError                                 Traceback (most recent call last)
File test.py:46
     44 d2 = pd.DataFrame(index=pd.Index(list('abcd'), name='x'),data=dict(y=range(1,5)))
     45 fig, ax = pplt.subplot()
---> 46 ax.pie('y', data=d2)

File C:\Anaconda3\Lib\site-packages\ultraplot\internals\inputs.py:365, in _preprocess_or_redirect.<locals>._decorator.<locals>._preprocess_or_redirect(self, *args, **kwargs)
    363     kwargs["color"] = color
    364 # Call main function
--> 365 return func(self, *args, **kwargs)

File C:\Anaconda3\Lib\site-packages\ultraplot\axes\plot.py:4505, in PlotAxes.pie(self, x, explode, labelpad, labeldistance, **kwargs)
   4501 _, x, kw = self._parse_1d_args(
   4502     x, autox=False, autoy=False, autoreverse=False, **kw
   4503 )
   4504 kw = self._parse_cycle(x.size, **kw)
-> 4505 objs = self._call_native(
   4506     "pie", x, explode=explode, labeldistance=pad, wedgeprops=wedge_kw, **kw
   4507 )
   4508 objs = tuple(cbook.silent_list(type(seq[0]).__name__, seq) for seq in objs)
   4509 self._fix_patch_edges(objs[0], **edgefix_kw, **wedge_kw)

File C:\Anaconda3\Lib\site-packages\ultraplot\axes\plot.py:1531, in PlotAxes._call_native(self, name, *args, **kwargs)
   1529         obj = getattr(self.projection, name)(*args, ax=self, **kwargs)
[...]

PR #260, opened 2025-06-14
Fix unused parameters being passed to pie chart

cvanelteren:
Addresses #259 

Will need to check this later today or tomorrow as I don't have time to check this now.


cvanelteren:
~~TODO: Still needs a test~~

cvanelteren:
Not entirely sure why for dataframes the code is forwarding the `legend_kw` or `colorbar_kw` but removing this will cause other tests to fail. What remains sure is that for pie charts these need not be forwarded; so the major contribution is to eliminate those from kwargs prior to plotting with matplotlib's pie function.

beckermr (approved):

PR #266, opened 2025-06-17
Add xdist to image compare

cvanelteren:
This is a continuation of #190. 

I added some supporting code to conftest to allow our tests to run in parallel. In addition it adds a custom html to show the results and is compliant with our approach of only keeping the failed images.

I also sorted the dependencies alphabetically in environment.yml to reduce my OCD ;-)

### DIFF
--- a/.github/workflows/build-ultraplot.yml
+++ b/.github/workflows/build-ultraplot.yml
@@ -43,7 +43,7 @@ jobs:
 
       - name: Test Ultraplot
         run: |
-          pytest --cov=ultraplot --cov-branch --cov-report term-missing --cov-report=xml ultraplot
+          pytest -n auto --cov=ultraplot --cov-branch --cov-report term-missing --cov-report=xml ultraplot
 
       - name: Upload coverage reports to Codecov
         uses: codecov/codecov-action@v5
@@ -76,14 +76,14 @@ jobs:
           git fetch origin ${{ github.event.pull_request.base.sha }}
           git checkout ${{ github.event.pull_request.base.sha }}
           python -c "import ultraplot as plt; plt.config.Configurator()._save_yaml('ultraplot.yml')"
-          pytest -W ignore --mpl-generate-path=baseline --mpl-default-style="./ultraplot.yml"
+          pytest -n auto -W ignore --mpl-generate-path=baseline --mpl-default-style="./ultraplot.yml"
           git checkout ${{ github.sha }}  # Return to PR branch
 
       - name: Image Comparison Ultraplot
         run: |
           mkdir -p results
           python -c "import ultraplot as plt; plt.config.Configurator()._save_yaml('ultraplot.yml')"
-          pytest -W ignore --mpl --mpl-baseline-path=baseline --mpl-generate-summary=html --mpl-results-path=./results/ --mpl-default-style="./ultraplot.yml" --store-failed-only ultraplot/tests
+          pytest -n auto -W ignore --mpl --mpl-baseline-path=baseline --mpl-generate-summary=html --mpl-results-path=./results/ --mpl-default-style="./ultraplot.yml" --store-failed-only ultraplot/tests
 
       # Return the html output of the comparison even if failed
       - name: Upload comparison failures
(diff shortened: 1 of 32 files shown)


cvanelteren:
On the GHA we are running automatically with two workers, not sure if we can force it to use more -- it's not much but it's something I guess. Locally it runs much faster.

cvanelteren:
I had to combat some racing conditions but the code can now reproduce the  results within the PR but may fail when comparing with main due to rc changes that carried over in the tests across tests (which is bad) on main.

cvanelteren:
Locally I am getting a ~8x times speed up for 8 cores. The GHA shows about double which makes sense for 2 workers.

The tests are failing due to the leaking of the random number generator not properly being reset (I think) and the rc params carrying over between tests.  I judged the results visually and they seem ok. 

Note that our tests are now inside `mpl_comparison_report.html`

beckermr:
Are we using threads or processes?

cvanelteren:
> Are we using threads or processes?

threads

beckermr:
OK. We really should figure out a way to ensure things do not leak. If things are leaking, then we likely won't get consistent test results due to race conditions. That is bad.

cvanelteren:
We currently have persistent changes in the tests that are changing rc params which is influencing later plots. I initially copied a lot of tests from Luke's test branch and added a bunch from the doc test, but didn't clean them properly -- which is not great.

beckermr:
Would using context managers for the RC params help? IDK if those are thread safe or not.

cvanelteren:
I am separating the new html template and then opening it up for review.

cvanelteren:
> Would using context managers for the RC params help? IDK if those are thread safe or not.

That's what I currently did. Will double check if that is thread safe.

beckermr:
For the RNGs, I wonder if something in pytest is getting the fixtures confused. Maybe we move to making the RNGs directly within each test?

cvanelteren:
> For the RNGs, I wonder if something in pytest is getting the fixtures confused. Maybe we move to making the RNGs directly within each test?

I am creating a new rng for each thread now.
I just checked our context manager is not thread safe. Will add it.

cvanelteren:
I am still expecting the tests will fail when comparing with main. Race conditions are gone locally :partying_face:  and the test suite is more robust. I may wanna move the plugin to another repo outside the main repository as it contains a lot more code than I was expecting when I started working on it.

beckermr:
I think the new rng test fixture is using a different effective seed. We should figure that out.

cvanelteren:
> I think the new rng test fixture is using a different effective seed. We should figure that out.

Should be the same no?  See https://github.com/Ultraplot/UltraPlot/blob/f0bd292d8472776336eb6ca83b88defa908adb1e/ultraplot/tests/conftest.py#L11

beckermr:
I think this line

```
return np.random.Generator(np.random.PCG64(SEED))
```

may need to be 

```
return np.random.default_rng(seed=SEED)
```

The image comparison tests are failing right now because the data in them is different. So something broke.

cvanelteren:
I believe these are the same for numpy, i.e. the default rng is `PCG64` but can change it.

beckermr:
Ohhhh I thought it was the MT one. Hrrmmmmm ok. We need to figure out why the tests now differ.

cvanelteren:
I am suspecting that the seed was not properly resetting in its current implementation which makes comparison hard as we don't know when a thread runs a particular test (the racing condition). 

I can confirm that generating the baseline from the PR and then running a comparison on that data will always match which indicates that the racing condition is not the problem, but the comparison with main will still fail.

I think what we could do is first merge a PR that ensures the numpy rng is set properly on main and then re-compare this with this PR, what do you think?

cvanelteren:
I will do that now -- time permitting. At least this PR is working with xdist now nicely which is great and the speed ups are very nice for local development :100:

cvanelteren:
I am going ahead an merging this as the issue is again related to the rng generation. I will confirm with the merged results offline is the issue still persists.

cvanelteren:
Found no issues other than then network layout not properly passing the seed, see #256

PR #268, opened 2025-06-17
set rng per test

cvanelteren:
To ease the transition of #266 this PR adds just the changes to the tests and the local rng generator.

### DIFF
--- a/ultraplot/tests/test_1dplots.py
+++ b/ultraplot/tests/test_1dplots.py
@@ -537,7 +537,6 @@ def test_networks(rng):
     # using the same seed
     from .conftest import SEED
 
-    np.random.seed(SEED)
     graphs = [
         nx.karate_club_graph(),
         nx.florentine_families_graph(),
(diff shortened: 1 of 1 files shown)


ultraplot/tests/test_1dplots.py
```diff
@@ -527,13 +526,18 @@ def test_heatmap_labels():
...
     Create a baseline network graph that tests
     a few features. It is not the prettiest graphs but highlights what the functions can do.
     """
     import networkx as nx
 
+    # Ensure that the seed for the networkx are
+    # using the same seed
+    from .conftest import SEED
+
+    np.random.seed(SEED)
```
beckermr: This will need a threading lock since we are modifying a global.

beckermr (changes requested):

ultraplot/tests/test_1dplots.py
cvanelteren: ~~I can't thread lock the layout within networkx, so this is fine as is I think as no two threads are accessing that function~~

cvanelteren: let me think of a better way to do this; otherwise we need to lock all functions that call `graph`

cvanelteren: We can use `loadscope` on the other PR https://pytest-xdist.readthedocs.io/en/stable/distribution.html#running-tests-across-multiple-cpus

cvanelteren: Think it is fixed now if we pass on seed explicitly to the functions that use rngs

cvanelteren:
From my viewing of the results, the results only differ in the random data. This PR and #266 would self correct as it would clean up the mess on main.

Issue #271, opened 2025-06-19
Setting `ticklen` parameter changes map extent in CartopyAxes


gepcel:
## Description
When using `CartopyAxes`, setting the `ticklen` parameter appears to unexpectedly modify the map extent.

## Steps to Reproduce
```python
import proplot as pplt

fig, ax = pplt.subplots(ncols=2, proj='cyl', share=False)
ax.format(
    lonlim=(10, 20), latlim=(10, 20),
    lonlines=3, latlines=3,
    lonlabels='both', latlabels='both'
)
# Now change ticklen of ax[1], cause extent change
ax[1].format(ticklen=.5)
# show the change
for a in ax:
    a.text(.5, .5, a.get_extent(), transform=a.transAxes, ha='center', va='center')
```

## Expected Behavior
The map extent should remain consistent between subplots regardless of the ticklen parameter setting.

## Actual Behavior
The subplot with modified ticklen shows a different extent compared to the unmodified subplot.

## Environment

* UltraPlot version: 1.57
* Cartopy version: 0.24.1
* Matplotlib version: 3.9.4

![Image](https://github.com/user-attachments/assets/817861f6-7ef9-40ae-9ee1-32325f5b2c6f)

cvanelteren:
Just a sanity check, your snippet states proplot -- I assume you ran this with ultraplot ;-)

cvanelteren:
Issue still persists on ultraplot will need to have a look at this a hack to work around this would to do the ticks first, but I agree that this should not happen.

cvanelteren:
I think I know what is happening. When you set the `ticklen` a `MaxNLocator` is created which divides the visible range. Since your ticks are every 3 degrees, it can't "nicely" fit this within your range so it adjusts your extent.  Not sure if this is a bug or a feature.  The `MaxNLocator` creates "nice" tick range but are difficult to predict what will be the results afterwards. We had some discussions on this when the ticks were added to the `GeoAxes`. We considered having just a linspace but then the resulting ranges look wrong, i.e. showing a lot of digits are not a nice looking value. 

A solution to this problem would therefore having a range to either the long or lat lines are divisible by it, but then still you may run into this issue.

cvanelteren:
I can see if I can force the extent after setting the ticks to circumvent this issue.

cvanelteren:
Found a fix see #272

PR #272, opened 2025-06-19
Fix issue where view is reset on setting ticklen

cvanelteren:
Addresses #271


cvanelteren:
Ranges are now unchanged
test

PR #273, opened 2025-06-19
Racing condition xdist fix

cvanelteren:
Attempting to set the rng generator explicitly rather than using an integer.

### DIFF
--- a/ultraplot/tests/conftest.py
+++ b/ultraplot/tests/conftest.py
@@ -80,16 +80,16 @@ def isolate_mpl_testing():
     # Create process-specific temporary directory for mpl results
     # This prevents file conflicts between parallel processes
     worker_id = os.environ.get("PYTEST_XDIST_WORKER", "master")
-    temp_dir = tempfile.mkdtemp(prefix=f"mpl_test_{worker_id}_")
-    os.environ["MPL_TEST_TEMP_DIR"] = temp_dir
+    with tempfile.TemporaryDirectory(prefix=f"mpl_test_{worker_id}_") as temp_dir:
+        os.environ["MPL_TEST_TEMP_DIR"] = temp_dir
 
-    yield
+        yield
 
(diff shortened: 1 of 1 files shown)


cvanelteren:
Resolved intermittent failures in `test_contour_legend_with_label` and `test_contour_legend_without_label` when running with `pytest -n 2` by implementing process-specific temporary directories for each xdist worker. The issue was caused by multiple parallel processes writing matplotlib image comparison results to the same file paths, leading to "Image files did not match" errors due to file system race conditions.

cvanelteren:
Note I am not completely sure on the proposed fix here but it cannot hurt. 

The three tests that are failing are:
- ultraplot.tests.test_1dplots.test_networks
- ultraplot.tests.test_legend.test_contour_legend_with_label
- ultraplot.tests.test_legend.test_contour_legend_without_label

Not really sure at this moment what is causing the issue. The network one is highly likely to be caused by the rng which may be fixed in the first commit here. The other two are a bit  of a mystery to me.

beckermr:
I doubt the change from a seed to an RNG will help networkx, but we can try it.

cvanelteren:
Ah damn was still drafting this. Couldn't hurt. Not sure why the last two tests are not working properly. It only happens on python 3.10 with mpl 3.90 I believe.

Not your fault didn't mark it as draft.

beckermr:
Ah sorry. It made the failures on the other PR worse, so we could back it out and try again if you wanted.

cvanelteren:
We can leave this in as I think it improved the situation from a robustness standpoint (the tmp folders); the network seed was probable not necessary but was a sanity check on my end and the decorator had to be fixed anyway.

cvanelteren:
I probed around nx codebase and couldn't directly see how they handled if the seed was an int, hence the generator. In their codebase they immediately called `.rand` on the object.

PR #275, opened 2025-06-19
Replace spring with forceatlas2

cvanelteren:
Until the issue of the decorator is fixed, we replace the layout with one that does not have this issue (and yours truly wrote).


cvanelteren:
This will cause the test to fail but that's ok.

cvanelteren:
Expecting tests fail. Merging for now to ensure one of the three tests are  passing in the future.

PR #276, opened 2025-06-19
Read lock on configurator

cvanelteren:
Attempt to fix the discrepancies between the tests by putting a read lock on the config properties.

### DIFF
--- a/ultraplot/config.py
+++ b/ultraplot/config.py
@@ -774,12 +774,15 @@ class Configurator(MutableMapping, dict):
         Return an `rc_matplotlib` or `rc_ultraplot` setting using dictionary notation
         (e.g., ``value = uplt.rc[name]``).
         """
-        key, _ = self._validate_key(key)  # might issue ultraplot removed/renamed error
-        try:
-            return rc_ultraplot[key]
-        except KeyError:
-            pass
-        return rc_matplotlib[key]  # might issue matplotlib removed/renamed error
+        with self._lock:
+            key, _ = self._validate_key(
+                key
+            )  # might issue ultraplot removed/renamed error
+            try:
+                return rc_ultraplot[key]
+            except KeyError:
+                pass
+            return rc_matplotlib[key]  # might issue matplotlib removed/renamed error
 
     def __setitem__(self, key, value):
         """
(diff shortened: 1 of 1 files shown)


cvanelteren:
~~TODO: replace with copy-on-write~~
- added thread local rc params

cvanelteren:
Not sure if I am making the situation worse. The issue is I think generated by threads modifying rc values which causes issues with other tests. 

The tests that we were initially failing are in particular moddifying the rc and causing issues. Will need to return to this as  this has high priority now.

beckermr:
I suspect that matplotlib itself would need to be thread safe in order for this to work. My guess is that it is not. 

Maybe try an approach where we force a small subset of the tests to be done serially?

cvanelteren:
That could indeed be the case, but I think the edits all go through this configurator object. This would mean that if we protect it, the matplotlib rc file would be protected too. I will go back now and check the thread safety as I was a bit hazzy when I looked at this last night.

cvanelteren:
Ok so the modifications within a context are now threadsafe. However, when a direct modification is performed, i.e.
`uplt.rc["someprop"] = somevalue" it will still modify the main thread. 

I am debating whether the current behavior makes sense. If we are modifying the rc directly from a thread without context, I would resume that the return from this thread should default back to the prior settings. It is currently not doing this only when contexts are used. Now this behavior could be wanted as we are directly modifying the rc param generally with a statement such as

```python
 uplt.rc["fontsize"] = 30
```

This would either force us to be specific in our tests to use only contexts, or we need to modify it such that the direct modification would also work. 
Either way having a thread safe context is nice (:partying_face: ) but not sure on the rest. This effectively creates the same situation as we had before that the tests
- ultraplot.tests.test_legend.test_contour_legend_with_label
- ultraplot.tests.test_legend.test_contour_legend_without_label
still fail when comparing with 5e837bafee9691220e7323a454c734360e97de7c

cvanelteren:
Now I just need to figure out why these tests are still failing...

ultraplot/config.py
```diff
@@ -375,7 +375,7 @@ def _infer_ultraplot_dict(kw_params):
     return kw_ultraplot
 
 
-def config_inline_backend(fmt=None):
+def config_inline_backend(fmt=None, rc_ultraplot=None):
```
beckermr, line 376: The keyword you added is not used in this function.

cvanelteren: line 409

cvanelteren: doesn't matter tho as the tests still show thicker linewidths which indicates an rc leak. I still have not managed to get the ultraplot rc to be thread safe (see newly added test).

beckermr: Ah thanks! I guess there was a global in this module before?

cvanelteren:
I need to separate the initialising of the colormaps before this can be fixed more to follow soon.

ultraplot/config.py
cvanelteren: Yes it  was ;-)

cvanelteren:
The context is now thread safe, but the setting of the parameter is not. I think this is acceptable behavior since it would be difficult to reset to a test without a barrier, which would lead to a similar mechanism as the context.

I think we should move the plugin to a separate repo under the organization to not bloat this repo and separate tests there. CodeCov is currently indicating a low percentage partly because of the many functions added under the plugin.

I will work on gettting `figure.py` to 100 percent coverage but the missing lines are not crucial right now to prevent review. As such I am opening this for review.

cvanelteren:
~~hmmm putting this back in draft as the thread safety fails on GHA. Fixed~~

cvanelteren:
I am putting a pin on this for a while. There is some development on xdist side that would maybe make this moot. Additionally, the `Configurator` and `_RcParams` should be unified providing a singular point of entry. I have gotten thread safety to work to some extent locally, but the ultraplot values are not properly updated on each thread, yielding a default matplotlib rc being thread safe.

PR #277, opened 2025-06-20
Revert xdist addition

cvanelteren:
This reverts:
-  500e45b159b029f20e0655c2feed49caa87b5171
- c939ae7a19c6f75f1f1a99ad1548ecf7a09f72de

And keeps these commits on #276. 

The reason for this revert is that I have suspicions that there is a racing condition on the writing and or reading on the RC params. I will need to properly isolate this dictionary to be able to add xdist reliable. In order to not halt other PRs or development this will go back to staging rather than keeping the error live.  Will continue to work on it, as it believe it is a minor thing that needs addressing for it to work properly but for now let's roll the dev changes back.


cvanelteren:
Pinging @beckermr just to have a second set of eyes on this to be sure that the reversion is going correctly.

beckermr (approved):

PR #281, opened 2025-06-21
Mv dynamic function to the subplotgrid

cvanelteren:
In order to support making RC params thread safe we need to ensure that our dynamically defined methods or classes are unified under __init__ and minimized where necessary. This is the firs in a sequence of refactors that moves some functions to be more appropriately defined which also aids in debugging and provides more clarity.

### DIFF
--- a/ultraplot/gridspec.py
+++ b/ultraplot/gridspec.py
@@ -18,6 +18,7 @@ from .config import rc
 from .internals import ic  # noqa: F401
 from .internals import _not_none, docstring, warnings
 from .utils import _fontsize_to_pt, units
+from .internals import warnings
 
 __all__ = ["GridSpec", "SubplotGrid", "SubplotsContainer"]  # deprecated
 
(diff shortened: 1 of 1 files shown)


beckermr (dismissed):
This pr changes the return type of the added commands. Is that intentional?

cvanelteren:
Return types are the same I believe.

cvanelteren:
Unless I am missing something -- what makes you think the return types are different? The `return` statement here is just whatever the wrapper returns. I used the wrapper otherwise the logic is the same everywhere.

cvanelteren:
Tests pass, will need to add some tests now. Not sure if I have the time today to do this. 
Will also add a comment on the return type as it may cause confusion to others.

cvanelteren:
Not that happy with the tests, but I am also no that keen on adding some of the methods for this PR as the focus is on replacing the dynamic content. So for now it's good enough. Open to review @beckermr

cvanelteren:
Will change the decorator to `apply_to_all`

ultraplot/gridspec.py
```diff
@@ -12,14 +12,17 @@
...
 
 from . import axes as paxes
 from .config import rc
 from .internals import ic  # noqa: F401
 from .internals import _not_none, docstring, warnings
 from .utils import _fontsize_to_pt, units
+from .internals import warnings
 
-__all__ = ["GridSpec", "SubplotGrid", "SubplotsContainer"]  # deprecated
+__all__ = ["GridSpec", "SubplotGrid"]  # deprecated
```
beckermr:
```suggestion
__all__ = ["GridSpec", "SubplotGrid"]
```

This comment can be removed.

ultraplot/gridspec.py
```diff
@@ -1528,38 +1543,6 @@ def __setitem__(self, key, value):
...
-            r"^(Returns\n-------\n)(.+)(\n\s+)(.+)",
-            r"\1SubplotGrid\2A grid of the resulting axes.",
-            doc,
-        )
-
-        # Apply the method
-        _grid_command.__qualname__ = f"SubplotGrid.{name}"
-        _grid_command.__name__ = name
-        _grid_command.__doc__ = doc
-        setattr(cls, name, _grid_command)
```
beckermr, line 1561: Because these lines are no longer executed, the new commands won't have correct doc strings, names, etc. This should be fixed.

ultraplot/tests/test_gridspec.py
```diff
@@ -0,0 +1,74 @@
...
+        axs.nonexistent()
+
+
+def test_docstring_injection():
+    """
+    @_apply_to_all should inject the docstring
+    """
+    fig, axs = uplt.subplots(nrows=1, ncols=2)
+    doc = axs.altx.__doc__
+    assert "Call `altx()` for every axes in the grid" in doc
```
beckermr: The doc string should match the original docstring?

beckermr (changes requested):
We need to fix the handling of the doc strings etc.

ultraplot/gridspec.py
cvanelteren: functool wraps should take care of this

ultraplot/tests/test_gridspec.py
cvanelteren: `axs` here is a `SubplotSpec`. The intent of the call is to assert t hat the decorator is correctly wrapping the docs. Unless you mean something else that I am missing (in case that is the case apologies ;-)!)

beckermr: I meant to say that the doc string for axs.altx should match the doc string for ax.altx. This applies to my main comment above as well.

cvanelteren: I'm not sure I fully agree here. `The SubplotGrid` is a distinct object, and the current docstring reflects its actual behavior: it applies `altx` to each `Axis` it contains. The method `altx` itself belongs to an `Axis`, not to the `GridSpec` or the `SubplotGrid` directly.

cvanelteren: I would however, agree if we provide a link to those methods

beckermr: ~Sure a link would be fine.~

Edit: See my request below. I really don't think it is a good idea to replace a helpful doc string with one that has very little information in it.

beckermr (changes requested):
Currently ultraplot does copy over the doc string

```python
IPython 8.31.0 -- An enhanced Interactive Python. Type '?' for help.

In [1]: import ultraplot as uplt

In [2]: fig, axs = uplt.subplots()

In [3]: axs.altx?
Signature: axs.altx(*args, **kwargs)
Docstring:
Add an axes locked to the same location with a
distinct x axis for every axes in the grid.
This is an alias and arguably more intuitive name for
`~ultraplot.axes.CartesianAxes.twiny`, which generates
two x axes with a shared ("twin") y axes.

Parameters
----------
**kwargs
    Passed to `~ultraplot.axes.CartesianAxes`. Supports all valid
    `~ultraplot.axes.CartesianAxes.format` keywords. You can optionally
    omit the x from keywords beginning with ``x`` -- for example
    ``ax.altx(lim=(0, 10))`` is equivalent to ``ax.altx(xlim=(0, 10))``.
    You can also change the default side for the axis spine, axis tick marks,
    axis tick labels, and/or axis labels by passing ``loc`` keywords. For example,
    ``ax.altx(loc='bottom')`` changes the default side from top to bottom.

Returns
-------
ultraplot.axes.CartesianAxes
    The resulting axes.

Note
----
This enforces the following default settings:

* Places the old x axis on the bottom and the new x
  axis on the top.
* Makes the old top spine invisible and the new bottom, left,
  and right spines invisible.
* Adjusts the x axis tick, tick label, and axis label positions
  according to the visible spine positions.
* Syncs the old and new y axis limits and scales, and makes the
  new y axis labels invisible.
File:      ~/mambaforge/envs/jax-galsim/lib/python3.10/site-packages/ultraplot/gridspec.py
Type:      method

```

Do we really want to change this? The new doc strings you wrote a very much not helpful.

cvanelteren:
I see, ok let's go with that.

cvanelteren:
Should be good now:

```
altx(*args, **kwargs) -> 'SubplotGrid' method of ultraplot.gridspec.S
ubplotGrid instance
    Add an axes locked to the same location with a
    distinct x axis for every axes in the grid.
    This is an alias and arguably more intuitive name for
    `~ultraplot.axes.CartesianAxes.twiny`, which generates
    two x axes with a shared ("twin") y axes.

[... etc]
```

Issue #288, opened 2025-06-26
Time series plot fails with non-standard calendars


pgf:
It seems UltraPlot doesn't handle non-standard calendars when dealing with DataArray time series.
xarray makes use of [cftime](https://unidata.github.io/cftime/) instead of ```numpy.datetime64``` for non-standard calendars (see xarray docs, [Time series data](https://docs.xarray.dev/en/stable/user-guide/time-series.html#)

```python
import numpy            as np
import xarray           as xr
import ultraplot        as uplt

# Time axis centered at mid-month
# Standard calendar
time1=xr.date_range("2000-01", periods=120, freq="MS")
time2=xr.date_range("2000-02", periods=120, freq="MS")
time=time1+0.5*(time2-time1)
# Non-standard calendar (uses cftime)
time1=xr.date_range("2000-01", periods=120, freq="MS", calendar="noleap")
time2=xr.date_range("2000-02", periods=120, freq="MS", calendar="noleap")
time_noleap=time1+0.5*(time2-time1)

da=xr.DataArray(data=np.sin(np.arange(0.,2*np.pi,np.pi/60.)),
               dims=("time",),
               coords={"time":time,},
               attrs={"long_name":"low freq signal", "units":"normalized"})

da_noleap=xr.DataArray(data=np.sin(2.*np.arange(0.,2*np.pi,np.pi/60.)),
               dims=("time",),
               coords={"time":time_noleap,},
               attrs={"long_name":"high freq signal", "units":"normalized"})

fig,axs = uplt.subplots(ncols=2)
axs.format(title=("Standard calendar", "Non-standard calendar"))
axs[0].plot(da)
axs[1].plot(da_noleap)

```
Result:
```python
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
Cell In[1], line 28
     26 axs.format(title=("Standard calendar", "Non-standard calendar"))
     27 axs[0].plot(da)
---> 28 axs[1].plot(da_noleap)

File [~/Library/local/miniforge3/envs/py3cf/lib/python3.13/site-packages/ultraplot/internals/inputs.py:365](http://localhost:8888/lab/tree/tmp/~/Library/local/miniforge3/envs/py3cf/lib/python3.13/site-packages/ultraplot/internals/inputs.py#line=364), in _preprocess_or_redirect.<locals>._decorator.<locals>._preprocess_or_redirect(self, *args, **kwargs)
    363     kwargs["color"] = color
[...]

PR #289, opened 2025-06-26
Add error when plotting cftime

cvanelteren:
Plotting on non-gregorian time scales raises an error. Matplotlib by default does not allow for this. Luckily there is a fix. By installing `nc-time-axis` some formatters are registered that allows for this. We can either implement this natively or add an install requirement

Currently I just implemented an error when one is plotting in a non-gregorian calendar (issue raised in #288).

### DIFF
--- a/ultraplot/internals/inputs.py
+++ b/ultraplot/internals/inputs.py
@@ -160,10 +160,9 @@ def _to_numpy_array(data, strip_units=False):
             result = np.atleast_1d(data.magnitude) * data.units
     else:
         result = np.atleast_1d(data)  # natively preserves masked arrays
-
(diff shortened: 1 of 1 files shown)


cvanelteren:
I am open for bundling the package directly. The source is just a simple formatters that are registered.

beckermr (changes requested):
The package seems well maintained, so let's depend on it. This PR needs a test.

cvanelteren:
Will add some integration. As @pgf pointed out in #288 we need to format the formatters to make them behave similarly to the default datetime for proper styling. Hopefully this is straightfoward.

cvanelteren:
Closing in favor of #344

Issue #292, opened 2025-06-28
Should LongitudeFormatter and LatitudeFormatter convert projected coordinates to dms by default?


gepcel:
Two examples first:

Example 1: notice the longitude labels of 2nd axis:

```
fig, axs = pplt.subplots(ncols=2,proj=['cyl', 'merc'],share=False)
axs.format(labels='all')
axs[0].format(lonlocator=(0.6, 113.25))
axs[1].format(lonlocator=(0.6, 113.25))
```

![Image](https://github.com/user-attachments/assets/5e42aeed-0833-4422-89d0-10f0cd2504a4)

Example 2: 
```
fig=plt.figure(figsize=(8,3))
ax0 = fig.add_subplot(1,2,1,projection=ccrs.PlateCarree())
ax0.gridlines(draw_labels=True,xlocs=(0.6, 113.25))
extent = ax0.get_extent(crs=ccrs.PlateCarree())
ax1 = fig.add_subplot(1,2,2, projection=ccrs.Mercator())
ax1.gridlines(draw_labels=True,xlocs=(0.6, 113.25))
```

![Image](https://github.com/user-attachments/assets/3fd9c4e7-53ff-4fa4-b1fd-bd3e43dd0d18)

Seems that cartopy converts coordinates of mercator to platecarree by default (like 66791.6944759644 to 0.6°E), but ultraplot does not.
And also, the ax.set_xitcks() method of cartopy also has a `crs` parameter ([doc](https://scitools.org.uk/cartopy/docs/latest/gallery/gridlines_and_labels/tick_labels.html#sphx-glr-gallery-gridlines-and-labels-tick-labels-py)).
But `lonlocator_kw` of ultraplot doesn't seem to accept `crs`, am I missing something?

cvanelteren:
Could be an issue introduced by the sharing; Proplot had no implementation for the sharing. Can put it on the docket.

cvanelteren:
Can confirm that dms is not properly handled after sharing is added. Need to figure out where this goes wrong exactly

cvanelteren:
Found the issue (line 1710)

cvanelteren:
![Image](https://github.com/user-attachments/assets/90c0931b-cb5f-4c8e-9b3a-b4efa13a98fd)

```python
import ultraplot as uplt, numpy as np, pandas as pd
l = (0.6, 113.25)
fig, ax = uplt.subplots(ncols=2, proj=("cyl", "merc"), share=0)
[...]

PR #293, opened 2025-06-28
Fix DMS not set on some projections

cvanelteren:
Addresses #292 where DMS is not set for some axes that should have it set.

### DIFF
--- a/ultraplot/ticker.py
+++ b/ultraplot/ticker.py
@@ -822,7 +822,7 @@ class FracFormatter(mticker.Formatter):
         return string
 
 
-class _CartopyFormatter(object):
+class _CartopyFormatter(_PlateCarreeFormatter):
     """
     Mixin class for cartopy formatters.
     """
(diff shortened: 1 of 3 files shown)


cvanelteren:
Some baselines will probable fail (but maybe not). Will add a test.

beckermr (changes requested):
This change confuses me. There were no cases where the code that's been removed was needed?

cvanelteren:
Actually you are right, I thought I added this but that's not true will do some more searching why it is not working.

cvanelteren:
It's a bit strange when I edit line 244 and remove the merc projector it is set to false and it works as it should for the mercator projection but it does not make sense since use_dms will be set to false.

cvanelteren:
I will wait before the issuer gives some clarity. The PR now only contains a test.

ultraplot/ticker.py
```diff
@@ -884,18 +884,6 @@ def __init__(self, lon0=0, *args, **kwargs):
...
-        Format the longitude, accounting for lon0 offset.
-        """
-        # Adjust longitude value based on lon0
-        adjusted_lon = x - self.lon0
-        # Normalize to -180 to 180 range
-        adjusted_lon = ((adjusted_lon + 180) % 360) - 180
-        print(x)
-        # Use the original formatter with the adjusted longitude
-        return super().__call__(adjusted_lon, pos)
-
```
beckermr, line 898: Why was this method removed?

cvanelteren: I added this as the lon0 was not moving correctly, but I think the issue was in the formatter not being overriden with the @poperty stuff -- potentially a change in the backend. I can confirm that it is shifting properly as the test is passing without this block, but not with this block. Also confirmed visually.

Issue #294, opened 2025-06-28
labelrotation in ax.colorbar() did not work


ybmy001:
Below is my code, as mentioned above.

```
import ultraplot as uplt

cmap = uplt.Colormap("plasma_r")
fig, ax = uplt.subplots()

for i, loc in enumerate(['b', 'l', 'r', 't']):
    ax.colorbar(cmap, loc=loc, label="label", labelloc="right", labelrotation=90 * i)

uplt.show(block=1)
```

ybmy001:
When colorbar location sets in the axis, e.g. ['ur', 'ul', 'll', 'lr'], add parameter `labelloc` raise error. ValueError: Location not understood.

```
import ultraplot as uplt

cmap = uplt.Colormap("plasma_r")
fig, ax = uplt.subplots()

for i, loc in enumerate(['ur', 'ul', 'll', 'lr']):
    # ax.colorbar(cmap, loc=loc, label="label")  # run OK
    ax.colorbar(cmap, loc=loc, label="label", labelloc="right")  # ValueError: Location not understood.

uplt.show(block=1)
```

cvanelteren:
You are absolutely right. I noticed this myself last week. Need to fix this.

cvanelteren:
Fix in #295.

ybmy001:
> You are absolutely right. I noticed this myself last week. Need to fix this.
Suddenly, I thought of another small bug, `cmap='rainbow_r'` not works in ultralplot, I found this bug for a long time in proplot.
below is my code, 

```
import ultraplot as uplt

fig, ax = uplt.subplots()

cmaps = ['jet', 'jet_r', 'rainbow', 'rainbow_r']
for i, loc in enumerate(['r', 'l', 't', 'b']):
    ax.colorbar(uplt.Colormap(cmaps[i]), loc=loc, label=cmaps[i])

uplt.show(block=1)
```

ybmy001:
`cmap='rainbow_r'` not works, another example. the result of `cmap='rainbow_r'` is same to `cmap='rainbow'`.

```
import ultraplot as uplt
import numpy as np

data = np.arange(9).reshape((3, 3))

fig, ax = uplt.subplots()
img = ax.pcolormesh(data, cmap='rainbow_r')
ax.colorbar(img)
uplt.show(block=1)
```

cvanelteren:
Thanks found a fix for it.

cvanelteren:
[...]

PR #295, opened 2025-06-28
Hotfix add all locations to colorbar label

cvanelteren:
Addresses #294

### DIFF
--- a/ultraplot/tests/test_colorbar.py
+++ b/ultraplot/tests/test_colorbar.py
@@ -454,13 +454,11 @@ def test_colorbar_invalid_fill_label_placement(invalid_labelloc):
 
 
 @pytest.mark.parametrize("unknown_loc", ["unknown", "custom", "weird_location", 123])
-def test_colorbar_label_placement_default(unknown_loc):
-    # Test the default case (final else statement)
-    # Unknown/unrecognized locations should default to obj.set_label(label)
+def test_colorbar_wrong_label_placement_should_raise_error(unknown_loc):
+    # Unknown locs should raise errors
     cmap = uplt.Colormap("plasma_r")
     title = "Test Label"
     fig, ax = uplt.subplots()
-    # This should not raise an error and should use the default behavior
     with pytest.raises(KeyError):
         cbar = ax.colorbar(cmap, loc=unknown_loc, label=title)
 
--- a/ultraplot/ticker.py
+++ b/ultraplot/ticker.py
@@ -892,6 +892,7 @@ class LongitudeFormatter(_CartopyFormatter, LongitudeFormatter):
         adjusted_lon = x - self.lon0
         # Normalize to -180 to 180 range
         adjusted_lon = ((adjusted_lon + 180) % 360) - 180
+        print(x)
         # Use the original formatter with the adjusted longitude
         return super().__call__(adjusted_lon, pos)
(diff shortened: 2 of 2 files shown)


ultraplot/tests/test_colorbar.py
```diff
@@ -353,3 +354,135 @@ def test_label_placement_fig_colorbar2():
...
+@pytest.mark.parametrize("unknown_loc", ["unknown", "custom", "weird_location", 123])
+def test_colorbar_label_placement_default(unknown_loc):
+    # Test the default case (final else statement)
+    # Unknown/unrecognized locations should default to obj.set_label(label)
+    cmap = uplt.Colormap("plasma_r")
+    title = "Test Label"
+    fig, ax = uplt.subplots()
+    # This should not raise an error and should use the default behavior
+    with pytest.raises(KeyError):
+        cbar = ax.colorbar(cmap, loc=unknown_loc, label=title)
```
beckermr: The comment here doesn't match the test result which only passes if an error is raised.

ultraplot/tests/test_colorbar.py
```diff
@@ -353,3 +354,135 @@ def test_label_placement_fig_colorbar2():
...
+    cbar = ax.colorbar(cmap, loc=cbarloc, labelloc=labelloc, title=title)
+
+    x_label = cbar.ax.xaxis.label.get_text()
+    y_label = cbar.ax.yaxis.label.get_text()
+
+    assert title in (x_label, y_label), (
+        f"Expected label '{title}' not found. "
+        f"xaxis label: '{x_label}', yaxis label: '{y_label}', "
+        f"labelloc='{labelloc}', cbarloc='{cbarloc}'"
+    )
```
beckermr, line 392: This assert needs to be specialized to the specific test case, only testing one of x or y.

ultraplot/tests/test_colorbar.py
```diff
@@ -353,3 +354,135 @@ def test_label_placement_fig_colorbar2():
...
+        or (
+            hasattr(cbar.ax.xaxis.label, "get_text")
+            and cbar.ax.xaxis.label.get_text() == title
+        )
+        or (
+            hasattr(cbar.ax.yaxis.label, "get_text")
+            and cbar.ax.yaxis.label.get_text() == title
+        )
+    )
+    assert label_found, f"Label not found for loc='{loc}' with labelloc=None"
```
beckermr, line 486: Again this assert needs to be specialized to the specific test case looking at either x or y or the title.

beckermr (changes requested):
The tests need work.

ultraplot/tests/test_colorbar.py
cvanelteren: You are correct that the locations may differ, but in the end we are only attaching it to x or y depending on the location. Checking for if the label exists would imply that the label is set correctly otherwise an error would be raised.

ultraplot/tests/test_colorbar.py
cvanelteren: Same as comment above

ultraplot/tests/test_colorbar.py
cvanelteren: Marked the test wrong. Label is corrected.

PR #296, opened 2025-06-29
fix late binding and proper reversal for funcs

cvanelteren:
A secondary issue came up in #294 where a callable is not properly inverse. 

This PR fixes a bug in the reversal of colormap segment data where callable segments were incorrectly captured due to Python’s late binding behavior in closures. Specifically, lambdas inside the dictionary comprehension referenced the loop variable data, which caused all callables to use the last data value. This is resolved by binding data as a default argument in the lambda. Explicit segment lists are unaffected and still reversed correctly.

### DIFF
--- a/ultraplot/colors.py
+++ b/ultraplot/colors.py
@@ -1271,14 +1271,17 @@ class ContinuousColormap(mcolors.LinearSegmentedColormap, _Colormap):
         --------
         matplotlib.colors.LinearSegmentedColormap.reversed
         """
-        # Reverse segments
-        segmentdata = {}
-        for key, data in self._segmentdata.items():
-            if callable(data):
-                segmentdata[key] = lambda x, func=data: func(1.0 - x)
-            else:
-                segmentdata[key] = tuple((1.0 - x, y1, y0) for x, y0, y1 in data)
 
+        # Reverse segments
+        def _reverse_data(data):
+            if callable(data):
+                return lambda x, func=data: func(1 - x)
+            else:
+                return [(1.0 - x, y1, y0) for x, y0, y1 in reversed(data)]
+
+        segmentdata = {
+            key: _reverse_data(data) for key, data in self._segmentdata.items()
+        }
         # Reverse gammas
         if name is None:
             name = self._make_name(suffix="r")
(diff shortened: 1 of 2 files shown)


cvanelteren:
image


```python
import ultraplot as uplt

fig, ax = uplt.subplots()

cmaps = ["jet", "jet_r", "rainbow", "rainbow_r"]
for i, loc in enumerate(["r", "l", "t", "b"]):
    ax.colorbar(uplt.Colormap(cmaps[i]), loc=loc, label=cmaps[i])
uplt.show(block=1)
```

ultraplot/colors.py
```diff
@@ -3156,7 +3157,7 @@ def _translate_key(self, original_key, mirror=True):
 
             # Try mirroring the non-lowered key
             if reverse:
-                original_key = original_key.strip("_r")
+                original_key = original_key.rstrip("_r")
```
beckermr: This method removes any of the characters in `_r`, not `_r` if the string ends with `_r`. I think you may want something like `original_key.rsplit("_r", maxsplit=1)[0]`

beckermr (changes requested):

ultraplot/colors.py
cvanelteren: You are correct, but then we should replace all strip functions by a split you suggest. My edit here merely replaces it with a more appropriate function. I think the original strip is wrong as it also deletes leading "_r".

beckermr: Right we should do that fix. We'll need another PR

cvanelteren: I can also include it in here, although I think this issue will not be a major things as the `_r` is added by ultraplot. If the colormap ends with `_r` it will be listed as `{cmapnameendingwith_r}_r` which should then split fine I think.

cvanelteren:
We can go ahead and merge this I guess @beckermr?

beckermr:
Didn't you want to fix the string _r thing here too?

cvanelteren:
Ah ok  that wasn't entirely clear. Will do that now

cvanelteren:
done ;-)!

beckermr (approved):

Issue #300, opened 2025-07-02
Feature Request: Add support for colorbars


nickbsb:
Please consider adding support for vertical orientation of inset colorbars, as it is useful for geographic and remote sensing visualizations.Thanks.

![Image](https://github.com/user-attachments/assets/4a0c6ad4-48d3-47cc-a470-922fdbad681b)

cvanelteren:
Sounds like a good feature. I am rather busy at the moment with other issues, but we welcome PRs. Have a look at our contributing guide or #298

cvanelteren:
Had some time, implemented in #301

cvanelteren:
Added on main.

PR #301, opened 2025-07-02
Feat vert inset cbars

cvanelteren:
This PR would address #300 

UltraPlot guesses the size of the colorbar as a pre-render. I would be better to move this all to dynamic as it can drastically differ if one runs this in say a jupyter notebook or a gtk window. This PR only adds logic to allow for this same pre-render to have vertical colorbars while still having the same defaults for the horizontal.

### DIFF
--- a/ultraplot/axes/base.py
+++ b/ultraplot/axes/base.py
@@ -2150,22 +2150,68 @@ class Axes(maxes.Axes):
             labspace += 1.2 * fontsize / 72
         labspace /= self._get_size_inches()[1]  # space for labels
 
+        # Determine colorbar dimensions and frame sizes based on orientation
+        # Bounds are x0, y0, width, height in axes-relative coordinates
+        if orientation == "horizontal":
+            # For horizontal: length is width, width is height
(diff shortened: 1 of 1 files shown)


cvanelteren:
~~NOTE: We may need to adjust the docs~~ not necessary, the tutorial is not stating anything about restriction on colorbars.

ultraplot/axes/base.py
```diff
@@ -2144,28 +2148,97 @@ def _parse_colorbar_inset(
...
+        if orientation == "horizontal":
+            labspace /= self._get_size_inches()[1]
         else:
-            labspace += 1.2 * fontsize / 72
-        labspace /= self._get_size_inches()[1]  # space for labels
+            labspace /= self._get_size_inches()[0]
+
+        # Bounds are x0, y0, width, height in axes-relative coordinates
+        # Location in axes-relative coordinates
+        # Bounds are x0, y0, width, height in axes-relative coordinates
```
beckermr:
```suggestion
```

PR #304, opened 2025-07-03
adjust value slighlty

cvanelteren:
minor adjustment of #301

### DIFF
--- a/ultraplot/axes/base.py
+++ b/ultraplot/axes/base.py
@@ -2150,7 +2150,7 @@ class Axes(maxes.Axes):
         fontsize = _fontsize_to_pt(fontsize)
         scale = 1.2
         if orientation == "vertical":
-            scale = 1.8  # we need a little more room
+            scale *= 2.1  # we need a little more room
         if label is not None:
             labspace += 2 * scale * fontsize / 72
         else:


beckermr (approved):
Do we want to add a test for this by adding an image to our image comparison tests?

cvanelteren:
Yes. Hold on. It's a bit tricky as the vert bars suffer from looking bad if you open the figure in gtk but look good on save

cvanelteren:
Was drafting a release and noticed this

cvanelteren:
Need to fix the framing for the inset

cvanelteren:
Ok I sat down and went through all the edge cases but I think I got it looking relatively nice now. I am afraid, however, that the tests will indicate a visual difference since the logic is now concerning all the locations for insets.

cvanelteren:
The resulting tests look good but differ in shape. Not sure why but likely related to the newly added refactors. Otherwise visually looks good.

cvanelteren:
See above for changes @beckermr

cvanelteren:
Can we go ahead an merge this @beckermr? I manually checked the errors and there is an image dimension error but visually they look similar.

beckermr:
yes merge away!

Issue #309, opened 2025-07-19
Two legends have appeared


lwq-star:
Using this code produced two legends, likely due to a conflict between ultraplot and seaborn.
```python
import ultraplot as uplt
import seaborn as sns

fig, axs = uplt.subplots(journal='nat2')

df = sns.load_dataset("titanic")
sns.violinplot(data=df, x="class", y="age", hue="alive")

axs.legend(
    loc='lower right',
    frameon=False,  
    title=None,
)
```

<img width="721" height="757" alt="Image" src="https://github.com/user-attachments/assets/2075576c-f113-45f4-a5c0-a83a5fdafa76" />

Everything works normally when ultraplot is not used.
```
import seaborn as sns
import matplotlib.pyplot as plt

df = sns.load_dataset("titanic")
sns.violinplot(data=df, x="class", y="age", hue="alive")
plt.legend(
    loc='lower right',
    frameon=False, 
    title=None, 
)
```

<img width="580" height="436" alt="Image" src="https://github.com/user-attachments/assets/e7424c01-647c-420d-b467-7a7a9d4756fc" />

cvanelteren:
I know what's happening here but cannot fix it until the end of this month. The issue is that seaborn adds the legend as well. I believe we can queue the legend to prevent the double from showing up.

cvanelteren:
The issue is caused by seaborn adding a legend as well.  I think the issue should be resolved within the violinplot functions that we provide but it produces an error that needs fixing. For now you can set `legend=False` in the violinplot call for seaborn.

cvanelteren:
The package currently does not support the nested violinplot like seaborn does. It should be possible by using some logic from the bar charts in the violinplot. PR is welcome.

lwq-star:
I discovered that the seaborn library has a` move_legend` function. The documentation for this function states: "Matplotlib legends do not expose public control over their position parameters. So this function creates a new legend, copying over the data from the original object, which is then removed." 
However, while using it, I encountered this error: `NotImplementedError: cannot remove artist.` I suspect this is caused by ultraplot. Can ultraplot achieve similar functionality?

```
import ultraplot as uplt
import seaborn as sns

fig, axs = uplt.subplots(journal='nat2')

df = sns.load_dataset("titanic")
ax = sns.violinplot(data=df, x="class", y="age", hue="alive")

# axs.legend(
#     loc='lower right',
#     frameon=False,  
#     title=None,
# )

[...]

PR #310, opened 2025-08-01
Sync `_legend_dict` on legend location change

cvanelteren:
An issue popped up in #309  where if locations are changed for legends they are not synced with `_legend_dict`. This PR fixes this by override the `set_loc` function of legend. 

Why was this an issue? UltraPlot allows for updating the styling of legends when a legend already exists on at that location. Since the internal reference is used, a sync issue appears if for whatever reason the legend is moved.


cvanelteren:
needs tests

beckermr (approved):

Issue #311, opened 2025-08-03
`alpha` cause facecolor being ignored in `scatter`


gepcel:
During using `scatter`, when `alpha` param is set, `facecolor='none'` is ignored. eg:

```python
fig,ax=pplt.subplot()
ax.scatter(0.4, 0.5, 100, fc='none', ec='k', alpha=.2)
ax.scatter(0.5, 0.5, 100, fc="none", ec="k")
ax.scatter(0.6, 0.5, 100, fc="none", ec="k", alpha=1)
fig.save('t2.png')
```

All 3 dots should have `none` facecolor` and black ecgecolor with difference alpha. But I got:

<img width="2959" height="2846" alt="Image" src="https://github.com/user-attachments/assets/98691560-4350-41cd-8094-c7e63f217c46" />

cvanelteren:
That is indeed a bug. Kinda weird that for 'white' it works like it is transparent:

<img width="651" height="658" alt="Image" src="https://github.com/user-attachments/assets/12a8b3fb-6a2c-4bd3-b1d0-ce5a85cc4a70" />

cvanelteren:
It's cut off but i set the alpha for the first one to 1 and it acts as if transparent which is odd.

cvanelteren:
Fixed in #312

cvanelteren:
Fixed on main. Free free to close the issue if it resolved your problem.

gepcel:
> That is indeed a bug. Kinda weird that for 'white' it works like it is transparent:
> 

From this image, Do you mind me asking what kind of editor you are using?

cvanelteren:
Zed!

gepcel:
I'm reopening this issue because I think it is related. Sometimes edgecolor is ignored. Let's see an example:

```
d = pd.DataFrame(dict(x=[1, 2], y=[1, 2], sizes=[300, 300]))
fig, ax = pplt.subplots(ncols=3)
ax[0].scatter("x", "y", s="sizes", data=d, fc="red8", ec="none", alpha=1)
ax[1].scatter("x", "y", s="sizes", data=d.loc[1], fc="red8", ec="none", alpha=1)
ax[2].scatter("x", "y", s="sizes", data=d.loc[1], fc="red8", ec="none")
ax.format(
    xlim=(0.8, 2.2),
    ylim=(0.9, 2.1),
    suptitle="ax[0]: Right ec with len(d)>1 and alpha"
[...]

PR #312, opened 2025-08-03
fix color being parsed for none

cvanelteren:
Addresses an issue where 'none' is parsed as a color (see #311)


cvanelteren:
Merging as the failed test is exactly one the part that is fixed.

Issue #313, opened 2025-08-06
Colormap not using vcenter for pcolormesh/pcolor etc.


beckermr:
ultraplot appears to be ignoring `vmin/vmax`. Further, for some ranges of data inputs, the output plot doesn't use the full range of the color map.

```python
import ultraplot as uplt
import numpy as np

rng = np.random.default_rng(seed=10)
mvals = rng.normal(size=(32, 32))

fig, axs = uplt.subplots(nrows=1, ncols=3, share=0)
for i, ax in enumerate(axs):
    if i == 1 or i == 2:
        vmin = -0.2
        vmax = 2.0
        if i == 2:
            mvals = np.clip(mvals, vmin, vmax)
        cmap = "spectral"
    else:
        cmap = "spectral"
        vmin = None
        vmax = None


    m = ax.pcolormesh(
        mvals,
        cmap=cmap,
        discrete=False,
        vmin=vmin,
        vmax=vmax,
    )
    ax.format(grid=False, xticklabels=[], xticks=[], yticklabels=[], yticks=[])
    ax.colorbar(
        m,
        loc="r",
        label=f"{i}"
    )
```

<img width="2014" height="534" alt="Image" src="https://github.com/user-attachments/assets/4c8eea3d-2a4d-4798-a684-a647774fa2ad" />

beckermr:
cc @cvanelteren 

IDK how we broke this, but it seems like a pretty fundamental issue.

beckermr:
The matplotlib code works as expected:

```python
import matplotlib.pyplot as plt
import numpy as np

rng = np.random.default_rng(seed=10)
mvals = rng.normal(size=(32, 32))

fig, axs = plt.subplots(nrows=1, ncols=3, figsize=(9, 3))
for i, ax in enumerate(axs.ravel()):
    if i == 1 or i == 2:
        vmin = -0.2
        vmax = 2.0
        if i == 2:
            mvals = np.clip(mvals, vmin, vmax)
        cmap = "spectral"
    else:
        cmap = "spectral"
        vmin = None
        vmax = None


    m = ax.pcolormesh(
        mvals,
        cmap=cmap,
        vmin=vmin,
        vmax=vmax,
    )
    fig.colorbar(
        m,
        ax=ax,
        label=f"{i}"
    )
```

<img width="1800" height="600" alt="Image" src="https://github.com/user-attachments/assets/71afaefd-ce1b-43f2-bb3b-c3b93fd591a0" />

[...]

PR #314, opened 2025-08-07
Fix edge case where vcenter is not properly set for diverging norms

cvanelteren:
Addresses #313 

Not sure if this is the best way to approach this. The issue in #313 is caused by the vcenter being 0 where the data is clearly not zerod around that value. I would prefer to push this code in the parsing of cmap directly inferring it from the input data when set, but the current colorbar handling is still rather opague to me. I wanted to move it to a separate class and clean it up but there seems that task proven to be more challenging than I initially thought. Open for feedback to fix this is another way.

Basically, the vcenter is now inferred if vmin and vmax is set, otherwise defaults to normal default vcenter = 0.

### DIFF
--- a/ultraplot/tests/test_color.py
+++ b/ultraplot/tests/test_color.py
@@ -1,4 +1,4 @@
-import ultraplot as uplt, pytest
+import ultraplot as uplt, numpy as np, pytest
 
 
 @pytest.mark.mpl_image_compare


cvanelteren:
Note that vcenter is the data value for which the space is exactly half. The PR currently infers it only if `fair` is set which treats each half as linearly similar. For non-linearly spaced spaces, vcenter may not be the mere average.

cvanelteren:
Actually I move it to `_parse_cmap` cuz i don't like how it is handled here. Sorry for the ping @beckermr

beckermr (approved):

beckermr (approved):

PR #319, opened 2025-08-13
Restore redirection for tripcolor for GeoPlotting

pgf:
Restore redirection for tripcolor for GeoPlotting that was left out of #222 (see #221 ).

PR #320, opened 2025-08-14
Fix issue where double decorator does not parse function name

cvanelteren:
Context is in #319


cvanelteren:
Needs a test

cvanelteren:
Not sure why the test if failing on a non-related part.

pgf:
If this fix the issue, the `_preprocess_or_redirect` decorator should be redundant also in the other tri* methods, tricontour(f), right?

cvanelteren:
Yes!

Issue #270, opened 2025-06-19
Plot Rendering Issue in Positron IDE with Ultraplot


gepcel:
First, I'm unsure if this is specifically an Ultraplot issue or if it's related to using Positron IDE with Python. Feel free to disregard if this isn't relevant to others. However, it's frustrating to have to switch IDEs whenever I need to visualize data during analysis.

**Issue Description:**
When using Ultraplot with Positron IDE ([main website](https://positron.posit.co/), [github](https://github.com/posit-dev/positron)), plots fail to render in the plot panel. For example:

```python
import ultraplot as pplt
fig, ax = pplt.subplots()
pplt.show()
```

**Error Message:**

```
Error rendering plot to 'Auto' size: Internal error: Method locally_modified_subplot_params() 
is disabled on ultraplot gridspecs. (-32603)
```

**Additional Context:**

* There was a period when this issue disappeared, but I'm uncertain whether this was due to package/IDE version changes or configuration adjustments (e.g., matplotlib backends).
* Current versions:
    * Ultraplot: 1.57
    * Positron IDE: 2025.07.0

cvanelteren:
I can see in our code that we have this block

```python
    # Delete attributes. Don't like having special setters and getters for some
    # settings and not others. Width and height ratios can be updated with update().
    # Also delete obsolete 'subplotpars' and built-in tight layout function.
    tight_layout = _disable_method("tight_layout")  # instead use custom tight layout
    subgridspec = _disable_method("subgridspec")  # instead use variable spaces
    get_width_ratios = _disable_method("get_width_ratios")
    get_height_ratios = _disable_method("get_height_ratios")
    set_width_ratios = _disable_method("set_width_ratios")
    set_height_ratios = _disable_method("set_height_ratios")
    get_subplot_params = _disable_method("get_subplot_params")
    locally_modified_subplot_params = _disable_method("locally_modified_subplot_params")
```

cvanelteren:
So the issue is definitely generated on our end but not the question remains why these methods were disabled by Luke 🤔 
The docs from matplotlib state:


> locally_modified_subplot_params()
[[source]](https://github.com/matplotlib/matplotlib/blob/v3.10.3/lib/matplotlib/gridspec.py#L420-L427)
    Return a list of the names of the subplot parameters explicitly set in the GridSpec.
[...]

PR #321, opened 2025-08-16
Enable gridspec methods

cvanelteren:
Enable some methods that are disable to push the updates through `update`. This issues popped up in #270 where some editors require the use of these methods.


cvanelteren:
Running this through GHA to test the waters, as I am not sure if these methods can cause other issues. Will need to evaluate if this will cause any issues -- probably not but still.

cvanelteren:
To fix the issue in #270,  we can remove the `RunTimeError` and raise a warning instead. To properly fix it we cannot remove the disablers as the funneling through `update` uses custom properties and would collide potentially with user settings.

cvanelteren:
I would be more included to override the methods when necessary rather than disabling them. Will look at this later.

cvanelteren:
Not planned for now.

PR #322, opened 2025-08-16
Update Cartopy references

rcomer:
The Cartopy docs have a new home on ReadTheDocs.  There is an intention to retire the whole www.scitools.org.uk site at some point so the version-specific links will also break.


cvanelteren:
Thanks

cvanelteren:
error is not related to this pr will merge this when I have access to my machine again.

PR #323, opened 2025-08-16
Fix inhomogeneous violin test

rcomer:
I think the current failure on main is because one of the violin plots uses `np.random` without a fixed seed.


cvanelteren:
Thanks -- missed it!

PR #336, opened 2025-08-27
Fix issue where fair was set to default for diverging colormaps

cvanelteren:
Diverging colormaps are by default set to fair. This PR introduces logic that fair is ignored if vmin and or vmax is given. 

Addresses #313

### DIFF
--- a/ultraplot/axes/plot.py
+++ b/ultraplot/axes/plot.py
@@ -2709,7 +2709,8 @@ class PlotAxes(base.Axes):
         norm = _not_none(norm, "div" if "diverging" in trues else "linear")
         # If using a diverging norm, fair=True, and vcenter not set, default to midpoint
         if norm in ("div", "diverging") or "diverging" in trues:
-            fair = norm_kw.get("fair", True)  # defaults to True
+            is_fair = True if vmin is None and vmax is None else False
+            fair = norm_kw.get("fair", is_fair)  # defaults to True
             vcenter = norm_kw.get("vcenter", 0)
             if fair and vcenter is None and vmin is not None and vmax is not None:
                 vcenter = 0.5 * (vmin + vmax)
--- a/ultraplot/axes/plot.py
+++ b/ultraplot/axes/plot.py
@@ -2709,13 +2709,13 @@ class PlotAxes(base.Axes):
         norm = _not_none(norm, "div" if "diverging" in trues else "linear")
         # If using a diverging norm, fair=True, and vcenter not set, default to midpoint
         if norm in ("div", "diverging") or "diverging" in trues:
-            # When vmin or vmax is given we cannot be sure the colormap is fair
-            is_fair = True if vmin is None and vmax is None else False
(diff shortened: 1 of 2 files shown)


ultraplot/axes/plot.py
```diff
@@ -2709,10 +2709,13 @@ def _parse_cmap(
...
         # If using a diverging norm, fair=True, and vcenter not set, default to midpoint
         if norm in ("div", "diverging") or "diverging" in trues:
-            fair = norm_kw.get("fair", True)  # defaults to True
+            # When vmin or vmax is given we cannot be sure the colormap is fair
+            is_fair = vmin is None and vmax is None
+            fair = norm_kw.get("fair", is_fair)
             vcenter = norm_kw.get("vcenter", 0)
-            if fair and vcenter is None and vmin is not None and vmax is not None:
-                vcenter = 0.5 * (vmin + vmax)
+            if fair == False:
```
beckermr:
```suggestion
            if not fair:
```

ultraplot/axes/plot.py
```diff
@@ -2709,10 +2709,13 @@ def _parse_cmap(
         norm = _not_none(norm, "div" if "diverging" in trues else "linear")
         # If using a diverging norm, fair=True, and vcenter not set, default to midpoint
         if norm in ("div", "diverging") or "diverging" in trues:
-            fair = norm_kw.get("fair", True)  # defaults to True
+            # When vmin or vmax is given we cannot be sure the colormap is fair
+            is_fair = vmin is None and vmax is None
+            fair = norm_kw.get("fair", is_fair)
```
beckermr: This bit will still override fair if it is set in norm_kw regardless of the value of vmin/vmax. That seems wrong.

ultraplot/axes/plot.py
```diff
@@ -2713,7 +2713,7 @@ def _parse_cmap(
             is_fair = vmin is None and vmax is None
             fair = norm_kw.get("fair", is_fair)
             vcenter = norm_kw.get("vcenter", 0)
-            if fair == False:
+            if fair:
```
beckermr: This swapped the logic.

beckermr: Now vcenter is not set since the logic was swapped. That looks wrong too?

cvanelteren:
@beckermr  my bad I got confused for a little bit. This should be correct I think

cvanelteren:
@beckermr  The test is there as `test_vcenter_values`

cvanelteren:
Tests look more diverging for the plots that fail.

beckermr:
Yep I actually think we need to close this PR and back out the previous one. This is the intended behavior and I was using the code wrong.

cvanelteren:
The source is assuming with #313  that the vcenter is 0, we go proceed with this then we can back this out I guess? Should we keep the test in with `diverging = False` or leave it out all together?

beckermr:
vcenter defaults to zero in the normalizers, so this PR is not needed.

PR #341, opened 2025-09-06
Fix some links for docs

cvanelteren:
As it says on the tin.


cvanelteren:
Confirm on build that `py:module:` is recognized (locally it is not) but `sphobjinv` does see it like that.

PR #343, opened 2025-09-08
Lazy loading colormaps

cvanelteren:
This PR changes the colormap loading to lazy loading -- effectively reducing the import time from to 0.38s cmaps: 0.00082s.

### DIFF
--- a/ultraplot/colors.py
+++ b/ultraplot/colors.py
@@ -3091,12 +3091,9 @@ class ColormapDatabase(mcm.ColormapRegistry):
             The source dictionary.
         """
         super().__init__(kwargs)
-        # The colormap is initialized with all the base colormaps
-        # We have to change the classes internally to Perceptual, Continuous or Discrete
-        # such that ultraplot knows what these objects are. We piggy back on the registering mechanism
-        # by overriding matplotlib's behavior
-        for name in tuple(self._cmaps.keys()):
-            self.register(self._cmaps[name], name=name)
+        # The colormap is initialized with all the base colormaps.
+        # These are converted to ultraplot's own colormap objects
+        # on the fly when they are first accessed.
 
     def _translate_deprecated(self, key):
         """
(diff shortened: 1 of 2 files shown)


ultraplot/colors.py
```diff
@@ -3188,9 +3185,48 @@ def __getitem__(self, key):
...
             key = key.removesuffix("_r")
+
         # Retrieve colormap
         if self._has_item(key):
-            value = self._cmaps[key].copy()
+            value = self._cmaps[key]
+
+            # Lazy loading from file
+            if isinstance(value, dict) and value.get("is_lazy"):
+                from . import colors as pcolors
```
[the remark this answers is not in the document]
cvanelteren: Thanks bot!

cvanelteren:
hmm this test that is failing is passing locally.. strange

ultraplot/colors.py
```diff
@@ -3090,13 +3090,10 @@ def __init__(self, kwargs):
...
             The source dictionary.
         """
-        super().__init__(kwargs)
-        # The colormap is initialized with all the base colormaps
-        # We have to change the classes internally to Perceptual, Continuous or Discrete
-        # such that ultraplot knows what these objects are. We piggy back on the registering mechanism
-        # by overriding matplotlib's behavior
-        for name in tuple(self._cmaps.keys()):
-            self.register(self._cmaps[name], name=name)
+        super().__init__({k.lower(): v for k, v in kwargs.items()})
```
beckermr, line 3098: Why switch to `k.lower()` here?

ultraplot/colors.py
```diff
@@ -3090,13 +3090,10 @@ def __init__(self, kwargs):
         kwargs : dict-like
             The source dictionary.
         """
-        super().__init__(kwargs)
-        # The colormap is initialized with all the base colormaps
-        # We have to change the classes internally to Perceptual, Continuous or Discrete
-        # such that ultraplot knows what these objects are. We piggy back on the registering mechanism
-        # by overriding matplotlib's behavior
-        for name in tuple(self._cmaps.keys()):
-            self.register(self._cmaps[name], name=name)
```
beckermr, line 3099: I guess we are no longer using the `register` method?

ultraplot/colors.py
```diff
@@ -3188,9 +3185,42 @@ def __getitem__(self, key):
...
+
+                if cmap:
+                    if is_default and cmap.name.lower() in CMAPS_CYCLIC:
+                        cmap.set_cyclic(True)
+                    self._cmaps[key] = cmap
+                    value = cmap
+                else:  # failed to load
+                    # remove from registry to avoid trying again
+                    del self._cmaps[key]
+                    raise KeyError(f"Failed to load colormap {key!r} from {path!r}")
```
beckermr: The previous code warned on failures at least according to the keyword. This code appears to raise an error?

ultraplot/colors.py
cvanelteren: We perform a lower command internally later when the colormap is registered.

ultraplot/colors.py
cvanelteren: I will change it back to a warn.

ultraplot/colors.py
cvanelteren: Ah good catch. We do need it as we override the `ColormapRegistery`. Moved the lazy loading from file to a private function to make this distinction clearer and still rely on `register`

PR #344, opened 2025-09-10
 Add cftime support for non-standard calendars

cvanelteren:
This PR introduces support for plotting data with non-standard calendars using `cftime`.

Key changes:
- Added `CFTimeConverter` to enable `matplotlib` to handle `cftime.datetime` objects.
- Implemented `AutoDatetimeLocator` and `AutoDatetimeFormatter` for automatic, "nice" tick generation on `cftime` axes, mimicking `matplotlib`'s behavior for standard `datetime` objects.
- Added a regression test using `pytest-mpl` to ensure visual consistency of plots with non-standard calendars.

This resolves issues where plots with non-standard calendars had incorrect or poorly formatted time axes.

Continuation of #289 

image

### DIFF
--- a/ultraplot/tests/test_tickers.py
+++ b/ultraplot/tests/test_tickers.py
@@ -1,0 +1,42 @@
+import pytest
+import numpy as np
+import xarray as xr
+import ultraplot as uplt
+
+
+@pytest.mark.mpl_image_compare
+def test_datetime_calendars_comparison():
+    # Time axis centered at mid-month
+    # Standard calendar
+    time1 = xr.date_range("2000-01", periods=120, freq="MS")
+    time2 = xr.date_range("2000-02", periods=120, freq="MS")
+    time = time1 + 0.5 * (time2 - time1)
+    # Non-standard calendar (uses cftime)
+    time1 = xr.date_range("2000-01", periods=120, freq="MS", calendar="noleap")
+    time2 = xr.date_range("2000-02", periods=120, freq="MS", calendar="noleap")
+    time_noleap = time1 + 0.5 * (time2 - time1)
+
+    da = xr.DataArray(
+        data=np.sin(np.arange(0.0, 2 * np.pi, np.pi / 60.0)),
+        dims=("time",),
+        coords={
+            "time": time,
+        },
+        attrs={"long_name": "low freq signal", "units": "normalized"},
+    )
+
+    da_noleap = xr.DataArray(
+        data=np.sin(2.0 * np.arange(0.0, 2 * np.pi, np.pi / 60.0)),
+        dims=("time",),
+        coords={
+            "time": time_noleap,
+        },
+        attrs={"long_name": "high freq signal", "units": "normalized"},
+    )
+
+    fig, axs = uplt.subplots(ncols=2)
+    axs.format(title=("Standard calendar", "Non-standard calendar"))
+    axs[0].plot(da)
+    axs[1].plot(da_noleap)
+
+    return fig
--- a/ultraplot/ticker.py
+++ b/ultraplot/ticker.py
@@ -7,8 +7,16 @@ import re
 from fractions import Fraction
 
 import matplotlib.axis as maxis
+import matplotlib.dates as mdates
 import matplotlib.ticker as mticker
+import matplotlib.transforms as mtransforms
+import matplotlib.units as munits
 import numpy as np
+
+try:
+    import cftime
+except ModuleNotFoundError:
+    cftime = None
 
 from .config import rc
 from .internals import ic  # noqa: F401
@@ -37,14 +45,17 @@ __all__ = [
     "SciFormatter",
     "SigFigFormatter",
     "FracFormatter",
+    "DatetimeFormatter",
+    "AutoDatetimeFormatter",
+    "AutoDatetimeLocator",
     "DegreeFormatter",
     "LongitudeFormatter",
     "LatitudeFormatter",
 ]
 
-REGEX_ZERO = re.compile("\\A[-\N{MINUS SIGN}]?0(.0*)?\\Z")
-REGEX_MINUS = re.compile("\\A[-\N{MINUS SIGN}]\\Z")
-REGEX_MINUS_ZERO = re.compile("\\A[-\N{MINUS SIGN}]0(.0*)?\\Z")
+REGEX_ZERO = re.compile("\\A[-\\N{MINUS SIGN}]?0(.0*)?\\Z")
+REGEX_MINUS = re.compile("\\A[-\\N{MINUS SIGN}]\\Z")
+REGEX_MINUS_ZERO = re.compile("\\A[-\\\N{MINUS SIGN}]0(.0*)?\\Z")
 
 _precision_docstring = """
 precision : int, default: {6, 2}
@@ -92,6 +103,18 @@ dms : bool, default: False
     ticks with minutes and seconds instead of decimals.
 """
 docstring._snippet_manager["ticker.dms"] = _dms_docstring
+
+
+_DEFAULT_RESOLUTION = "DAILY"
+_TIME_UNITS = "days since 2000-01-01"
+_RESOLUTION_TO_FORMAT = {
+    "SECONDLY": "%H:%M:%S",
+    "MINUTELY": "%H:%M",
+    "HOURLY": "%Y-%m-%d %H:%M",
+    "DAILY": "%Y-%m-%d",
+    "MONTHLY": "%Y-%m",
+    "YEARLY": "%Y",
+}
 
 
 def _default_precision_zerotrim(precision=None, zerotrim=None):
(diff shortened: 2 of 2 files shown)


cvanelteren:
~~Still fixing one test but rest should be ready for review.~~

ultraplot/ticker.py
```diff
@@ -704,7 +727,7 @@ def __call__(self, x, pos=None):  # noqa: U100
         sign = parts[1][0].replace("+", "")
         exponent = parts[1][1:].lstrip("0")
         if exponent:
-            exponent = f"10^{{{sign}{exponent}}}"
+            exponent = f"10^ {{{sign}{exponent}}}"
```
beckermr: Let's remove this space that was added.

```suggestion
            exponent = f"10^{{{sign}{exponent}}}"
```

ultraplot/ticker.py
```diff
@@ -808,20 +831,238 @@ def __call__(self, x, pos=None):  # noqa: U100
...
         else:
             if frac.numerator == 1 and symbol:  # numerator is +/-1
-                string = f"{symbol:s}/{frac.denominator:d}"
+                string = f"{symbol:s}/{{frac.denominator:d}}"
             elif frac.numerator == -1 and symbol:
-                string = f"-{symbol:s}/{frac.denominator:d}"
+                string = f"-{{symbol:s}}/{{frac.denominator:d}}"
             else:  # and again make sure we use unicode minus!
-                string = f"{frac.numerator:d}{symbol:s}/{frac.denominator:d}"
+                string = f"{frac.numerator:d}{symbol:s}/{{frac.denominator:d}}"
```
beckermr: This block of changes makes no sense. Why are some branches using `{{...}}` and others are not?

ultraplot/ticker.py
```diff
@@ -808,20 +831,238 @@ def __call__(self, x, pos=None):  # noqa: U100
...
         string = AutoFormatter._minus_format(string)
         return string
 
 
+class DatetimeFormatter(mticker.Formatter):
```
beckermr: This class name could refer to python datetimes. The class needs to be renamed to be more specific to cftime.

ultraplot/ticker.py
```diff
@@ -808,20 +831,238 @@ def __call__(self, x, pos=None):  # noqa: U100
...
+
+    def __call__(self, x, pos=None):
+        if isinstance(x, cftime.datetime):
+            dt = x
+        else:
+            dt = cftime.num2date(x, self._units, calendar=self._calendar)
+        return dt.strftime(self._format)
+
+
+class AutoDatetimeFormatter(mticker.Formatter):
```
beckermr: Same comment on this class name.

cvanelteren:
We may want to style the formatting further. In particular this

```
_RESOLUTION_TO_FORMAT = {
    "SECONDLY": "%H:%M:%S",
    "MINUTELY": "%H:%M",
    "HOURLY": "%Y-%m-%d %H:%M",
    "DAILY": "%Y-%m-%d",
    "MONTHLY": "%Y-%m",
    "YEARLY": "%Y",
}
```

cvanelteren:
Got the coverage to 65 percent from 0. Little bit done with adding tests and this should be good enough for now.

ultraplot/ticker.py
cvanelteren: Fixed already!

ultraplot/ticker.py
cvanelteren: done

ultraplot/ticker.py
cvanelteren: done

ultraplot/ticker.py
```diff
@@ -808,7 +832,7 @@ def __call__(self, x, pos=None):  # noqa: U100
             if frac.numerator == 1 and symbol:
                 string = f"{symbol:s}"
             elif frac.numerator == -1 and symbol:
-                string = f"-{symbol:s}"
+                string = f"-{{symbol:s}}"
```
beckermr:
```suggestion
                string = f"-{symbol:s}"
```

ultraplot/ticker.py
```diff
@@ -777,7 +801,7 @@ def __init__(self, symbol="", number=1):
         Parameters
         ----------
         symbol : str, default: ''
-            The constant symbol, e.g. ``r'$\pi$'``.
+            The constant symbol, e.g. ``r'$\\pi$'``.
```
beckermr:
```suggestion
            The constant symbol, e.g. ``r'$\pi$'``.
```

beckermr: I know linters complain here, but this bit of text is supposed to show an example bit of python code. The extra slash is not correct for a raw string.

ultraplot/ticker.py
beckermr: Somehow this one was missed.

ultraplot/ticker.py
```diff
@@ -822,6 +846,437 @@ def __call__(self, x, pos=None):  # noqa: U100
...
+                steps=[1, 2, 4, 7, 10],
+                prune="both",
+            ).tick_values(vmin, vmax)
+            for dt_num in days_numeric:
+                try:
+                    ticks_cftime.append(
+                        cftime.num2date(dt_num, self.date_unit, calendar=self.calendar)
+                    )
+                except ValueError:
+                    pass
```
beckermr: Why do error handling here but not in other cases?

ultraplot/ticker.py
```diff
@@ -822,6 +846,437 @@ def __call__(self, x, pos=None):  # noqa: U100
...
+            ticks_cftime = [t for t in ticks_cftime if t.year != 0]
+
+        # Convert the generated cftime.datetime objects back to numeric values
+        ticks_numeric = []
+        for t in ticks_cftime:
+            try:
+                num_val = cftime.date2num(t, self.date_unit, calendar=self.calendar)
+                if vmin <= num_val <= vmax:  # Final filter for actual view interval
+                    ticks_numeric.append(num_val)
+            except ValueError:  # Handle potential issues with invalid cftime dates
```
beckermr, line 1255: What is the effect of swallowing this error? Can you provide an example?

ultraplot/tests/test_tickers.py
```diff
@@ -0,0 +1,483 @@
...
+            cftime.date2num(
+                cftime.DatetimeGregorian(2000, 1, 1), "days since 2000-01-01"
+            ),
+            cftime.date2num(
+                cftime.DatetimeGregorian(2000, 12, 31), "days since 2000-01-01"
+            ),
+        ),  # Case 2: Monthly resolution
+    ],
+)
+def test_auto_datetime_locator_tick_values(start_date, end_date):
```
beckermr: This test should confirm the returned ticks are at the right resolution.

ultraplot/ticker.py
cvanelteren: Some calendars do not have certain dates. For example `proleptic_gregorian` does not have year 0. I will check if I added everything properly. Thanks.

cvanelteren: I updated the code to unify the safe casting. No rush in reviewing this I think we need to carefully merge this as I  am not too sure yet on the formatting. I think we may want to use the "largest unit" possible, e.g. years when resolution is years or months when it is months.

ultraplot/ticker.py
cvanelteren: See comment above.

ultraplot/tests/test_tickers.py
cvanelteren: added

cvanelteren:
If you can, can you provide some feedback on this pr @pgf?

ultraplot/tests/test_tickers.py
```diff
@@ -0,0 +1,576 @@
...
+    from ultraplot.ticker import AutoCFDatetimeLocator
+
+    locator = AutoCFDatetimeLocator(calendar="gregorian")
+    import cftime
+
+    ticks = locator.tick_values(start_date, end_date)
+    assert len(ticks) > 0  # Ensure ticks are generated
+
+
+@pytest.mark.parametrize(
```
beckermr, line 404: There are no test cases here that are expected to raise. Did you intend to add them? Should we add some?

ultraplot/ticker.py
```diff
@@ -94,6 +106,18 @@
...
+_DEFAULT_RESOLUTION = "DAILY"
+_TIME_UNITS = "days since 2000-01-01"
+_RESOLUTION_TO_FORMAT = {
+    "SECONDLY": "%H:%M:%S",
+    "MINUTELY": "%H:%M",
+    "HOURLY": "%Y-%m-%d %H:%M",
+    "DAILY": "%Y-%m-%d",
+    "MONTHLY": "%Y-%m",
+    "YEARLY": "%Y",
+}
```
beckermr: These defaults should be handled in the config files in the normal way.

ultraplot/ticker.py
```diff
@@ -822,6 +846,492 @@ def __call__(self, x, pos=None):  # noqa: U100
...
+
+class AutoCFDatetimeLocator(mticker.Locator):
+    """Determines tick locations when plotting `cftime.datetime` data."""
+
+    real_world_calendars = (
+        "gregorian",
+        "julian",
+        "proleptic_gregorian",
+        "standard",
+    )
```
beckermr: Can we get these from cftime automatically somehow? Otherwise we have to update them if they go out of sync and that will cause a bug eventually.

ultraplot/ticker.py
```diff
@@ -822,6 +846,492 @@ def __call__(self, x, pos=None):  # noqa: U100
...
+        self.calendar = calendar
+        self.date_unit = date_unit or _TIME_UNITS
+        if not self.date_unit.lower().startswith("days since"):
+            emsg = "The date unit must be days since for a NetCDF time locator."
+            raise ValueError(emsg)
+        self.resolution = _DEFAULT_RESOLUTION
+        self._cached_resolution = {}
+
+        # New: Default max ticks for Matplotlib's MaxNLocator, independent of resolution thresholds
+        self._max_display_ticks = 7  # A reasonable number for actual axis ticks
```
beckermr: Does this need to be in the config file?

ultraplot/ticker.py
```diff
@@ -897,3 +1407,141 @@ def __init__(self, *args, **kwargs):
...
+            if date_type is not None:
+                datemin = date_type(2000, 1, 1)
+                datemax = date_type(2010, 1, 1)
+            else:
+                # Fallback if date_type is None
+                datemin = cftime.DatetimeProlepticGregorian(2000, 1, 1)
+                datemax = cftime.DatetimeProlepticGregorian(2010, 1, 1)
+        except (TypeError, AttributeError):
+            datemin = cftime.DatetimeProlepticGregorian(2000, 1, 1)
+            datemax = cftime.DatetimeProlepticGregorian(2010, 1, 1)
```
beckermr: It seems like these dates are arbitrary and will go out of date in the future. Can we tie them to the current date somehow, or if not, they need to go in the config file.

ultraplot/ticker.py
```diff
@@ -897,3 +1407,141 @@ def __init__(self, *args, **kwargs):
...
         """
         super().__init__(*args, **kwargs)
+
+
+class CFTimeConverter(mdates.DateConverter):
+    """
+    Converter for cftime.datetime data.
+    """
+
+    standard_unit = "days since 2000-01-01"
```
beckermr: Defaults should go in the config file.

beckermr (changes requested):
Things are getting closer. In general, we need to figure out which of the various defaults and constants need to be configurable. My general feeling is that all of them should be configurable, but I also don't quite follow all of the code.

ultraplot/ticker.py
cvanelteren: There is no public api for this. We could use `_calendars`but that may be brittle. I now look for the calendar attribute.

ultraplot/ticker.py
cvanelteren: Sure let's do it.

ultraplot/tests/test_tickers.py
cvanelteren: Removed the block.

ultraplot/ticker.py
cvanelteren: I updated it to make the time relative to the current year.

ultraplot/ticker.py
```diff
@@ -1437,16 +1431,17 @@ def axisinfo(unit, axis):
...
             if date_type is not None:
-                datemin = date_type(2000, 1, 1)
-                datemax = date_type(2010, 1, 1)
+                datemin = date_type(year - 10, 1, 1)
+                datemax = date_type(year, 1, 1)
             else:
                 # Fallback if date_type is None
-                datemin = cftime.DatetimeProlepticGregorian(2000, 1, 1)
-                datemax = cftime.DatetimeProlepticGregorian(2010, 1, 1)
+                datemin = cftime.DatetimeProlepticGregorian(year - 10, 1, 1)
```
pgf: Calendar specific sub-classes like DatetimeProlepticGregorian are marked for removal from cftime in future releases, see: [here](https://github.com/Unidata/cftime/blob/312b8d7991de9ccfd0816e3f892b42f72ca81325/src/cftime/__init__.py#L8) and [here](https://github.com/Unidata/cftime/blob/312b8d7991de9ccfd0816e3f892b42f72ca81325/src/cftime/_cftime.pyx#L2083).
Should be replaced by cftime.datetime and the appropriate calendar.

cvanelteren: is there any quick way to get al the calendars besides using the private var `_calendars`?

ultraplot/ticker.py
cvanelteren: seems that this is deprecated so I will use the `_calendars` attribute for now

ultraplot/ticker.py
cvanelteren: Ignore my comment above. I removed the try block.

ultraplot/ticker.py
```diff
@@ -93,6 +105,15 @@
...
 
+_RESOLUTION_TO_FORMAT = {
+    "SECONDLY": "%H:%M:%S",
+    "MINUTELY": "%H:%M",
+    "HOURLY": "%Y-%m-%d %H:%M",
+    "DAILY": "%Y-%m-%d",
+    "MONTHLY": "%Y-%m",
+    "YEARLY": "%Y",
+}
+
```
beckermr: Should this be an rc option? How does it related to the similarly named one you added above?

beckermr (changes requested):
One more question. Looking good!

ultraplot/ticker.py
cvanelteren: Think this is double; it is already present under rc right now

cvanelteren: Will remove

cvanelteren: done

beckermr (approved):

cvanelteren:
Maybe I will add a bit more tests in the next few days but otherwise this is good to go.

cvanelteren:
Autoformatting prefix and suffix are bugged. Will do this in another PR. I think the coverage is now 78% which is good enough for now. Will wait for codecov confirmation

PR #349, opened 2025-09-16
Fix sharing

### LINKED ISSUE #348: Inconsistent behavior of axis label sharing
According to doc of [ultraplot.subplots](https://ultraplot.readthedocs.io/en/latest/api/ultraplot.ui.subplots.html) (or ultraplot.Figure), there are four levels of axis sharing.

- 0 or False: No axis sharing. This also sets the default spanx and spany values to False.

- 1 or 'labels' or 'labs': Only draw axis labels on the bottommost row or leftmost column of subplots. Tick labels still appear on every subplot.

- 2 or 'limits' or 'lims': As above but force the axis limits, scales, and tick locations to be identical. Tick labels still appear on every subplot.

- 3 or True: As above but only show the tick labels on the bottommost row and leftmost column of subplots.

- 4 or 'all': As above but also share the axis limits, scales, and tick locations between subplots not in the same row or column.

However, I found that ultraplot does not respect this convention exactly in ultraplot=1.60.2.
For example, `share=1`, `share="labs"`, `share=labels` do not provide the expected behavior (though these options do provide identical results). 
Here is a minimum example:

```
%matplotlib
import ultraplot as uplt
fig, axs = uplt.subplots(None, 2, 3, sharey='labels')
axs.format(ylabel='PDF')

fig, axs = uplt.subplots(None, 2, 3, sharey='labs')
axs.format(ylabel='PDF')

fig, axs = uplt.subplots(None, 2, 3, sharey=1)
axs.format(ylabel='PDF')
```

In contrast, ultraplot=1.50 works as expected (left panels in screenshot below).

Image

-----

However, interestingly, I notice that the example in [userguide](https://ultraplot.readthedocs.io/en/latest/subplots.html#Axis-label-sharing) using `uplt.figure` works as expected.
So there seems an inconsistency between uplt.subplots and Figure.subplots?

BTW, also note that in the [user guide](https://ultraplot.readthedocs.io/en/latest/subplots.html#axis-label-sharing), point 3 does not provide the full list of the **five** available options (compare to [proplot](https://proplot.readthedocs.io/en/latest/subplots.html#Axis-label-sharing)), though it says "five" just above the list. 
Is this change intentional or a misoperation?

cvanelteren:
Closes #348 

We recently introduced axis sharing to allow for top right axis sharing but inadvertently caused a bug for axis sharing levels 1. This PR should correct this issue.

##What's changed?
- Added a check `_determine_axis_visibility` to propagate the visibility if the sharing level is below 3.

### DIFF
--- a/ultraplot/figure.py
+++ b/ultraplot/figure.py
@@ -1058,7 +1058,7 @@ class Figure(mfigure.Figure):
         """
         We take the average here as the sharex and sharey should be the same value. In case this changes in the future we can track down the error easily
         """
-        return 0.5 * (self.figure._sharex + self.figure._sharey)
+        return min(self.figure._sharex, self.figure._sharey)
 
     def _add_axes_panel(self, ax, side=None, **kwargs):
         """
--- a/ultraplot/tests/test_sharing.py
+++ b/ultraplot/tests/test_sharing.py
@@ -1,0 +1,86 @@
+import pytest, ultraplot as uplt
+
+"""
+Sharing levels for subplots determine the visbility of the axis labels and tick labels.
+
+Axis labels are pushed to the border subplots when the sharing level is greater than 1.
+
+Ticks are visible only on the border plots when the sharing levels is greater than 2.
+
+Or more verbosely:
+    sharey = 0: no sharing, all labels and ticks visible
+    sharey = 1: share axis, all labels and ticks visible
+    sharey = 2: share limits
+    sharey = 3 or True, share both ticks and labels
+A similar story holds for sharex.
+"""
+
+
+@pytest.mark.parametrize("share_level", [0, "labels", "labs", 1, True])
+@pytest.mark.mpl_image_compare
+def test_sharing_levels_y(share_level):
+    """
+    Test sharing levels for y-axis: left and right ticks/labels.
+    """
+    fig, axs = uplt.subplots(None, 2, 3, sharey=share_level)
+    axs.format(ylabel="Y")
+    axs.format(title=f"sharey = {share_level}")
+    fig.canvas.draw()  # needed for checks
+
+    if fig._sharey < 3:
+        border_axes = set(axs)
+    else:
+        # Reduce border_axes to a set of axes for left and right
+        border_axes = set()
+        for direction in ["left", "right"]:
+            axes = fig._get_border_axes().get(direction, [])
+            if isinstance(axes, (list, tuple, set)):
+                border_axes.update(axes)
+            else:
+                border_axes.add(axes)
+    for axi in axs:
+        tick_params = axi.yaxis.get_tick_params()
+        for direction in ["left", "right"]:
+            label_key = f"label{direction}"
+            visible = tick_params.get(label_key, False)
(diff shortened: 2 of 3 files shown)


ultraplot/figure.py
```diff
@@ -1971,7 +1965,7 @@ def format(
         # When we apply formatting to all axes, we need
         # to potentially adjust the labels.
 
-        if len(axs) == len(self.axes) and self._get_sharing_level() > 0:
+        if len(axs) == len(self.axes) and (self._sharex >= 3 and self._sharey >= 3):
```
beckermr: Why is this not `self._sharex >= 0 and self._sharey >= 0`? 

The threshold has gone up and I don't follow why.

ultraplot/tests/test_sharing.py
```diff
@@ -0,0 +1,98 @@
...
+Axis labels are pushed to the border subplots when the sharing level is greater than 1.
+
+Ticks are visible only on the border plots when the sharing level is greater than 2.
+
+Or more verbosely:
+    sharey = 0: no sharing, all labels and ticks visible
+    sharey = 1: share axis, all labels and ticks visible
+    sharey = 2: share limits
+    sharey = 3 or True, share both ticks and labels
+A similar story holds for sharex.
```
beckermr, line 15: This comment should specify when limits are shared for all levels. Also IDK what "share axis" means here.

ultraplot/figure.py
cvanelteren: Correct -- not sure why i put the 3 here.

cvanelteren: Think some tests will fail for geo but will check

cvanelteren:
It is starting to shape up. Need to

- [ ] ensure that sharing between cartesian panels and geo axis occurs correctly. It can be force to look similar by forcing share = false but this is hacky. I need to somehow align the api to ensure that we look for the correct properties on the geo axes as the yticks and xticks are set off.
- [ ] need to check labelloc right for panels too on cartesian axes

cvanelteren:
Ok 19 tests left to fix, but some of them look good anyway. There are some things that needs fixing.

Primarily, I need to look at the interface between geo and cartesian panels.

cvanelteren:
Some tests fail due to image mismatch  but visually look ok; sometimes there is a title that is slightly shifted. The test `test_geo_with_panels` now fails because I copied the formatters. Panels for geo now copy the formatters for the relevant range (y or x) so that plotting is within the same data range. Formatters are updated too when relevant. This effectively creates some crowding on the x-values in `test_geo_with_panels`

beckermr:
This PR is pretty hefty and so will take time for me to review. Thanks for the hard work!

cvanelteren:
Thanks. Thinking of breaking it up  to make it more manageable. Will see if I can do that to ensure tractability.

cvanelteren:
Will do a diff tomorrow!

cvanelteren:
I split this PR into two, see #372 and #373. The should be handled in sequential order @beckermr

cvanelteren:
Superseded by #372

PR #353, opened 2025-09-29
Fix order of label transfer

### LINKED ISSUE #352: Top panel overrides xticklabel size
As it says on the tin:

```python
layout = [
    [4, 4, 4, 5, 5, 5],
    [1, 1, 2, 2, 3, 3],
]
fig, ax = uplt.subplots(layout, share=0)
ax.panel("r")
ax.panel("t")
ax.format(xlabel="test")
ax.format(xlabelsize=30)
``

`xlabelsize` gets ignored when I add top panel but works fine with right

### PR DESCRIPTION
**Pull request by cvanelteren**
Closes #352 

The `_transfer_label(src, dest)` function is responsible for transferring text and font properties from a source text object (`src`) to a destination text object (`dest`). Currently, the function sets all font properties of `dest` to match those of `src` every time it is called:

```python
font_properties = src.get_font_properties()
dest.set_fontproperties(font_properties)
```

This approach causes an issue when `_transfer_label` is called multiple times on the same `dest` object. Any customizations or updates made to `dest`'s font properties after the first transfer are lost, as they are overwritten by `src`'s properties on subsequent calls.

#### Example Scenario

1. `_transfer_label(src, dest)` is called — `dest` now has `src`'s font properties.
2. The user or another function updates `dest`'s font size or family.
3. `_transfer_label(src, dest)` is called again — the update is lost, as `src`'s font properties overwrite `dest`'s.

### Solution

To address this, the PR introduces a merge strategy for font properties:
- When transferring, the function now merges `src` and `dest` font properties.
- For each property, if `dest` has a value set, it is preserved; otherwise, the value from `src` is used.
- This ensures that any customizations to `dest` are retained across multiple calls to `_transfer_label`.

### Summary of Changes

- Added a `merge_font_properties` helper function.
- Updated `_transfer_label` to use merged font properties, preserving `dest`'s customizations.

### Impact

This change prevents accidental loss of user or programmatic updates to text object font properties, making the label transfer process more robust and predictable.

### DIFF
--- a/ultraplot/internals/labels.py
+++ b/ultraplot/internals/labels.py
@@ -10,7 +10,9 @@ from matplotlib.font_manager import FontProperties
 from . import ic  # noqa: F401
 
 
-def merge_font_properties(dest_fp, src_fp):
+def merge_font_properties(
+    dest_fp: FontProperties, src_fp: FontProperties
+) -> FontProperties:
     # Prefer dest_fp's values if set, otherwise use src_fp's
     return FontProperties(
         family=dest_fp.get_family() or src_fp.get_family(),
--- a/ultraplot/internals/labels.py
+++ b/ultraplot/internals/labels.py
@@ -4,8 +4,22 @@ Utilities related to matplotlib text labels.
 """
 import matplotlib.patheffects as mpatheffects
 import matplotlib.text as mtext
+from matplotlib.font_manager import FontProperties
+
 
 from . import ic  # noqa: F401
+
+
+def merge_font_properties(dest_fp, src_fp):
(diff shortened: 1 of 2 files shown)


cvanelteren:
> I don't think this is a viable strategy. It means that if you set a property once but then want to change it on the same plot, you cannot.

that's the current behavior this should fix that

cvanelteren:
You can see this from the example in #352

cvanelteren:
I modified to add src on first pass and then prefer dest to fix #352

beckermr:
Did you test this using the example code in the issue?

PR #365, opened 2025-10-08
Fix circular import rc setting

### LINKED ISSUE #364: circular import when loading rc with cycle
An error is thrown when setting up the circle in the ultraplotrc. The persists from #https://github.com/proplot-dev/proplot/issues/335#issue-1127212871.

```
---------------------------------------------------------------------------
ImportError                               Traceback (most recent call last)
Cell In[1], line 10
      8 import numpy as np
      9 import xcesm
---> 10 import ultraplot as pplt
     11 import matplotlib.pyplot as plt
     12 import nc_time_axis 

File /glade/work/che43/conda-envs/neu/lib/python3.11/site-packages/ultraplot/__init__.py:34
     32 # Import everything to top level
     33 with _benchmark("config"):
---> 34     from .config import *  # noqa: F401 F403
     35 with _benchmark("proj"):
     36     from .proj import *  # noqa: F401 F403

File /glade/work/che43/conda-envs/neu/lib/python3.11/site-packages/ultraplot/config.py:1825
   1821 rc_ultraplot = rcsetup._rc_ultraplot_default.copy()  # a validated rcParams-style dict
   1823 #: Instance of `Configurator`. This controls both `rc_matplotlib` and `rc_ultraplot`
   1824 #: settings. See the :ref:`configuration guide <ug_config>` for details.
-> 1825 rc = Configurator(skip_cycle=True)
   1827 # Deprecated
   1828 RcConfigurator = warnings._rename_objs(
   1829     "0.8.0",
   1830     RcConfigurator=Configurator,
   1831 )

File /glade/work/che43/conda-envs/neu/lib/python3.11/site-packages/ultraplot/config.py:769, in Configurator.__init__(self, local, user, default, **kwargs)
    763 """
    764 Parameters
    765 ----------
    766 %(rc.params)s
    767 """
    768 self._context = []
--> 769 self._init(local=local, user=user, default=default, **kwargs)

File /glade/work/che43/conda-envs/neu/lib/python3.11/site-packages/ultraplot/config.py:879, in Configurator._init(self, local, user, default, skip_cycle)
    877     user_path = self.user_file()
    878     if os.path.isfile(user_path):
--> 879         self.load(user_path)
    881 # Update from local paths
    882 if local:

File /glade/work/che43/conda-envs/neu/lib/python3.11/site-packages/ultraplot/config.py:1683, in Configurator.load(self, path)
   1681 rcdict = self._load_file(path)
   1682 for key, value in rcdict.items():
-> 1683     self.__setitem__(key, value)

File /glade/work/che43/conda-envs/neu/lib/python3.11/site-packages/ultraplot/config.py:788, in Configurator.__setitem__(self, key, value)
    783 def __setitem__(self, key, value):
    784     """
    785     Modify an `rc_matplotlib` or `rc_ultraplot` setting using dictionary notation
    786     (e.g., ``uplt.rc[name] = value``).
    787     """
--> 788     kw_ultraplot, kw_matplotlib = self._get_item_dicts(key, value)
    789     rc_ultraplot.update(kw_ultraplot)
    790     rc_matplotlib.update(kw_matplotlib)

File /glade/work/che43/conda-envs/neu/lib/python3.11/site-packages/ultraplot/config.py:997, in Configurator._get_item_dicts(self, key, value, skip_cycle)
    994 # Cycler
    995 # NOTE: Have to skip this step during initial ultraplot import
    996 elif contains("cycle") and not skip_cycle:
--> 997     from .colors import _get_cmap_subtype
    999     cmap = _get_cmap_subtype(value, "discrete")
   1000     kw_matplotlib["axes.prop_cycle"] = cycler.cycler("color", cmap.colors)

File /glade/work/che43/conda-envs/neu/lib/python3.11/site-packages/ultraplot/colors.py:32
     29 import numpy as np
     30 import numpy.ma as ma
---> 32 from .config import rc
     33 from .internals import ic  # noqa: F401
     34 from .internals import (
     35     _kwargs_to_args,
     36     _not_none,
   (...)
     40     warnings,
     41 )

ImportError: cannot import name 'rc' from partially initialized module 'ultraplot.config' (most likely due to a circular import) (/glade/work/che43/conda-envs/neu/lib/python3.11/site-packages/ultraplot/config.py)
```

cvanelteren:
Closes #364

This pull request resolves a circular import error that occurred when setting the `cycle` parameter from a user `rc` file. It also includes a major architectural refactoring of the configuration system to make it more robust, extensible, and easier to maintain.

### **The Problem**

The initial issue was caused by a circular dependency:
- `ultraplot.config` needed to import from `ultraplot.colors` to validate and process the `cycle` setting.
- `ultraplot.colors` needed to import the `rc` object from `ultraplot.config` for its own internal configuration.

This circular dependency made the application's initialization fragile and order-dependent, leading to the reported `ImportError`.

### **The solution: adding handlers to Configurator**

Instead of a temporary patch, this PR implements a more fundamental solution by refactoring the configuration system to follow the **Dependency Inversion Principle**.

The key changes are:

1.  **Decoupled `config` and `colors` modules:** The `Configurator` in `config.py` is now completely agnostic of color-specific logic. It no longer imports from `colors.py`.
2.  **Introduced a Handler Registry:** The `Configurator` now has a mechanism (`register_handler`) that allows other modules to register their own functions for handling specific settings.
3.  **Moved Logic to the Responsible Module:** The logic for processing the `cycle` setting has been moved into a `_cycle_handler` function within `colors.py`. This module is now responsible for registering its handler with the `Configurator`.
4.  **Removed Workarounds:** This architectural change completely eliminates the circular dependency, allowing for the removal of all previous workarounds, including the `skip_cycle` parameter and the complex deferred settings logic.

### Added tests

This refactoring was accompanied by a significant testing effort to ensure its correctness and prevent regressions:

1.  **New Handler Test:** A new test file, `test_handlers.py`, was created to specifically unit-test the new handler registration and execution mechanism.
2.  **Hardened Configuration Test:** The test for loading `rc` files (`test_config.py`) was completely rewritten to be more robust. After a long debugging process, the final version now directly and atomically tests the `rc.load()` method, ensuring the test is isolated from the user's local environment and any module-reloading side effects.

### PR fixes:

- The circular dependency is resolved.
- The initialization process is now more robust, predictable, and extensible.
- The changes are thoroughly tested with both new and improved tests.

### DIFF
--- a/ultraplot/config.py
+++ b/ultraplot/config.py
@@ -160,6 +160,34 @@ docstring._snippet_manager["rc.font_args"] = _font_docstring
 docstring._snippet_manager["rc.cmap_exts"] = _cmap_exts_docstring
 docstring._snippet_manager["rc.cycle_exts"] = _cycle_exts_docstring
 
+_rc_register_handler_docstring = """
+        Register a callback function to be executed when a setting is modified.
+
+        This is an extension point for "special" settings that require complex
+        logic or have side-effects, such as updating other matplotlib settings.
+        It is used internally to decouple the configuration system from other
+        subsystems and avoid circular imports.
+
+        Parameters
+        ----------
+        name : str
+            The name of the setting (e.g., ``'cycle'``).
+        func : callable
+            The handler function to be executed. The function must accept a
+            single positional argument, which is the new `value` of the
+            setting, and must return a dictionary. The keys of the dictionary
+            should be valid ``matplotlib`` rc setting names, and the values
+            will be applied to the ``rc_matplotlib`` object.
+
+        Example
+        -------
+        >>> def _cycle_handler(value):
+        ...     # ... logic to create a cycler object from the value ...
+        ...     return {'axes.prop_cycle': new_cycler}
+        >>> rc.register_handler('cycle', _cycle_handler)
+        """
+docstring._snippet_manager["rc.register_handler"] = _rc_register_handler_docstring
+
 
 def _init_user_file():
     """
@@ -773,32 +801,8 @@ class Configurator(MutableMapping, dict):
         self, name: str, func: Callable[[Any], Dict[str, Any]]
     ) -> None:
         """
-        Register a callback function to be executed when a setting is modified.
-
-        This is an extension point for "special" settings that require complex
-        logic or have side-effects, such as updating other matplotlib settings.
-        It is used internally to decouple the configuration system from other
-        subsystems and avoid circular imports.
-
-        Parameters
-        ----------
-        name : str
-            The name of the setting (e.g., ``'cycle'``).
-        func : callable
-            The handler function to be executed. The function must accept a
-            single positional argument, which is the new `value` of the
-            setting, and must return a dictionary. The keys of the dictionary
-            should be valid ``matplotlib`` rc setting names, and the values
-            will be applied to the ``rc_matplotlib`` object.
-
-        Example
-        -------
-        >>> def _cycle_handler(value):
-        ...     # ... logic to create a cycler object from the value ...
-        ...     return {'axes.prop_cycle': new_cycler}
-        >>> rc.register_handler('cycle', _cycle_handler)
+        %(rc.register_handler)s
         """
-
         self._setting_handlers[name] = func
 
     def __getitem__(self, key):


ultraplot/__init__.py
```diff
@@ -78,9 +78,13 @@
...
 # Validate colormap names and propagate 'cycle' to 'axes.prop_cycle'
 # NOTE: cmap.sequential also updates siblings 'cmap' and 'image.cmap'
-from .config import rc
+from .config import rc, _deferred_rc_settings
 from .internals import rcsetup, warnings
 
+for k, v in _deferred_rc_settings.items():
+    rc[k] = v
+_deferred_rc_settings.clear()
+
```
beckermr:
```suggestion
del _deferred_rc_settings
```

ultraplot/config.py
```diff
@@ -1678,7 +1684,11 @@ def load(self, path):
         """
         rcdict = self._load_file(path)
         for key, value in rcdict.items():
-            self.__setitem__(key, value)
+            if skip_cycle and "cycle" in key:
```
beckermr:
```suggestion
            if skip_cycle and key == "cycle":
```

Do we skip a key like "cycle_for_blah" too or do we only want to skip the key "cycle"? or do we skip keys that match say "*.cycle" or something?

ultraplot/config.py
```diff
@@ -991,12 +998,14 @@ def _get_item_dicts(self, key, value, skip_cycle=False):
 
         # Cycler
         # NOTE: Have to skip this step during initial ultraplot import
-        elif contains("cycle") and not skip_cycle:
-            from .colors import _get_cmap_subtype
+        elif contains("cycle"):
+            kw_ultraplot["cycle"] = value
```
beckermr: Why is this now being set when it was not before?

ultraplot/__init__.py
beckermr: We only need this on import once, right?

ultraplot/tests/test_config.py
```diff
@@ -17,3 +18,23 @@ def test_wrong_keyword_reset():
     fig, ax = uplt.subplots(proj="cyl")
     ax.format(coastcolor="black")
     fig.canvas.draw()
+
+
+def test_cycle_in_rc_file(tmp_path, monkeypatch):
```
beckermr: Please add this test as a separate PR so we can ensure it fails with the current code.

ultraplot/__init__.py
cvanelteren: Yes!

ultraplot/config.py
```diff
@@ -1678,7 +1687,11 @@ def load(self, path):
         """
         rcdict = self._load_file(path)
         for key, value in rcdict.items():
-            self.__setitem__(key, value)
+            if skip_cycle and key.endswith("cycle"):
```
beckermr:
```suggestion
            if skip_cycle and key.endswith(".cycle"):
```

Is this better?

ultraplot/tests/test_config.py
cvanelteren: I checked it does.

ultraplot/config.py
beckermr: Or maybe
```suggestion
            if skip_cycle and (key.endswith(".cycle") or key == "cycle"):
```

cvanelteren:
@beckermr I am pushing a more profound change that is more robust and less frail as the one pushed now. I will update the text above. One moment.

ultraplot/colors.py
```diff
@@ -30,6 +30,25 @@
 import numpy.ma as ma
 
 from .config import rc
+import cycler
```
beckermr: This import looks odd. Is cycler really a whole separate package?

cvanelteren:
Note that the handlers are only intended for settings that are "problematic" like the one with cycle. I think this PR is cleaner now as we separate the dependency between the `configurator` rand `colors`. The `Configurator` now does not need to know how colors handles the property.

ultraplot/config.py
```diff
@@ -18,6 +18,7 @@
 from collections.abc import MutableMapping
 from numbers import Real
 
+
 import cycler
```
beckermr, line 22: NVM I guess it is. Weird.

ultraplot/colors.py
cvanelteren: Yes (https://matplotlib.org/cycler/) but we need to use the internal one -- uno momento

cvanelteren: Ok just added the internal cycler.

cvanelteren:
Fingers crossed that the test pass :crossed_fingers:

cvanelteren:
Will wait for the tests to pass and then merge this. Just added the API docs for this -- in case we forget in the future (likely ;-)).

cvanelteren:
- [x] check produced docs on docstring 
image

PR #372, opened 2025-10-18
Test/update label sharing tests

cvanelteren:
The previous implementation had complex and distributed logic for determining tick and label visibility, which was difficult to maintain and extend. This refactoring centralizes the logic within the `Figure` class, making it more robust and easier to understand.
Summary
- Refactored figure-level ticklabel sharing and fixed a side-leak bug
- Standardized tick_params key usage across Matplotlib versions via per-axis key mapping
- Made panel-group sharing symmetric so top/right-only panels promote sharing level to 3
- Kept behavior and warnings for unsupported/mixed subplot types

Files changed
- ultraplot/ultraplot/figure.py
  - Refactor `_share_ticklabels` into smaller helpers:
    - New private helpers:
      - `_label_key_map()`: computes version-dependent label keys for tick_params.
      - `_group_axes_by_axis(axes, axis)`: groups axes by row (x) or column (y).
      - `_compute_baseline_tick_state(group_axes, axis, label_keys)`: builds baseline ticklabel state from main axes only; warns on unsupported/mixed types.
      - `_apply_border_mask(axi, baseline, sides, outer_axes)`: enforces figure borders and suppresses opposite panel sides; uses per-axis `_label_key`.
      - `_effective_share_level(axi, axis, sides)`: computes effective share level considering panel groups and neighbors; checks both relevant sides (fixes prior “last side wins” leak).
      - `_set_ticklabel_state(axi, axis, state)`: applies state to cartesian via set_tick_params (with x/y-to-bool cleanup) or to geo via `_toggle_gridliner_labels`.
    - Behavior retained:
      - Baseline built only from main axes.
      - Panels included in application.
      - Warnings for mixed types or unsupported axes.
      - Geo/cartesian pathways preserved.
    - Bug fix:
      - Removed reliance on a loop-local `side` after iteration; share-level promotion now checks all relevant sides for the given axis.
  - Panel creation: use per-axis `_label_key` mapping when reading/writing tick_params
    - When disabling main-axis tick labels due to a panel:
      - Replaced direct calls like `ax.xaxis.set_tick_params(labeltop=False)` with `ax.xaxis.set_tick_params(**{ax._label_key("labeltop"): False})` (and similarly for `labelbottom`, `labelleft`, `labelright`).
    - When setting panel axis labels:
      - Writes use `pax._label_key(...)`.
      - Reads use `ax._label_key(...)` for `get_tick_params()[...]`.
    - Cleanup:
      - Removed a stray debug print in the effective share-level logic (seen during edit context).
- ultraplot/ultraplot/axes/base.py
  - `_apply_auto_share()`:
    - Make panel-group membership symmetric:
      - For x: if there are top panels, set `self._panel_sharex_group = True` so the main axes is included in the x-panel group even without bottom panels.
      - For y: if there are right panels, set `self._panel_sharey_group = True` so the main axes is included in the y-panel group even without left panels.
    - Outcome:
      - With only right (or top) panels present, sharing level now correctly promotes to 3 for that axis on the main axes.
- Tests
  - Numerous tests are added and changed to ensure that we offer a robust verification of the new logic.

Behavioral notes
- Matplotlib ≤ 3.9 compatibility: all new writes/reads to tick_params use per-axis `_label_key` to map `labeltop/labelbottom` to `labelright/labelleft` as needed.
- Warnings for mixed/unsupported subplot types remain unchanged.
- Figure `stale` semantics preserved.

User-visible impact
- Ticks should be shared "as one would expect"; if one can draw a line from one axis without hitting another axis, the ticks should be visible when appropriate, otherwise they are shared.

### DIFF
--- a/ultraplot/tests/test_subplots.py
+++ b/ultraplot/tests/test_subplots.py
@@ -20,6 +20,18 @@ def test_align_labels():
 
 
 @pytest.mark.mpl_image_compare
+@pytest.mark.parametrize("share", [0, 1, 2, 3, 4])
+def test_all_share_levels(share):
+    N = 10
+    x = np.arange(N)
+    fig, ax = uplt.subplots(nrows=2, ncols=2, share=share)
+    ax[0].plot(x, x)
+    ax[-1].plot(x * 1000, x * 1000)
+    ax.format(xlabel="xlabel", ylabel="ylabel", suptitle=f"Share level={share}")
+    return fig
+
+
+@pytest.mark.mpl_image_compare
 def test_share_all_basic():
     """
     Test sharing level all.


ultraplot/gridspec.py
```diff
@@ -195,19 +195,20 @@ def _get_rows_columns(self, ncols=None):
         row2, col2 = divmod(self.num2, ncols)
         return row1, row2, col1, col2
 
-    def _get_grid_span(self, hidden=False) -> (int, int, int, int):
+    def _get_grid_span(self, hidden=True) -> (int, int, int, int):
```
beckermr: Why did this default change?

beckermr:
The following tests need a closer look:

- test_panels_without_sharing_1 (change in tick labels on main plot y-axis)
- test_uneven_levels (change in x-axis tick labels)
- test_geo_with_panels (change in bottom-right panel tick labels and ticks)
- test_align_labels (change in panel E x tick labels)
- test_share_all_basic (everything looks broken, x axis range is wrong)

Some of these cases look like bugs. Others might be the desired outcome. I don't know which is which.

cvanelteren:
Thanks for the careful review @beckermr! I think the ticks are related to the axes limits not being shared. Not sure how that changed -- but will get to the bottom of this. The tick params for the axis I fixed locally. Will check the final ones and recheck. After this I will get back to you.

ultraplot/gridspec.py
cvanelteren: I think I needed it by default -- but will move it back and adjust accordingly.

cvanelteren:
> What is the difference between this PR and the other one that is similar?

I split it to avoid overwhelming number of changes; #372 and #372 are together the same as #349

cvanelteren:
Weird I set the limits explicitly in the function `_add_panel_axes` and then the sharing broke. I removed that stuff now and check the generated results.

cvanelteren:
Looking better!

What should we decide on `test_geo_with_panels` do we set the ticks to be on depending on the main axis? @beckermr

beckermr:
I took a look again. I see the following:

- test_panels_without_sharing_1: The panel on the right axis has its own tick labels in the new code. Is that correct?
- test_geo_with_panels: The panels do not have tick labels and the case you asked about.

> What should we decide on test_geo_with_panels do we set the ticks to be on depending on the main axis? 

Yes. I'd say if the main axis is a geo axis, then the panel should have tick labels on since its axis is not shared.

cvanelteren:
For `test_panels_without_sharing` the right plot should have the ticks where they are now:

```python
    fig, axs = uplt.subplots(ncols=2, share=True, refwidth=1.5, includepanels=False)
    axs.panel("left", share=False)
    fig.format(ylabel="ylabel", xlabel="xlabel")
```

The logic for the current panel for the geo would need to change if we want that to be off for the test that is currently failing. Right now it will check if it is on a border but will automatically draw it for right axes; I am fine with it as is, as the default right now also puts these ticks on. However, it may make more sense to make it dependent on the main axis in general. Since in this case all ticks are off, it should also be off for the panel; but open to both tbh.

beckermr:
The panels definitely need tick labels in the `test_geo_with_panels`. We don't allow shared geo and cartesian axes so none of the panel's choices should be shared with the main axis in this case and thus the tick labels should be set with their default of being on.

cvanelteren:
We should and do support geo panels (see https://github.com/Ultraplot/UltraPlot/issues/237) but only for rectilinear projections. The panels currently behave in a similar fashion were they check if the main axis has the ticks on. For the right panels these are by default off always but for the left the would turn on. This is similar to how the cartesian panels operate. I will add a few more test to make this clear.

For example:
image

but for a Cartesian right panel we get:

image

For the GeoAxes, however, by default all the ticks are off which causes a disconnect.

cvanelteren:
Was using a wrong python env for the pics above which is why they look odd but the story still holds.

beckermr:
I see. Thanks for the explanation. 

The only thing I see left is that in the latest image comparison the panel `test_geo_with_panels` has a problem with the xtick locations and/or labels.

cvanelteren:
The crowding occurs due to the size of the font for that panel. We can make it like this:

image

cvanelteren:
Found the issue with the geo labels; `auto` had to be set when calling `_set_lim`.

beckermr:
What is this error on MPL 39?

```
FAILED ultraplot/tests/test_figure.py::test_suptitle_alignment - KeyError: 'labeltop'
FAILED ultraplot/tests/test_geographic.py::test_panels_geo - KeyError: 'labeltop'
FAILED ultraplot/tests/test_inset.py::test_inset_basic - KeyError: 'labeltop'
FAILED ultraplot/tests/test_statistical_plotting.py::test_panel_dist - KeyError: 'labeltop'
FAILED ultraplot/tests/test_subplots.py::test_title_deflection - KeyError: 'labeltop'
FAILED ultraplot/tests/test_subplots.py::test_panel_sharing_top_right[layout0] - KeyError: 'labeltop'
FAILED ultraplot/tests/test_subplots.py::test_panel_ticklabels_all_sides_share_and_no_share[True] - KeyError: 'labeltop'
```

cvanelteren:
The key labelling changed in mpl 3.10; I thought I provided a fix for this with a translation function. Will carefully examine. Still ironing out some odd edge cases

cvanelteren:
Will add some more tests tomorrow. Every combination is passing on the tests now indicating that the labels are correctly used internally. May also change how the private conversion handles directly by making it available as a global var. A bit annoying to make this dual compatibility.

cvanelteren:
I think I added sufficient number of tests now to make this merge. I will wait for the tests to finish. I updated the text on top for internal reference so we can track what changed exactly. As most of the stuff is in wards facing it is key to keep track of what we changed so we don't get lost  -- as I did many times in this PR. 

I think the sharing is more robust for the ticks now and allows for flexible sharing. I will turn to #373 as that one follows this PR and probable needs updating.

cvanelteren:
image
Failing tests are expected. Merging.

PR #373, opened 2025-10-18
Refactor: Improve tick and label sharing logic

cvanelteren:
This commit introduces a major refactoring of the tick and label sharing mechanism in UltraPlot.

The previous implementation had complex and distributed logic for determining tick and label visibility, which was difficult to maintain and extend. This refactoring centralizes the logic within the `Figure` class, making it more robust and easier to understand.

Key changes:
- A new `_share_ticklabels` method in `Figure` now handles all tick label sharing.
- The `_get_border_axes` method has been improved to be more accurate.
- The `_Crawler` utility in `ultraplot/utils.py` has been rewritten to better handle complex layouts with panels and mixed axes types.
- Redundant and complex logic has been removed from `CartesianAxes`, `GeoAxes`, and other modules.

### DIFF
--- a/ultraplot/gridspec.py
+++ b/ultraplot/gridspec.py
@@ -412,13 +412,15 @@ class GridSpec(mgridspec.GridSpec):
             num1, num2 = self._encode_indices(num1, num2)
         return _SubplotSpec(self, num1, num2)
 
-    def _encode_indices(self, *args, which=None):
+    def _encode_indices(self, *args, which=None, panel=False):
         """
-        Convert indices from the "unhidden" gridspec geometry into indices for the
+        Convert indices from the selected gridspec geometry into indices for the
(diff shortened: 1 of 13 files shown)


ultraplot/gridspec.py
```diff
@@ -195,19 +195,20 @@ def _get_rows_columns(self, ncols=None):
         row2, col2 = divmod(self.num2, ncols)
         return row1, row2, col1, col2
 
-    def _get_grid_span(self, hidden=False) -> (int, int, int, int):
+    def _get_grid_span(self, hidden=True) -> (int, int, int, int):
```
beckermr, line 198: Why did this default change?

beckermr:
I looked at some of the test failures and they definitely need to be fixed in some cases.

cvanelteren:
This PR is not needed anymore after #372

PR #376, opened 2025-10-21
Allow non-rectangular grids to use side labels

### LINKED ISSUE #375: Error adding rowlabels for custom gridspec
Hey!
Will begin by thanking you for continuing the work with ultraplot!

I've encountered what I think is an error with the formatting of row labels using a custom gridspec. If the first column of a row is empty, the row does not seem to count. Here is a minimal example:

```python
import ultraplot as plt
gridspec = [
    [1, 1, 2, 2],
    [0, 3, 3, 0],
    [4, 4, 5, 5],
]

fig, ax = plt.subplots(
    gridspec,
)
ax.format(rowlabels=["1", "2", "3"])
```
which raises an error: `ValueError: Got 3 left labels but found 2 axes along the left side of the figure.`

If I add a plot to the first column of the 2nd row, things works as expected.

```python
gridspec = [
    [1, 1, 2, 2],
    [6, 3, 3, 0],
    [4, 4, 5, 5],
]

fig, ax = plt.subplots(
    gridspec,
)
ax.format(rowlabels=["1", "2", "3"])
```
produces 

Image

cvanelteren:
Closes #375 

The checks in `def _get_align_axes(self, side)` assumes that the plots are on a rectangular grid. However, for non rectangular layouts this causes an error. This PR uses `_get_border_axes` to determine the outside.

image

### DIFF
--- a/ultraplot/tests/test_subplots.py
+++ b/ultraplot/tests/test_subplots.py
@@ -348,3 +348,18 @@ def test_non_rectangular_outside_labels_top():
     ax.format(leftlabels=[1, 3, 4])
     ax.format(toplabels=[1, 2])
     return fig
+
+
+@pytest.mark.mpl_image_compare
+def test_outside_labels_with_panels():
+    fig, ax = uplt.subplots(
+        ncols=2,
+        nrows=2,
+    )
+    # Create extreme case where we add a lot of panels
+    # This should push the left labels further left
+    for idx in range(5):
+        ax[0].panel("left")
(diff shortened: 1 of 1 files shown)


cvanelteren:
Need to fix a few edge cases

cvanelteren:
Added a few examples in #375

cvanelteren:
Reminder for myself. I use the grid reference here directly which changes in #373 and #372

PR #377, opened 2025-10-22
Add version checker for UltraPlot

cvanelteren:
This PR would introduce a version check to inform users if a newer version of the package is available.


cvanelteren:
Hmm ok I hear you. The uptick for the more recent versions is not as high as the previous ones. I thought nudging people to upgrade would be a good idea. We could leave it out and put it in the refs indeed; do agree that package managers should handle this. Will think about it a bit and otherwise close this.

cvanelteren:
Turned the default to off and added some tests.

PR #383, opened 2025-10-25
Feature: Threadsafe RcParams

cvanelteren:
This PR makes our configuration system thread-safe and adds support for thread-local storage. This is particularly important when running tests in parallel (e.g., with pytest-xdist), where each test thread needs to modify configuration parameters without affecting other threads.

I started playing around with a recent PR that would be merged in the future(https://github.com/matplotlib/pytest-mpl/pull/242#issuecomment-3418829567). This PR would prep for making it threadsafe -- speeding up local development. GHA has 2 threads per worker so it could be beneficial there, and otherwise not hurt.

# The Problem

Currently, when multiple threads try to access or modify configuration parameters at the same time, we can get race conditions and data corruption. Also, there's no way for a thread to make temporary changes that are automatically cleaned up when the thread exits.

# Proposed fix

The PR implements a thread-safe version of `_RcParams` that:
1. Uses a reentrant lock (`threading.RLock`) to prevent race conditions
2. Adds thread-local storage so each thread can have its own temporary configuration changes
3. Implements a context manager interface for clean management of thread-local changes
4. Properly merges thread-local changes back into the main configuration

**Added tests**
- `test_rcparams_thread_safety`

### DIFF
--- a/ultraplot/axes/base.py
+++ b/ultraplot/axes/base.py
@@ -6,10 +6,11 @@ Implements basic shared functionality.
 import copy
 import inspect
 import re
+import sys
 import types
-from numbers import Integral, Number
-from typing import Union, Iterable, MutableMapping, Optional, Tuple
 from collections.abc import Iterable as IterableType
+from numbers import Integral, Number
+from typing import Iterable, MutableMapping, Optional, Tuple, Union
 
 try:
     # From python 3.12
@@ -34,12 +35,11 @@ import numpy as np
 from matplotlib import cbook
 from packaging import version
 
-from .. import legend as plegend
 from .. import colors as pcolors
 from .. import constructor
+from .. import legend as plegend
 from .. import ticker as pticker
 from ..config import rc
-from ..internals import ic  # noqa: F401
 from ..internals import (
     _kwargs_to_args,
     _not_none,
@@ -51,6 +51,7 @@ from ..internals import (
     _version_mpl,
     docstring,
     guides,
+    ic,  # noqa: F401
     labels,
     rcsetup,
     warnings,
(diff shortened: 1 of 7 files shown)


cvanelteren:
Need to do some final checks to see how this interacts with the `Configurator` class. Furthermore, we need to override the default MPL rc params as it is not threadsafe currently. Perhaps it is easier to control this in `Configurator` in general. Will get back to this when I have the time.

ultraplot/internals/rcsetup.py
```diff
@@ -587,22 +614,33 @@ def __iter__(self):
...
+                # If key doesn't exist in validation, add it with default validator
+                self._validate[key] = lambda x: x
+                # Re-validate with new validator
+                value = self._validate[key](value)
+            except (ValueError, TypeError) as error:
+                raise ValueError(f"Key {key}: {error}") from None
+            if key is not None:
+                # Store in both thread-local storage and main dictionary
+                self._local.changes[key] = value
+                dict.__setitem__(self, key, value)
```
beckermr: This block doesn't make sense to me. I think we need to test if we're in the context manager block and then only make a local change if we are. Otherwise threads will make changes to the global settings which could interfere with each other.

cvanelteren: I am leaning towards moving this logic (thread safety) inside the `Configurator` we both need to control ultraplot's rc and matplotlib's rc making this PR a bit moot and #384 an essential step in this direction.

cvanelteren:
> I think there may be a bug in the logic.

Think you are right. I believe I fixed it now. I updated the docstring to reflect the intent of the block. The idea is that we have a main thread which contains the correct information, but threads can locally modify the rc params without affecting the main thread. I will document this as we are not resolving any conflicting settings that happens on the threads themselves, the main thread contains the "correct" information.

cvanelteren:
hmm the threadsafety wrapping broke something -- will address this when I have more time.

cvanelteren:
The issue is not caused by thread safety, I think. Closing the branch for now as the work is not necessary and only related to testing.

PR #384, opened 2025-10-27
Refactor: make rc_ultraplot and rc_matplotlib internal

cvanelteren:
This PR moves the global variables of `rc_ultraplot` and `rc_matplotlib` internal to the `Configurator` class.

# Why?

Currently we have global variables that are being passed around which causes us to track `rc_matplotlib`, `rc_ultraplot`and `rc` separately. If we wanna move to a thread safe `rc` setter our `Configurator` class needs to encapsulate the rc parameters properly. It eases our understanding of the how UltraPlot is configured without having to see where the `rc` are update. The general use of `rc` is focused on the `Configurator` anyway so making them apart of this setting is a logical encapsulation.

It also moves the logic from `__init__.py`  to a new `sync`  class method. 

# Todos
- [x] Refactor `rc` dicts inside `Configurator`
- [x] Remove old deprication warnings
- [x] adjust document if necessary
- [x] add unittests

### DIFF
--- a/.github/workflows/build-ultraplot.yml
+++ b/.github/workflows/build-ultraplot.yml
@@ -113,6 +113,21 @@ jobs:
           cache-environment: true
           cache-downloads: false
 
+      - name: Resolve baseline reference
+        id: baseline-ref
+        run: |
+          if [ "${IS_PR}" = "true" ]; then
+            BASE_REF="${{ github.event.pull_request.base.ref }}"
+            git fetch origin "${BASE_REF}"
+            BASE_SHA="$(git rev-parse "origin/${BASE_REF}")"
+          else
+            BASE_REF="${GITHUB_REF_NAME:-main}"
+            BASE_SHA="${GITHUB_SHA}"
+          fi
+          echo "base_ref=${BASE_REF}" >> "${GITHUB_OUTPUT}"
+          echo "base_sha=${BASE_SHA}" >> "${GITHUB_OUTPUT}"
+          echo "Resolved baseline ref=${BASE_REF} sha=${BASE_SHA}"
+
       # Cache Baseline Figures (Restore step)
       - name: Cache Baseline Figures
         id: cache-baseline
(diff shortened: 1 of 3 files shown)


cvanelteren:
Note: I may just remove the inputs as the user should not be concerned with the defaults rc`s

cvanelteren:
Still reaching the time-out ...

beckermr (changes requested):
Is there a way we can do this without breaking the API?

ultraplot/config.py
```diff
@@ -405,64 +402,6 @@ def _infer_ultraplot_dict(kw_params):
     return kw_ultraplot
 
 
-def config_inline_backend(fmt=None):
```
beckermr, line 400: Why was this function removed?

cvanelteren: It is moved as a class function now. It was moved there as it was referencing as global var (`rc_ultraplot`). The encapsulation is more appropriate with this recent change.

cvanelteren: It is here now: https://github.com/Ultraplot/UltraPlot/pull/384/files#diff-849c439d454bf26e1102dce232b08a417344e1d97a4add022d4088cd00b743caR1060

Note this PR looks more heavy than it is. I moved `_int`  and `__init__` to the top of the class to make it read easier. The only major change is moving `rc_ultraplot` and `rc_matplotlib` as class variables and adjusting the logic elsewhere in the code (mainly a couple of references in `figure.py`)

cvanelteren: To add to this. I looked into to how this function is called, and it also happens inside `Configurator`. I first thought the function was called by `IPython` directly -- but it is not.

cvanelteren:
> Is there a way we can do this without breaking the API?

It doesn't change the api that much. Users are guided to use `rc`  (the `Configurator` class) anyway. That remains unchanged here. The only difference here is that we don't have to manage the rc params separately. Before we had 3 objects (`rc`,`rc_ultraplot` and `rc_matplotlib`). The `rc` objects would ensure that we are retrieving or setting the correct matplotlib settings. I think it is cleaner to just use `rc` and make the respective `rc` params for ultraplot and matplotlib available from this configuration rather bookkeeping them separately. In this way, we could also ensure that we can do thread safety by providing blocks and proper context protection in the class.

beckermr:
An API change is an API change even if it is small. We'll need to bump to 2.0 after this IMHO.

cvanelteren:
I see a version increase merely as a version increase, but perhaps we should add an api warning for few Y releases (x.Y) before we make the change.

ultraplot/config.py
```diff
@@ -1797,6 +1815,20 @@ def save(self, path=None, user=True, comment=None, backup=True, description=Fals
...
+        """
+        for _src in (self.rc_ultraplot, self.rc_matplotlib):
+            for _key in _src:  # loop through unsynced properties
+                if "color" not in _key:
+                    continue
+                try:
+                    _src[_key] = _src[_key]
+                except ValueError as err:
+                    warnings._warn_ultraplot(f"Invalid user rc file setting: {err}")
+                    _src[_key] = "black"  # fill value
```
beckermr, line 1822: I don't follow how this code works to be honest. Also the method should probably be called `sync_colors` since it only deals with colors?

cvanelteren: Code was originally part of init. Merely wrapped it. I think it triggers the validators.

beckermr (approved):

ultraplot/config.py
cvanelteren: > I don't follow how this code works to be honest. Also the method should probably be called `sync_colors` since it only deals with colors?

actually taht is a good point -- will change it for clarity

beckermr: I don't see how the settings from one rc object end up in the other. It must happen on the backend as you say.

cvanelteren: At least for the `rc_ultraplot` it will run a validator on setting a key (see `rcsetup.py`); matplotlib prolly has a similar setting.

Initially I was also scratching my head at this section in the code.  Will replace it as we progress to something a bit cleaner.

cvanelteren:
~~I believe this is not the issue -- I am closing it without removing the branch.~~ Wrong PR!

beckermr (approved):

PR #385, opened 2025-10-27
Bump to python 3.14

cvanelteren:
Rejoice Python 3.14 is coming and we should prepare. This PR bumps the max python version.

https://www.python.org/downloads/release/python-3140/


cvanelteren:
Does not work at this moment due to basemap not being compatible. Going back to draft.

cvanelteren:
I am running python 3.14 now locally but it is a bit funky to setup. At least the code is running fine so we could upgrade when the packages are figured out (in particular basemap).

cvanelteren:
Error seems to be related to the GHA -- been running 3.14 fine for the last weeks.

PR #387, opened 2025-10-28
Hotfix: minor update in sharing logic

cvanelteren:
By default set interfaces to be a border in case we are looking across a a different in axis types, e.g. cartesian and or geo.


cvanelteren:
@beckermr any clue why our GHA are timing out? Current time out is set to 60 not sure if this is cumulative or per job.

beckermr:
I suspect the job is using too much memory, not timing out.

PR #391, opened 2025-10-29
Skip test_demos on gha

cvanelteren:
I am suspecting that `test_demos` are drawing too much RAM. This PR skips them on GHA while still recording them for our coverage.


cvanelteren:
Bypassing rules as the runner is canceled due to it being present on main.

PR #406, opened 2025-11-12
Add external context mode for axes

### LINKED ISSUE #405: Superfluous legend entries when plotting with seaplot
Using seaborn's `pointplot` and `lineplot`, extra legend entries appear with ultraplot, but not with matplotlib.

```python
import pandas as pd
import seaborn
import ultraplot as uplt

fig, ax = uplt.subplots()
df = pd.DataFrame(
    {
        "xcol": np.concatenate([np.arange(10)] * 2),
        "ycol": np.concatenate([np.arange(10), 1.5 * np.arange(10)]),
        "hcol": ["h1"] * 10 + ["h2"] * 10,
    }
)

seaborn.lineplot(df, x="xcol", y="ycol", hue="hcol")
```

This results in
Image

With `fig, ax = matplotlib.pyplot.subplots()`, there are no extra entries (`y`, `ymin`),

Image

cvanelteren:
Closes #405
This PR resolves spurious legend entries when plotting with seaborn on UltraPlot axes by tagging seaborn-generated helper artists (e.g., CI bands) as synthetic and excluding them during legend collection, suppressing mid-call legend/colorbar creation within seaborn to avoid duplicates, and disabling auto label inference for primary lines in seaborn contexts so legends show only hue categories unless users explicitly label additional bands.

# Key changes
- Tag seaborn-generated helper artists (CI bands via `fill_between`, error shading) as synthetic:
  - Force underscore labels (e.g., `_ultraplot_fill`, `_ultraplot_shade`, `_ultraplot_fade`) unless user explicitly provides a label.
  - Mark artists with `_ultraplot_synthetic=True`.
- Legend collection rejects synthetic helper artists (including when seaborn passes explicit handles).
- Suppress on-the-fly legend/colorbar creation during seaborn calls to avoid duplicates.
- Disable auto label inference for primary line plots when called from seaborn to prevent stray “y” labels.

# Behavior
- For seaborn `lineplot(..., hue=...)`, legend now shows only hue categories.
- CI/error bands are hidden unless the user requests labels (`shadelabel`/`fadelabel` or `label=`).
- User-specified labels are respected and shown.

Compatibility
- No changes for non-seaborn workflows.
- Users can explicitly opt-in to show shaded bands by passing labels.

### DIFF
--- a/ultraplot/axes/plot.py
+++ b/ultraplot/axes/plot.py
@@ -5373,6 +5373,11 @@ class PlotAxes(base.Axes):
             center_levels=center_levels,
             **kw,
         )
+        # Change the default behavior for weights/C to compute
+        # the total of the weights, not their average.
+        reduce_C_function = kw.get("reduce_C_function", None)
+        if reduce_C_function is None:
+            kw["reduce_C_function"] = np.sum
         norm = kw.get("norm", None)
         if norm is not None and not isinstance(norm, pcolors.DiscreteNorm):
             norm.vmin = norm.vmax = None  # remove nonsense values
--- a/ultraplot/tests/test_plot.py
+++ b/ultraplot/tests/test_plot.py
@@ -693,3 +693,38 @@ def test_curved_quiver_color_and_cmap(rng, cmap):
     fig, ax = uplt.subplots()
     ax.curved_quiver(X, Y, U, V, color=color, cmap=cmap)
     return fig
+
+
+def test_histogram_norms():
+    """
+    Check that all histograms-like plotting functions
+    use the sum of the weights.
+    """
+    rng = np.random.default_rng(seed=100)
+    x, y = rng.normal(size=(2, 100))
+    w = rng.uniform(size=100)
+
+    fig, axs = uplt.subplots()
+    _, _, bars = axs.hist(x, weights=w, bins=5)
+    tot_weights = np.sum([bar.get_height() for bar in bars])
+    np.testing.assert_allclose(tot_weights, np.sum(w))
+
+    fig, axs = uplt.subplots()
+    _, _, _, qm = axs.hist2d(x, y, weights=w, bins=5)
+    tot_weights = np.sum(qm.get_array())
+    np.testing.assert_allclose(tot_weights, np.sum(w))
+
+    fig, axs = uplt.subplots()
+    pc = axs.hexbin(x, y, weights=w, gridsize=5)
+    tot_weights = np.sum(pc.get_array())
+    np.testing.assert_allclose(tot_weights, np.sum(w))
+
+    # check that a different reduce_C_function produces
+    # a different result
+    fig, axs = uplt.subplots()
+    pc = axs.hexbin(x, y, weights=w, gridsize=5, reduce_C_function=np.max)
+    tot_weights = np.sum(pc.get_array())
+    # check they are not equal and that the different is not
+    # due to floating point errors
+    assert tot_weights != np.sum(w)
+    assert not np.allclose(tot_weights, np.sum(w))


cvanelteren:
Note that ideally we just ask the seaborn developers to make their private labels available with `_` but this would be easier and sets a precedent for interfacing with other packages to control how it behaves within ultraplot.

cvanelteren:
Note to self -- run the suite before pushing PRs!

cvanelteren:
The import statements are handled by my editor -- bit annoying.

cvanelteren:
Will add a few more tests to reach the target and then open it for review

cvanelteren:
You’re right that stack-frame detection is brittle. I’ve added an explicit, opt‑in “external mode” so users can mark an axes as being driven by an external plotting library. When external mode is on, we defer on-the-fly guide creation (legends/colorbars) during plotting; users explicitly call `ax.legend()` or `ax.colorbar()` once at the end. This avoids depending on stack inspection and generalizes to any library, not just seaborn.

A few important points:
- Backward compatibility: Default remains unchanged — we still use our current auto-detection by default. External mode is additive and opt‑in, so out-of-the-box behavior isn’t affected.
- Scope and mixing: External mode is per-axes and can be scoped with a context manager. That makes it easy to mix external and UltraPlot-native calls on the same subplot; just wrap the external calls in `with ax.external(): ...` and finish with `ax.legend()`.
- Extensibility: We can later thread this flag through more integration points if we want to “turn more of UltraPlot off” while the flag is set (e.g., skipping our automatic label inference or disabling synthetic tagging). For now, we’ve taken the least intrusive step that addresses the fragility concern: make guide creation deterministic and user-controlled.

If you’re happy with this direction, we can also add:
- A figure-wide helper (`fig.set_external(True|False)`) and an rcParam to force external mode for newly created axes in a context. Both would be non-breaking additions.

Want me to expand this to cover auto-label inference and synthetic tagging when external mode is set, or leave that for a follow-up?

beckermr:
Yes let's do this all in one go. We should do the following.

- Remove all of the stack-frame stuff (I don't want to ship that and then have to maintain it. People should use the generic mechanism.)
- Implement quto-label inference and synthetic tagging when external mode is set.

I don't think a fully external figure instance is that useful since in that case why not just use the external library? So we can skip that feature for now.

cvanelteren:
Didn't mean to ping my bad!

cvanelteren:
Tests pass. Just need to doc this properly. Pinged you for review @beckermr

ultraplot/tests/test_integration.py
```diff
@@ -93,22 +156,41 @@ def test_seaborn_swarmplot():
     return fig
 
 
-@pytest.mark.mpl_image_compare
```
beckermr, line 96: Is there a reason we no longer run an image comparison here. I realize it may fail this once, but I think we should keep it for the future.

ultraplot/axes/base.py
```diff
@@ -700,7 +701,53 @@ def __call__(self, ax, renderer):  # noqa: U100
...
+    """
+
+    def set_external(self, value=True):
+        """
+        Set explicit external-mode override for this axes.
+
+        value:
+          - True: force external behavior (defer on-the-fly guides, etc.)
+          - False: force UltraPlot behavior
+          - None: clear override; fallback to auto detection at call sites
```
beckermr: Does `None` work now? I thought we remove that so I think this doc string is out of date.

ultraplot/axes/plot.py
```diff
@@ -1512,25 +1509,6 @@ def _parse_vert(
     return kwargs
 
 
-def _inside_seaborn_call():
```
beckermr, line 1515: I guess this will be a breaking change. We should bump the minor version on the next release.

ultraplot/axes/plot.py
```diff
@@ -2265,9 +2266,32 @@ def _add_error_shading(
...
+                    for _obj in guides._iter_iterables(eobj):
+                        try:
+                            setattr(_obj, "_ultraplot_synthetic", True)
+                            if hasattr(_obj, "set_label"):
+                                _obj.set_label("_ultraplot_shade")
+                        except Exception:
+                            pass
                 eobjs.append(eobj)
 
+        kwargs["distribution"] = distribution
```
beckermr: This looks like a duplicate line.

ultraplot/axes/plot.py
```diff
@@ -4363,7 +4405,13 @@ def _apply_fill(
...
     ):
         """
-        Apply area shading.
+        Apply area shading (fill_between / fill_betweenx) with seaborn helper tagging.
+
+        We tag seaborn-generated confidence interval / helper polygons as synthetic
+        unless the user explicitly supplies a label (label/labels/value/values or
+        shadelabel/fadelabel passed through error shading). Synthetic artists are
+        marked with _ultraplot_synthetic=True and given an underscore label so they
+        are ignored by legend collection.
```
beckermr: This doc string is very helpful! I think "seaborn" -> "externally-created" since now the mechanism is for any external package?

cvanelteren: It will work for others as well (but I am not fully aware of the ins and out of other packages if they internally tag some plots. We can make it more general and then add seaborn as a specific example.

ultraplot/tests/test_integration.py
cvanelteren: Not sure what I was thinking here; will restore it.

ultraplot/axes/plot.py
cvanelteren: I cleaned up the docstring.

ultraplot/axes/base.py
beckermr: We need to address this comment and then this PR should be good to merge.

ultraplot/axes/base.py
```diff
@@ -700,7 +701,52 @@ def __call__(self, ax, renderer):  # noqa: U100
...
+    def set_external(self, value=True):
+        """
+        Set explicit external-mode override for this axes.
+
+        value:
+          - True: force external behavior (defer on-the-fly guides, etc.)
+          - False: force UltraPlot behavior
+        """
+        if value not in (True, False):
+            raise ValueError("set_external expects True, False, or None")
```
beckermr:
```suggestion
            raise ValueError("set_external expects True or False")
```

ultraplot/axes/plot.py
```diff
@@ -8,50 +8,47 @@
...
+import matplotlib.contour as mcontour
 import matplotlib.image as mimage
 import matplotlib.lines as mlines
 import matplotlib.patches as mpatches
-import matplotlib.ticker as mticker
 import matplotlib.pyplot as mplt
-import matplotlib as mpl
-from packaging import version
+import matplotlib.ticker as mticker
+import networkx as nx
```
beckermr:
```suggestion
```

beckermr:
pre-commit.ci autofix

beckermr:
Something is not right with this failed test:

Screenshot 2025-11-18 at 6 56 46 AM

cvanelteren:
hmm thought I fixed that. Will address

cvanelteren:
It was like that already -- not sure why either.

cvanelteren:
Ah wait I see what you mean -- corrected.

ultraplot/axes/plot.py
```diff
@@ -4621,7 +4740,7 @@ def _apply_bar(
         xs, hs, kw = self._parse_1d_args(xs, hs, orientation=orientation, **kw)
         edgefix_kw = _pop_params(kw, self._fix_patch_edges)
         if absolute_width is None:
-            absolute_width = _inside_seaborn_call()
+            absolute_width = False or bar_semantics == "seaborn"
```
beckermr: I'm still confused by this bit. External mode should not have to know anything about seaborn at all. 

The idea with the external mode is when you set an axis to "external", it should behave as if it was a standard matploitlib axis. 

There are two ways to do this that I can see:

1. (what is done in this PR) Pass the ultraplot axis direcly to the other library after marking it as external and modify all of the ultraplot internals to turn itself off. In this case, the code should never need info about the settings of another library (e.g., seaborn). Instead it should put back the equivalent matplotlib defaults and let the other library deal with its own overrides. In the code above, the variable `absolute_width` should be set to the matplotlib default when in external model. There should be no need for a separate bar configuration mode that is used to monkey patch the internals.

2. In external mode, defer all calls to methods on the axis directly to the methods on the parent via something like direct invocation through the class itself (i.e., `getattr(matplotlib.axes.Axes, <name of method>)(self, ...)`. IDK how feasible this approach is. It might be as simple an override of `__getattribute__` when in external mode to defer all attribute access to the parent.

In case 2, the PR might look like adding something like adding

```python
def __getattribute__(self, name):
    orig_attr = object.__getattribute__(self, name)

    # only defer callable methods and they have to be on the parent axis
    if object.__getattribute__(self, "_in_external_context")() and callable(obj_attr):
        mpl_attr = getattr(matplotlib.axes.Axes, name, None)
        if mpl_attr is None:
            return obj_attr
        else:
            # we may need to use functools.partial to insert
            # self at the start of the args here...
            return mpl_attr
    else:
        return object.__getattribute__(self, name)
```

to the class with ~no changes anywhere else. (Be careful with infinite recursion here. See https://stackoverflow.com/questions/371753/how-do-i-implement-getattribute-without-an-infinite-recursion-error.)

Does this make sense to you or did I miss something?

beckermr: Approach 2 might fail if MPL has data containers that are also themselves callable.

PR #408, opened 2025-11-15
fix: change default `reduce_C_function` to `np.sum` for `hexbin`

beckermr:
This PR fixes some unexpected behavior w/ `hexbin`. 

- The matplotlib `hexbin` parameter `C` has been mapped to the parameter `weights` in ultraplot. 
- For the majority (maybe all?) other histogram-like plots in ultraplot (and matplotlib) the `weights` are summed for the points in each bin (along with possibly a global normalization correction). 
- However, the default for `hexbin` is to average the `weights` in each bin. 

This PR changes the default for `hexbin` to sum the `weights` in each bin so that the semantics of how `weights` are used is consistent across plotting functions.

### DIFF
--- a/ultraplot/tests/test_plot.py
+++ b/ultraplot/tests/test_plot.py
@@ -617,3 +617,38 @@ def test_curved_quiver_color_and_cmap(rng, cmap):
     fig, ax = uplt.subplots()
     ax.curved_quiver(X, Y, U, V, color=color, cmap=cmap)
     return fig
+
+
+def test_histogram_norms():
+    """
+    Check that all histograms-like plotting functions
+    use the sum of the weights.
+    """
+    rng = np.random.default_rng(seed=100)
+    x, y = rng.normal(size=(2, 100))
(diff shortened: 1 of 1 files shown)


cvanelteren:
Can you add a test for this to run over the bin plots and ensure we have the same behavior?

beckermr:
Any suggestions for how to write the test?

cvanelteren:
I think we can do a simple check where take all hist related functions and check that the patches contain the sum; doesn't have to be super elaborate but just a safeguard for if this changes on mpl or our end in the future.

We could also check if what function is called with patches to see what the function put as input and check that this is a sum to ensure we are consistent across the board. 

We may also need some smoke tests, like does passing a cumstom func properly propagate and set the values accordingly. This would not necessarily need a visual check.

beckermr:
ok @cvanelteren this is ready for another look!

cvanelteren:
Great thanks!

PR #413, opened 2025-11-16
Hotfix: bar labels cause limit to reset for unaffected axis.

### LINKED ISSUE #412: Limits are reset when calling `bar_labels`
> When I tried to add labels for each bar, it strangely only drew a single bar. Does this mean I have to add the labels manually myself?
> 
> ```
> import numpy as np
> import pandas as pd
> import ultraplot as uplt
> 
> x = np.arange(5)
> y = [10, 15, 7, 12, 9]
> data = pd.DataFrame(y, index=x)
> 
> cycle = uplt.Cycle("RdYlGn", 5)
> 
> fig, ax = uplt.subplots(journal="nat1")
> for idx, row in data.iterrows():
>     ax.bar(idx, row.iloc[0], cycle=cycle, edgecolor="black", bar_labels=True)
> ```
> 
> Image 

 _Originally posted by @lwq-star in [#409](https://github.com/Ultraplot/UltraPlot/issues/409#issuecomment-3538185127)_

cvanelteren:
Closes #412 

This PR addresses an issue where limits are reset when adding bar labels in a loop (see #412). The issue occurs because the `bar_labels` expects that the bar are in a container which will set the range on all the bars.  This PR fixes this by considering `dataLim` directly rather than the `BarContainer`.

### DIFF
--- a/ultraplot/axes/base.py
+++ b/ultraplot/axes/base.py
@@ -2626,9 +2626,19 @@ class Axes(maxes.Axes):
                     setattr(sibling, f"_share{which}", None)
                 this_ax = getattr(self, f"{which}axis")
                 sib_ax = getattr(sibling, f"{which}axis")
-                # Reset formatters
-                this_ax.major = copy.deepcopy(this_ax.major)
-                this_ax.minor = copy.deepcopy(this_ax.minor)
+                # Reset formatters by creating new Ticker objects.
+                # A deepcopy can trigger redraws.
+                new_major = maxis.Ticker()
+                if this_ax.major:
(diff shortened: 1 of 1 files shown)


cvanelteren:
Hmm had a look at this for a while and I am not really understanding how the xlimits are set. For `BarContainers` this goes well but not for iterative calls -- will return to this. Putting it to draft.

cvanelteren:
The difference on the plot is minor some shifting but everything else looks ok.

cvanelteren:
Done

beckermr (approved):

cvanelteren:
Great. I will draft a release and bump the number.

PR #427, opened 2025-12-12
Swap visual tests with hash comparison

### LINKED ISSUE #424: [FIX/Discussion] Slow unittests
We need to address the growing number of tests and their increasing run times. As most our tests are visual, we should explore other methods for testing the fidelity of the codebase. I see a few potential options.

## Hashing (preferred)
pytest-mpl allows for testing hashes rather than images directly. This has two direct benefits. For one, it is fast, and second, we don't have to store the images -- just the hashes.

## Reduce number of visual tests (not-preferred)
We should perhaps curate the number of tests to specifically test some but not all properties. I don't really now which ones would be best, but I can imagine a few essential ones and tooling the rest to be logical checks only.

cvanelteren:
Closes #424 
As we expand our tests, the runners are running longer and longer. I propose that we move from visual tests to hashes on the runners.  The idea would be that we can see what tests fails such that they can be testsed locally to see how they differ visually.

We should move also to nox in case the tests are visually off in a specific combination of python and matplotlib.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -11,6 +11,6 @@ ci:
 
 repos:
   - repo: https://github.com/psf/black-pre-commit-mirror
-    rev: 25.11.0
+    rev: 25.12.0
     hooks:
       - id: black
--- a/ultraplot/axes/base.py
+++ b/ultraplot/axes/base.py
@@ -354,7 +354,9 @@ abcbbox, titlebbox : bool, default: :rc:`abc.bbox` and :rc:`title.bbox`
     inside the axes. This can help them stand out on top of artists plotted
     inside the axes.
 abcpad : float or unit-spec, default: :rc:`abc.pad`
-    The padding for the inner and outer titles and a-b-c labels.
+    Horizontal offset to shift the a-b-c label position. Positive values move
+    the label right, negative values move it left. This is separate from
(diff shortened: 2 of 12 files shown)


cvanelteren:
Would love your input here @beckermr!

cvanelteren:
Caching is a good idea (see latest commit). I was under the impression that the hash comparison was also faster. The generation on the baseline would be slow, but the hash check is fast.

cvanelteren:
I will do some speed tests hang on

beckermr:
The slowest part of this is making the plots. Comparing them via hashes or directly through differencing the images will have a time difference but the change will be small relative to everything else.

cvanelteren:
Ok yeah great. Another would be to use `pytest-xdist` which recently got updated to allow to work with `pytest-mpl`. The latter will give a minor boost as I think there are only 2 cores per runner.

cvanelteren:
Note on PR history: initial attempts used pytest‑mpl hash libraries. We’ve switched to image baseline comparisons instead.
Current behavior: baselines are generated once per PR+matrix entry and cached (keyed by PR number, Python, Matplotlib), then reused for subsequent runs. Hash library generation/checks are removed and comparisons now use `--mpl-baseline-path`.

cvanelteren:
With two runners the tests run significantly faster -- but not sure on the racing condition.

PR #429, opened 2025-12-13
Fix: Boundary labels now visible when setting lonlim/latlim

### LINKED ISSUE #428: lon, lat labels disappear after ax.format
I like to write a function make_masemap() to obtain the base map, and then draw pcolormesh based on it. However, I have discovered two bugs. On the one hand, when ax. format() is after fig, ax=make_masemap(), the latitude and longitude labels will disappear. I suspect this is caused by ax. format(). On the other hand, I set ax.grid (False), but the gridlines still exist. Below is my code:

import cartopy.crs as ccrs
from cartopy.mpl.ticker import LongitudeFormatter, LatitudeFormatter
import ultraplot as uplt

```
def make_basemap():
    box = [160, 170, 10, 20]

    fig, ax = uplt.subplots(nrows=1, ncols=1, proj=ccrs.PlateCarree(140.7), share=0)

    lon_formatter = LongitudeFormatter(number_format='g', degree_symbol='°', dateline_direction_label=True)
    lat_formatter = LatitudeFormatter(number_format='g', degree_symbol='°')
    ax.set_xticks([box[0], (box[0] + box[1]) / 2, box[1]], crs=ccrs.PlateCarree())
    ax.set_yticks([box[2], (box[2] + box[3]) / 2, box[3]], crs=ccrs.PlateCarree())
    ax.xaxis.set_major_formatter(lon_formatter)
    ax.yaxis.set_major_formatter(lat_formatter)
    ax.set_extent(box, crs=ccrs.PlateCarree())
    ax.tick_params(axis="both", length=2)
    ax.coastlines("110m")
    ax.grid(False)

    return fig, ax


def main():
    # I like to use this to initial basemap, and plot image after this line.
    fig, ax = make_basemap()

    # The latitude and longitude labels will disappear when
    # ax.format() is run after ax.xaxis.set_major_formatter(lon_formatter)
    ax.format(abc='A')

    fig.savefig(r"D:\bbbb.png", dpi=500)
```

Image

cvanelteren:
Closes #428 

Fixed issue where the outermost gridline labels would be hidden when setting `lonlim` or `latlim` on geographic axes. The problem had two causes: (1) cartopy's gridliner `xlim`/`ylim` were not properly synchronized with the view intervals - only `ylim` was set and to an incorrect value, and (2) the extent was being shifted rather than expanded, causing the leftmost boundary to exclude gridlines at exactly the limit value. The fix ensures both `xlim` and `ylim` are set to the actual view intervals with appropriate expansion, and that extent bounds are expanded symmetrically. Added comprehensive unit tests to verify boundary labels are created and visible for positive/negative longitudes and various latitude ranges.

### DIFF
--- a/ultraplot/tests/test_geographic.py
+++ b/ultraplot/tests/test_geographic.py
@@ -460,7 +460,10 @@ def test_sharing_geo_limits():
     after_lat = ax[1]._lataxis.get_view_interval()
 
     # We are sharing y which is the latitude axis
-    assert all([np.allclose(i, j) for i, j in zip(expectation["latlim"], after_lat)])
+    # Account for small epsilon expansion in extent (0.5 degrees per side)
+    assert all(
+        [np.allclose(i, j, atol=1.0) for i, j in zip(expectation["latlim"], after_lat)]
+    )
     # We are not sharing longitude yet
     assert all(
         [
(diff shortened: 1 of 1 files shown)


cvanelteren:
A bunch of tests will fail on geo, but they look visually ok in my opinion.

PR #432, opened 2025-12-16
Fix GeoAxes panel alignment with aspect-constrained projections

### LINKED ISSUE #431: Panel not reshaping to main plot for GeoAxes with Cartopy backend
I don't know if you saw this issue, but when I drew the image on the right, the first thing that came to my mind was the` panel()` method in the example code. But it seems that the parameter 'high' is derived from the main axes. Although the top panel did a good job in completing the task in this image, I believe there may also be corresponding issues in certain specific images. I don't know how to solve this problem, although I can use the `fig.add_axes()` method to implement it, I believe someone will encounter corresponding problems and questions.
Image

cvanelteren:
Closes #431 

This PR fixes the issue where panels added to GeoAxes with large padding values do not properly align with the map boundaries. The problem occurs because cartopy's `apply_aspect()` method shrinks the main axes to maintain the projection's aspect ratio, but panels (created in separate gridspec slots) remain at their original positions, creating unwanted gaps. The fix introduces a new` _adjust_panel_positions()` method that is called after `apply_aspect()` in both `draw()` and `get_tightbbox()` to dynamically reposition all panels so they properly flank the visible map boundaries while maintaining the original spacing gaps, ensuring proper alignment regardless of the projection's aspect ratio constraints.

### DIFF
--- a/ultraplot/axes/geo.py
+++ b/ultraplot/axes/geo.py
@@ -674,6 +674,11 @@ class GeoAxes(shared._SharedAxes, plot.PlotAxes):
     def _apply_aspect_and_adjust_panels(self, *, tol=1e-9):
         """
         Apply aspect and then align panels to the adjusted axes box.
+
+        Notes
+        -----
+        Cartopy and basemap use different tolerances when detecting whether
+        apply_aspect() actually changed the axes position.
         """
         self.apply_aspect()
         self._adjust_panel_positions(tol=tol)
@@ -1427,6 +1432,7 @@ class _CartopyAxes(GeoAxes, _GeoAxes):
     _name = "cartopy"
     _name_aliases = ("geo", "geographic")  # default 'geographic' axes
     _proj_class = Projection
+    _PANEL_TOL = 1e-9
     _proj_north = (
         pproj.NorthPolarStereo,
         pproj.NorthPolarGnomonic,
@@ -1971,7 +1977,7 @@ class _CartopyAxes(GeoAxes, _GeoAxes):
         after the main axes has applied its aspect but before the panel axes are drawn.
         """
         super().draw(renderer, *args, **kwargs)
-        self._adjust_panel_positions()
+        self._adjust_panel_positions(tol=self._PANEL_TOL)
 
     def get_tightbbox(self, renderer, *args, **kwargs):
         # Perform extra post-processing steps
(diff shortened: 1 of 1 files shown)


cvanelteren:
Will need to add some tests but the logic can already be checked.

ultraplot/tests/test_geographic.py
```diff
@@ -1535,7 +1535,13 @@ def test_boundary_labels_negative_longitude():
         grid=False,
     )
     fig.canvas.draw()
-    _check_boundary_labels(ax[0], ["120°W", "90°W", "60°W"], ["20°N", "35°N", "50°N"])
+    # Note: Cartopy hides the boundary label at 20°N due to it being exactly at the limit
```
cvanelteren, line 1538: editor changed this -- can ignore

ultraplot/tests/test_geographic.py
```diff
@@ -1473,26 +1473,26 @@ def _check_boundary_labels(ax, expected_lon_labels, expected_lat_labels):
...
-    # Check latitude labels
+    # Check latitude labels - only verify the visible ones match expected
     lat_texts = [
         label.get_text() for label in gl.left_label_artists if label.get_visible()
     ]
-    assert len(gl.left_label_artists) == len(expected_lat_labels), (
-        f"Should have {len(expected_lat_labels)} latitude labels, "
-        f"got {len(gl.left_label_artists)}"
+    assert len(lat_texts) == len(expected_lat_labels), (
+        f"Should have {len(expected_lat_labels)} visible latitude labels, "
```
cvanelteren, line 1494: same here can be ignored

ultraplot/tests/test_geographic.py
```diff
@@ -1473,26 +1473,26 @@ def _check_boundary_labels(ax, expected_lon_labels, expected_lat_labels):
...
 
-    # Check longitude labels
+    # Check longitude labels - only verify the visible ones match expected
     lon_texts = [
         label.get_text() for label in gl.bottom_label_artists if label.get_visible()
     ]
-    assert len(gl.bottom_label_artists) == len(expected_lon_labels), (
-        f"Should have {len(expected_lon_labels)} longitude labels, "
-        f"got {len(gl.bottom_label_artists)}"
+    assert len(lon_texts) == len(expected_lon_labels), (
```
cvanelteren, line 1480: see comments below

cvanelteren:
I will be away next week so pushing a few changes across the major PRS in case you wanted to have a look already @beckermr 

Otherwise don't bother and we can pick it up in the new year.

beckermr (approved):
There is a lot of repeated code here. Not sure if we care. Otherwise lgtm.

cvanelteren:
Thanks!

PR #433, opened 2025-12-16
Add Copernicus Publications figure standard widths

Holmgren825:
Adds standard 1- and 2-column figure widths for journals from [Copernicus Publications](https://publications.copernicus.org/), e.g. The Cryosphere, Weather and Climate Dynamics, and Natural Hazards and Earth System Sciences.


cvanelteren:
Thanks!

PR #442, opened 2025-12-31
Avoid title overlap with abc labels

cvanelteren:
This PR avoids title overlap with a‑b‑c labels when both are placed in the same location. During draw, it measures the rendered widths of the two texts and, if needed, scales the title font down to fit within the available horizontal space for left, right, and centered alignments. The change is localized to `_update_title_position`, so it only affects shared‑location titles and leaves other title behavior unchanged.

### DIFF
--- a/ultraplot/axes/base.py
+++ b/ultraplot/axes/base.py
@@ -354,7 +354,9 @@ abcbbox, titlebbox : bool, default: :rc:`abc.bbox` and :rc:`title.bbox`
     inside the axes. This can help them stand out on top of artists plotted
     inside the axes.
 abcpad : float or unit-spec, default: :rc:`abc.pad`
-    The padding for the inner and outer titles and a-b-c labels.
+    Horizontal offset to shift the a-b-c label position. Positive values move
+    the label right, negative values move it left. This is separate from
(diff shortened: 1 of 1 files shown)


cvanelteren:
There is no real issue that closes this, more an issue I run into sometimes where the titles will overlap with the labels.

ultraplot/axes/base.py
```diff
@@ -3042,7 +3046,9 @@ def _update_title_position(self, renderer):
         # Offset title away from a-b-c label
         atext, ttext = aobj.get_text(), tobj.get_text()
         awidth = twidth = 0
-        pad = (abcpad / 72) / self._get_size_inches()[0]
+        width_inches = self._get_size_inches()[0]
+        pad = (abcpad / 72) / width_inches
+        abc_pad = (self._abc_pad / 72) / width_inches
```
beckermr: Please 

1. leave a comment on the difference between abcpad, sel._abc_pad, pad, and abc_pad. 

2. Rename the variables to better match their usage and meaning. 

I get the unit conversion, but I don't understand what each variable means, what they are used for, and where they come from.

cvanelteren: Done!

PR #443, opened 2025-12-31
Fix dualx alignment on log axes

### LINKED ISSUE #441: dualx not correctly shown
There is a `dualx` funcitonality that could be very useful in labelling the wavelength and wavenumber simultaneously.  Here is an example from a paper:

Image

The x-axis at the bottom shows the wavenumber (cycle per km) while the one at the top shows the wavelength (km), as `wavenumber = 1 / wavelength`.

I tried this with the following simple codes:
```python
import ultraplot as uplt
import numpy as np
import xarray as xr

data = xr.DataArray(np.random.randn(100)+2, dims=['wn'], coords={'wn':np.logspace(-2, 2, 100)})

fig, axes = uplt.subplots(figsize=(5,4))

ax = axes[0]
ax.plot(data)
ax.set_xscale('log')
ax.set_yscale('log')
ax.format(xformatter='log')
ax.set_xlabel('wavenumber (cycle per km)')
ax.dualx(
    lambda x: 1 / x,
    label="length scale (km)",
    formatter="log",
    grid=True,
    color='k',
    gridcolor='k',
)
```

which shows 

Image

It is a bit strange to me that the two x-axes do not align exactly.  Just want to know how to do this correctly.  Thanks very much.

cvanelteren:
This fixes dual‑axis alignment on log scales by applying the dual transform in data space before the parent scale, rather than after it. Previously, the function saw log‑transformed values, which produced inconsistent limits/ticks and could hide the secondary axis entirely. A regression test now exercises `dualx` on a log x‑axis and asserts the transformed tick positions remain finite.
Closes #441

### DIFF
--- a/.github/workflows/build-ultraplot.yml
+++ b/.github/workflows/build-ultraplot.yml
@@ -98,7 +98,7 @@ jobs:
       # Return the html output of the comparison even if failed
       - name: Upload comparison failures
         if: always()
-        uses: actions/upload-artifact@v5
+        uses: actions/upload-artifact@v6
         with:
           name: failed-comparisons-${{ inputs.python-version }}-${{ inputs.matplotlib-version }}-${{ github.sha }}
           path: results/*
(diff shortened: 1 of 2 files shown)


cvanelteren:
@beckermr I merged a few PRs as they are minor and there are others that require proper review -- wanted to unburden you a bit.

PR #465, opened 2026-01-13
Add gallery

### LINKED ISSUE #456: Feature request: Add gallery to docs
We should show case the power of UlraPlot for casual browsers. I think something similar to a gallery overview that should entice browsers and viewer and perhaps inspire to make new complicated plots.

The plots should focus:
- ...on the core value that we add in UltraPlot
- ...unique features that UltraPlot adds (layout sharing, legends, abc etc)
- ...[?]

I was thinking over leveraging the existing jupytext and linking them immediately, but perhaps a more curated list is better.

cvanelteren:
Add gallery plots of a curated selection of subplots

Closes #456 
image

### DIFF
--- a/docs/make.bat
+++ b/docs/make.bat
@@ -1,0 +1,41 @@
+@echo off
+REM Minimal make.bat for Sphinx documentation
+REM You can set these variables from the command line.
+
+REM Set default variables
+set SPHINXOPTS=
+set SPHINXBUILD=sphinx-build
+set SPHINXPROJ=UltraPlot
+set SOURCEDIR=.
+set BUILDDIR=_build
+
+REM Check if no arguments were provided (show help)
+if "%1"=="" goto help
+
+REM Route to the appropriate target
+if "%1"=="help" goto help
+if "%1"=="clean" goto clean
+
+REM Catch-all target: route all unknown targets to Sphinx
+goto catchall
+
+
+:help
+REM Put it first so that "make" without argument is like "make help".
+%SPHINXBUILD% -M help "%SOURCEDIR%" "%BUILDDIR%" %SPHINXOPTS%
+goto :eof
+
+
+:clean
+REM Make clean ignore .git folder
+REM The /q doesn't raise error when files/folders not found
+if exist api\ rmdir /s /q api\
+if exist "%BUILDDIR%\html\" rmdir /s /q "%BUILDDIR%\html\"
+if exist "%BUILDDIR%\doctrees\" rmdir /s /q "%BUILDDIR%\doctrees\"
+goto :eof
+
+
+:catchall
+REM Route target to Sphinx using the "make mode" option
+%SPHINXBUILD% -M %1 "%SOURCEDIR%" "%BUILDDIR%" %SPHINXOPTS%
+goto :eof


cvanelteren:
- [x] Need to clean up the cylinder plot to mask the inside of the cylinder as it currently does not make any sense.
- [x] Need to fix the headers leaking from the H2 and onwards in the gallery in the main toc on the left

cvanelteren:
I added a gallery to the docs and did some behind the scenes fixes.

Just have a look at https://ultraplot--465.org.readthedocs.build/en/465/gallery/index.html
Can you guys give a go ahead for this @gepcel @beckermr ?

cvanelteren:
There is a bug in the shared labels but that is unrelated to the gallery -- I am trying to fix this now.

docs/examples/geo/03_projections_features.py
gepcel, line 1: There's a warning message about `GeoAxes can only be shared for rectilinear projections`. Are we keeping the warn here?

cvanelteren: Ah fair point, would need to disable the sharing for that subplot -- thanks!

docs/examples/plot_types/03_lollipop.py
gepcel, line 1: There's also two warning message `Got conflicting or duplicate keyword arguments: {'colors': 'C1', 'color': (0.0, 0.4470588235294118, 0.6980392156862745, 1.0)}`

docs/examples/plot_types/01_curved_quiver.py
gepcel, line 1: This chart doesn't look very aesthetic. Don't you think the lines and arrows are a bit too thick?

docs/examples/plot_types/03_lollipop.py
cvanelteren: Thanks -- this should be fixed but I'd not related to the PR

docs/examples/plot_types/01_curved_quiver.py
cvanelteren: like this?

cvanelteren: or this 
image

gepcel (approved):

docs/examples/geo/01_robin_tracks.py
```diff
@@ -0,0 +1,37 @@
...
+lat_a = 25 * np.sin(np.deg2rad(lon * 1.5))
+lat_b = -15 * np.cos(np.deg2rad(lon * 1.2))
+
+fig, ax = uplt.subplots(proj="robin", proj_kw={"lon0": 0}, refwidth=4)
+ax.plot(lon, lat_a, transform=ccrs.PlateCarree(), lw=2, label="Track A")
+ax.plot(lon, lat_b, transform=ccrs.PlateCarree(), lw=2, label="Track B")
+ax.scatter([-140, -40, 60, 150], [10, -20, 30, -5], transform=ccrs.PlateCarree())
+
+ax.format(title="Global trajectories", lonlines=60, latlines=30)
+ax.legend(loc="lower left", frame=False)
```
gepcel: `loc="lower center"` would be better.

cvanelteren: Thanks implemented.

PR #469, opened 2026-01-15
Fix legend span inference with panels

### LINKED ISSUE #468: Gridspec indexing not matching with multi-legends
> Related to #461. When plot multiple legends, rows and cols of loc_axs may result in wrong value [here ](https://github.com/Ultraplot/UltraPlot/blob/8edd69431a4d048e0aeb4c3771a5a1a8144a5850/ultraplot/figure.py#L2842) due to hiden panel. I've tried many examples, sometimes legend will be placed in wrong places, and sometimes it will raise errors. e.g:
> 
> ```python
> import ultraplot as pplt
> fig,axs=pplt.subplots(ncols=3, nrows=3,refwidth='3cm')
> axs.plot([1,2], [1,2], label='line')
> fig.legend(ref=axs[0],loc='b') //comment this line, error will go away
> fig.legend(ref=axs[:, 1],loc='r')
> ``` 

 _Originally posted by @gepcel in [#459](https://github.com/Ultraplot/UltraPlot/issues/459#issuecomment-3752596852)_

cvanelteren:
Legend span inference used panel-inflated indices after prior legends added panel rows/cols, yielding invalid gridspec indices for list refs.

Decode subplot indices to non-panel grid before computing span and add regression tests for multi-legend ordering.

Closes #468

### DIFF
--- a/.github/workflows/build-ultraplot.yml
+++ b/.github/workflows/build-ultraplot.yml
@@ -113,7 +113,7 @@ jobs:
 
           mkdir -p results
           python -c "import ultraplot as plt; plt.config.Configurator()._save_yaml('ultraplot.yml')"
-          pytest -x -W ignore -n auto\
+          pytest -x -W ignore \
             --mpl \
             --mpl-baseline-path=./ultraplot/tests/baseline \
             --mpl-results-path=./results/ \


cvanelteren:
Docs error will be fixed on #465

PR #490, opened 2026-01-21
Fix: Move locator back to top level

### LINKED ISSUE #489: After this commit I had to change from `uplt.Locator` to `uplt.constructor.Locator`, to avoid a `AttributeError: module 'ultraplot' has no attribute 'Locator'` error.
After this commit I had to change from `uplt.Locator` to `uplt.constructor.Locator`, to avoid a `AttributeError: module 'ultraplot' has no attribute 'Locator'` error.
Is this the expected behaviour?

_Originally posted by @pgf in https://github.com/Ultraplot/UltraPlot/issues/439#issuecomment-3772513414_

cvanelteren:
Closes #489


cvanelteren:
May need to add an automatic detector for `__all__` listings in the future.

PR #430, opened 2025-12-14
Add UltraLayout: Advanced constraint-based positioning for non-orthogonal subplot arrangements

cvanelteren:
Closes #297,#270

This PR introduces **UltraLayout**, a novel constraint-based layout system using kiwisolver that enables aesthetically pleasing positioning for non-orthogonal subplot arrangements (e.g., `[[1, 1, 2, 2], [0, 3, 3, 0]]` where subplot 3 should be centered between subplots 1 and 2). The existing gridspec works great for orthogonal layouts but fails to produce visually balanced results when subplots span multiple rows/columns in non-aligned ways. UltraLayout automatically detects non-orthogonal layouts, applies constraint satisfaction to compute optimal positions that respect spacing, ratios, and aesthetic balance (e.g., centering subplots between neighbors), and gracefully falls back to the standard grid layout if kiwisolver is not installed or for orthogonal arrangements. This enhancement is fully backward compatible, adds zero overhead for standard grid layouts, includes comprehensive test coverage in `ultraplot/tests/test_ultralayout.py`, and provides a foundation for more sophisticated layout constraints in the future.

Make subplot param probes backend-safe: `GridSpec.get_subplot_params` now returns a read-only snapshot of our margins/spaces and `locally_modified_subplot_params` returns False. This keeps Positron/other backends from erroring while preserving our restriction on mutating Matplotlib’s subplot params.

PR #494, opened 2026-01-22
Fix/animation noise

### LINKED ISSUE #488: Points to note when creating animations using `ultraplot`
Hello,

I really enjoy using `ultraplot` to create publication-quality plots with simple and straightforward settings.
Likewise, when I applied it to animation creation, I ran into a couple of weird problems.
So, I would like to keep them as notes here.

### 1. Need to set `matplotlib.use("Agg")`
When creating an animation with Python scripts, it is necessary to set the backend to `Agg`; otherwise, the resulting video contains totally horizontal noise.

### 2. Set `tight=False`
Set `tight=False` to avoid changing layout during animation. This change might be triggered when using the `format` method in the `update` function used for [`FuncAnimation`](https://matplotlib.org/stable/api/_as_gen/matplotlib.animation.FuncAnimation.html#matplotlib.animation.FuncAnimation).

### 3. Axis labels (`toplabels`, `leftlabels`) are outside the area
Labels like `toplabels` and `leftlabels` seem to be placed outside the figure rendering area.
This might be due to the tight layout, but I haven't found a solution yet.

This issue may be irrelevant to those who only want to create static images, but I would be happy if it is useful to someone.

cvanelteren:
This PR adds an `animation.force_draw_idle` rcParam (default on) and uses it to gate a targeted FuncAnimation hook that swaps `draw_idle` to full draws on interactive canvases, reducing backend redraw artifacts without affecting headless renders, while keeping the earlier layout convergence work intact

Closes #488 

Drafting for now as I am not happy with it yet.

### DIFF
--- a/.github/workflows/build-ultraplot.yml
+++ b/.github/workflows/build-ultraplot.yml
@@ -54,6 +54,8 @@ jobs:
   compare-baseline:
     name: Compare baseline Python ${{ inputs.python-version }} with MPL ${{ inputs.matplotlib-version }}
     runs-on: ubuntu-latest
+    env:
+      IS_PR: ${{ github.event_name == 'pull_request' }}
     defaults:
       run:
         shell: bash -el {0}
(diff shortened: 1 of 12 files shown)


cvanelteren:
Well we would have to replace how the layout is drawn -- I believe the issue is coming from the fact that ultraplot does a lot of layout correction on draw which causes an issue with FuncAnimation. Another option would be to provide a `FuncAnimation` ourselves that patches this internally. Since we are replacing the layout algorithm in #430 we could try and fix it with a patch on that.

beckermr:
I see. The less patching we do, the better. I think given the nature of the fix, we should try and solve this cleanly with the new layout handling.

PR #495, opened 2026-01-23
Feature: integrate pycirclize into UltraPlot

### LINKED ISSUE #472: Feature request: Add Chord diagram

cvanelteren:
This PR will integrated pycirclize into UltraPlot exposing chord diagrams, radar plots, and phylogeny plot. It wraps PyCirclize into an UltraPlot flavored manner adding the convenience of normal plotting while harnessing the strenghts of this well developed package.

Putting this in draft for now but it seems to work nicely with main

image

Closes #472

### DIFF
--- a/ultraplot/axes/geo.py
+++ b/ultraplot/axes/geo.py
@@ -2139,9 +2139,6 @@ class GeoAxes(shared._SharedAxes, plot.PlotAxes):
         for size, which in zip(sizes, ["major", "minor"]):
             params.update({"length": size})
             params.pop("grid_alpha", None)
-            # Avoid overriding gridliner label toggles via tick_params defaults.
-            for key in ("labeltop", "labelbottom", "labelleft", "labelright"):
-                params.pop(key, None)
             self.tick_params(
                 axis=x_or_y,
                 which=which,
@@ -2559,7 +2556,6 @@ class _CartopyAxes(GeoAxes, _GeoAxes):
         # (x, y) coordinate pairs (each corner), so something like (-180, 180, -90, 90)
         # will result in *line*, causing error! We correct this here.
         eps_small = 1e-10  # bug with full -180, 180 range when lon_0 != 0
-        eps_label = 0.5  # larger epsilon to ensure boundary labels are included
         lon0 = self._get_lon0()
         proj = type(self.projection).__name__
         north = isinstance(self.projection, self._proj_north)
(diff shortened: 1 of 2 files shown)


pyproject.toml
```diff
@@ -38,6 +38,10 @@ dependencies= [
 ]
 dynamic = ["version"]
 
+[project.optional-dependencies]
+circos = ["pycirclize>=1.10.1"]
+all = ["pycirclize>=1.10.1"]
```
cvanelteren: Need to work on this a bit -- do we want to provide a minimal install and all bells and whizzles? I guess for people (like me) that does not use geo this may be nice to reduce the disk space

ultraplot/tests/test_plot.py
```diff
@@ -733,6 +733,139 @@ def test_curved_quiver_color_and_cmap(rng, cmap):
...
+    uplt.close(fig)
+
+
+def test_circos_bed_smoke(tmp_path):
+    """Smoke test for BED-based circlize wrapper."""
+    try:
+        from ultraplot.axes.plot_types.circlize import _import_pycirclize
+
+        _import_pycirclize()
+    except ImportError:
```
cvanelteren: I think this is overkill and too specific to how this was developed

cvanelteren: hmm on second thought maybe this is ok

pyproject.toml
cvanelteren: Will change this to have default install everything and instead add a minimal installer

cvanelteren:
Opening this up for review. The new layout is still not stabilizing enough or the CI are too strict. The failing tests are on differences that are not caused by the code introduced here (as it is additive).

ultraplot/axes/plot.py
```diff
@@ -299,6 +299,185 @@
...
+circular : bool, optional
+    Whether to draw circular grid lines.
+cmap : str or dict, optional
+    Colormap name or row-name-to-color mapping. If omitted, UltraPlot's
+    color cycle is used.
+show_grid_label : bool, optional
+    Whether to show radial grid labels.
+grid_interval_ratio : float or None, optional
+    Grid interval ratio (0 to 1).
+grid_line_kws, grid_label_kws : dict-like, optional
```
gepcel: I'd like to bring this to your attention. In ultraplot, we use a lot of _kw, like abc_kw, title_kw, rc_kw, norm_kw, which is efficient, unified, and reduces keystrokes. For these kws here, should we maintain consistency with the ultraplot style or with the pyCirclize style?

cvanelteren: Think you are right

cvanelteren: I will also move the defaults to RC while I am at it

gepcel: I really admire your efficiency. Since you're already tinkering with defaults and RC, when creating plots, there's a warning message: `UltraPlotWarning: Tick label sharing not implemented for <class 'ultraplot.axes.polar.PolarAxes'> subplots.` It's fine with me, and no big deal. But if you'd like take a look, maybe turn tick label sharing default to off if it's not support, and warn if users manually turn it on? If it will not complicate other thing.

cvanelteren: That warning is correct, to avoid it explicitly set sharing to 0. Sharing polar axes does not make sense for multiple subplots -- we may want to add a surpress flag in case it gets annoying

gepcel (approved):

PR #498, opened 2026-01-23
Fix font lazy load

### LINKED ISSUE #497: Math font not displaying properly?
Hello!

Math text font does not appear to be displayed correctly since v1.71.0, showing the following warning in log:
```bash
findfont: Font family ['regular'] not found. Falling back to DejaVu Sans.
findfont: Font family ['regular'] not found. Falling back to DejaVu Sans.
findfont: Font family ['regular'] not found. Falling back to DejaVu Sans.
```

This log can be seen in [the font documents](https://ultraplot.readthedocs.io/en/latest/fonts.html#Math-text-fonts) as well, where all math fonts are changed to DejaVu Sans.

This seems to be a regression issue.

cvanelteren:
This PR loads font at the proper stage so it is not skipped and created before a MathParser us isued.

Closes #497

### DIFF
--- a/ultraplot/__init__.py
+++ b/ultraplot/__init__.py
@@ -67,6 +67,9 @@ def _setup():
             register_cycles,
             register_fonts,
         )
+        from .internals import (
+            fonts as _fonts,  # noqa: F401 - ensure mathtext override is active
+        )
         from .internals import rcsetup, warnings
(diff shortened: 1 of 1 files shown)


cvanelteren:
Overriding failure here.

PR #504, opened 2026-01-24
Fix idle draw animation

### LINKED ISSUE #488: Points to note when creating animations using `ultraplot`

cvanelteren:
Updated the animation draw pipeline to avoid repeated layout recomputation, which was causing layout jitter during animations. Auto‑layout now only runs when the layout is actually dirty or on the initial draw, and layout dirtiness is tracked across format calls, subplot/panel creation, and user‑driven size changes. Added a regression test to ensure draw_idle no longer retriggers layout. CI has been hardened: selected‑test mode now falls back to the full suite when nodeids are missing (exit code 4/5), baseline generation does the same on base commits, and coverage is collected in a single‑process pass to avoid xdist worker gaps. Finally, CI logs were made quieter with short tracebacks and warnings suppressed so failures are easier to spot.

Closes #488

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -34,6 +34,7 @@ classifiers = [
 dependencies= [
     "numpy>=1.26.0",
     "matplotlib>=3.9,<3.11",
+    "pycirclize>=1.10.1",
     "typing-extensions; python_version < '3.12'",
 ]
 dynamic = ["version"]
--- a/INSTALL.rst
+++ b/INSTALL.rst
@@ -10,6 +10,14 @@ with ``pip`` or ``conda`` as follows:
    pip install ultraplot
    conda install -c conda-forge ultraplot
 
+The default install includes optional features (for example, pyCirclize-based plots).
+For a minimal install, use ``--no-deps`` and install the core requirements:
+
+.. code-block:: bash
+
+   pip install ultraplot --no-deps
+   pip install -r requirements-minimal.txt
+
 Likewise, an existing installation of ultraplot can be upgraded to the latest version with:
 
 .. code-block:: bash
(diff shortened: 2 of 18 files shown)


pyproject.toml
```diff
@@ -58,6 +58,9 @@ ignore = ["I001", "I002", "I003", "I004"]
 [tool.basedpyright]
 exclude = ["**/*.ipynb"]
 
+[tool.pytest.ini_options]
+filterwarnings = [
+    "ignore:'resetCache' deprecated - use 'reset_cache':DeprecationWarning:matplotlib._fontconfig_pattern",
```
beckermr, line 64:
```suggestion
    "ignore:'resetCache' deprecated - use 'reset_cache':DeprecationWarning:matplotlib._fontconfig_pattern",
]
```

cvanelteren:
This PR is not fixed yet. There are some funky things still happening on this PR.

pyproject.toml
cvanelteren: For some reason I cannot accept this change on the web but I don't see what this doesexactly given the code -- other than the extra lien

PR #515, opened 2026-01-27
Separate docs dependencies from user dependencies

cvanelteren:
This PR takes steps to separate the dev environment from the user environment. Currently we are bundling the dependencies to build into the end-users environment. This results in unncessary bloat if one does not want to build docs locally.

In addition, it add a new doc dependency (sphinx-sitemap) to aid in the indexing the RTD page such that it can be indexed easier by search engines.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -64,7 +64,7 @@ docs = [
     "jupytext",
     "lxml-html-clean",
     "m2r2",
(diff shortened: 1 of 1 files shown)


cvanelteren:
Note we can also do this for pytest in this or a separate PR.

PR #520, opened 2026-01-28
Keep gridline labels when updating ticklen

cvanelteren:
Closes #520

This PR prevents ticklabels from being removed by setting the tick properties.


gepcel:
typos: should close 519, not 520

Issue #524, opened 2026-01-29
Fix CI issue


cvanelteren:
I wanted to make the CI more streamlined and snappy but my recent merges may have worsened it. I need to prioritize fixing this to ensure we can have visual fidelity across new additions or changes.

cvanelteren:
A solution would be to shift to perceptual hashes, see https://github.com/matplotlib/pytest-mpl/pull/259/changes

cvanelteren:
The gha are also stalling as we hit nearly 1k tests. I need to update the selection criteria to work on baseline and tests.

cvanelteren:
Ok the CI is passing now -- will keep this open until I am sure that it is correctly fixed.

cvanelteren:
Think I fixed it

cvanelteren:
Still seeing some regressions - bit annoying.

cvanelteren:
I think we are good. There is still some caching on most of the PRs but other than that xdist is integrated and reduced the time to run the tests by a factor of 4.

PR #528, opened 2026-02-01
Stabilize pytest-mpl styling and colormap lookups

cvanelteren:
This tightens a few reliability issues in the test/docs pipeline. The demo font test now chooses a deterministic, preferred font so its mpl baseline is stable locally. Colormap lookups accept colormap objects directly, which fixes the Read the Docs build error when a ContinuousColormap is passed through. Added a small regression test for colormap object lookups.


cvanelteren:
Locally it closes #524

cvanelteren:
Merging to avoid the base difference

PR #529, opened 2026-02-01
Internal: Refactor colorbar to decouple from axis. Introduces UltraColorbar and UltraColorbarLayout

cvanelteren:
This PR refactors the colorbar logic to remove it from the `base` and `figure` class to ensure that we focus our efforts on the logic rather than mixing axis handling with colorbar handling. It achieves this by introducing a new module `colorbar` and with it two essential classes `UltraColorbar` and `UltraColorbar` layout. 

This PR will mark the first step towards this, and the legend object will need a similar rework.


cvanelteren:
Failing tests are not related to the PR (directly -- there is a slight layout drift that I still need to address).

cvanelteren:
For example: 
image

PR #530, opened 2026-02-01
Pin docutils for docs builds

cvanelteren:
The docs build currently fails because m2r2 expects `docutils.core.ErrorString`, which was removed in docutils 0.22. This pins docutils to <0.22 in the docs extra to keep Read the Docs working until we replace or update the m2r2 dependency.


cvanelteren:
Oh weird the build on #529 works now..

cvanelteren:
Not sure what went wrong but this is not required anymore.

PR #532, opened 2026-02-01
Fix UltraLayout spans for GridSpec slices

### LINKED ISSUE #531: UltraLayout not providing proper spacing for geo plots
new method:
Image

old method:
Image

cvanelteren:
UltraLayout wasn’t honoring GridSpec spans created via slicing (e.g., gs[0, :2]), so the solver treated those subplots as single cells and mis‑allocated space. This update merges spanned cells in the internal layout_array when slices are created and clears cached UltraLayout positions so the solver sees the correct geometry. The result restores expected sizing for span‑based layouts without changing the public API.

Closes #531


cvanelteren:
image


looks great!

PR #549, opened 2026-02-03
Add rc default for text borderstyle

cvanelteren:
By defualt matplotlib uses miter for borderstyle on text, this makes the annotated text look jagged. It would look nicer if this was beveled. Proposal to move to beveled by default and add an rc parameter to config this.
- add rc param text.borderstyle with validation and default miter
- wire Axes.text/label updates to respect rc default while keeping explicit overrides
- add tests covering rc default and explicit override behavior

Closes #548

PR #550, opened 2026-02-04
Feature: add curved annotation

cvanelteren:
Sometimes I need to annotate curved lines -- but there is no implementation for it native to matplotlib.  This PR integrates the logic from Thomas Kühn in UltraPlot:

image

### DIFF
--- a/.github/workflows/build-ultraplot.yml
+++ b/.github/workflows/build-ultraplot.yml
@@ -121,9 +121,9 @@ jobs:
         with:
           path: ./ultraplot/tests/baseline # The directory to cache
           # Key is based on OS, Python/Matplotlib versions, and the base commit SHA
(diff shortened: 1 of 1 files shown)


ultraplot/axes/base.py
```diff
@@ -3919,6 +3919,96 @@ def text(
...
+        else:
+            transform = self._get_transform(transform)
+        kwargs["transform"] = transform
+
+        from ..text import CurvedText
+
+        obj = CurvedText(x, y, text, axes=self, **kwargs)
+
+        if borderstyle is None:
+            try:
```
cvanelteren: Need to replace this with the rc on the other PR #549

beckermr (changes requested):
Defaults in the RC as usual.

PR #559, opened 2026-02-06
Feature: Add top-aligned ribbon flow plot type

cvanelteren:
This PR adds a new ribbon flow plot type via ax.ribbon for fixed-row, top-aligned transition diagrams, distinct from Sankey layering because category rows remain globally stable across periods. It includes a reusable implementation in plot_types/ribbon.py, PlotAxes API wiring, a smoke test in test_plot.py, and a new plot-types gallery example using synthetic generic categories so the feature is documented without domain-specific assumptions.

### DIFF
--- a/docs/subplots.py
+++ b/docs/subplots.py
@@ -373,8 +373,9 @@ for (name1, equal), (name2, group) in zip(equals, groups):
 #    `~matplotlib.figure.Figure.supxlabel` and `~matplotlib.figure.Figure.supylabel`,
 #    these labels are aligned between gridspec edges rather than figure edges.
 # #. Supporting five sharing "levels". These values can be passed to `sharex`,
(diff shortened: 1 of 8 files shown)


ultraplot/axes/plot.py
```diff
@@ -2305,6 +2305,100 @@ def _looks_like_links(values):
         diagrams = sankey.finish()
         return diagrams[0] if len(diagrams) == 1 else diagrams
 
+    @docstring._snippet_manager
+    def ribbon(
```
beckermr, line 2309: Defaults in the rc.

cvanelteren: done

beckermr (approved):
Idk if the test failures matter or not. Otherwise Lgtm.

cvanelteren:
The tests are still a bit flaky -- been trying to fix it but it feels like whack-a-mole

PR #561, opened 2026-02-06
Add opt-in pixel snapping for subplot layout

cvanelteren:
This is an attempt to fix the CI; In order to lockdown the small drifts that occur when tests fails, we can snap the corners of the axis to the nearest pixel.


cvanelteren:
Need to merge to test on other branches.

PR #574, opened 2026-02-09
Fix suptitle top spacing for center/top vertical alignment

### LINKED ISSUE #573: Suptitle can overlap top content when suptitle_kw va is center/top
## Summary
When `fig.format(..., suptitle_kw={"va": "center"})` or `va="top"` is used, the suptitle anchor is moved without compensating for the text bbox, so the title can overlap top labels/axes content.

## Reproduction
```python
import ultraplot as uplt
fig, axs = uplt.subplots(ncols=2)
fig.format(
    suptitle='Long figure title\nsecond line',
    suptitle_kw={'va': 'top'},
    toplabels=('left', 'right'),
)
fig.canvas.draw()
```

## Expected
Top spacing remains consistent regardless of suptitle vertical alignment (`bottom`, `center`, `top`).

## Actual
For `center`/`top`, the suptitle can intrude into top content because layout uses the title anchor position instead of the suptitle bbox bottom.

## Notes
This appears after recent suptitle alignment updates where `suptitle_kw` alignment properties are preserved.

cvanelteren:
This fixes a suptitle layout regression where `suptitle_kw` vertical alignments (`va='center'` or `va='top'`) could let the suptitle overlap top content. The alignment code now computes the target top offset from surrounding content and then positions the suptitle so its bbox bottom lands at that offset, preserving spacing regardless of alignment. It also adds a regression test covering `va in {'bottom','center','top'}` to enforce consistent top spacing. Closes #573.

### DIFF
--- a/ultraplot/figure.py
+++ b/ultraplot/figure.py
@@ -2322,18 +2322,23 @@ class Figure(mfigure.Figure):
         ha = self._suptitle.get_ha()
         va = self._suptitle.get_va()
 
-        # Use original centering algorithm for positioning (regardless of alignment)
+        # Use original centering algorithm for horizontal positioning.
         x, _ = self._get_align_coord(
             "top",
             axs,
             includepanels=self._includepanels,
             align=ha,
         )
-        y = self._get_offset_coord("top", axs, renderer, pad=pad, extra=labs)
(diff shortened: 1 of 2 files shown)


cvanelteren:
Still need to fix the regression of that one test.

Issue #578, opened 2026-02-11
Something need to be sorted about `geoaxes._toggle_ticks`.


gepcel:
The [`_toggle_ticks` function](https://github.com/Ultraplot/UltraPlot/blob/ed9d7372abf8edb473007ca2cebd3852dd2e5278/ultraplot/axes/geo.py#L1103) behavior does not match the [documentation](https://github.com/Ultraplot/UltraPlot/blob/ed9d7372abf8edb473007ca2cebd3852dd2e5278/ultraplot/axes/geo.py#L134).

Firstly, the documentation states that  `lonlabels`, `latlabels`, and `labels` can accept types such as `str`, `bool`, or `sequence`. However, the `_toggle_ticks` function currently only supports str inputs.

Additionally, using `labels='lb'` to display ticks only on the bottom and left sides does work functionally, but it triggers a warning message that is confusing and incorrect:

`UltraPlotWarning: Not toggling label='lb'. Input was not understood. Valid values are ['left', 'right', 'top', 'bottom', 'all', 'both']`

cvanelteren:
Toggle ticks is not part of the public API so there is no docs to match it with. We can make it part of the public API if you want/ if this is convenient.

gepcel:
Ok, you misunderstood me. Let me give a example.

<img width="1500" height="512" alt="Image" src="https://github.com/user-attachments/assets/158c7823-dcf7-4739-8a73-131e06a4f039" />

In the example above, `labels='lb` does work functionally. But the warning message is wrong about `Input was not understood`.

The documents indicates that `lonlabels`, `latlabels`, and `labels` can accept types such as `str`, `bool`, or `sequence`:

<img width="676" height="636" alt="Image" src="https://github.com/user-attachments/assets/934a1665-79d1-49e4-b8da-92657e70dfa1" />

So the warning message is wrong again about `Valid values are ['left', 'right', 'top', 'bottom', 'all', 'both']`. Combinations of `l r b t` are also valid.

I'm not suggesting the api to be public. I'm just guessing that the warning message comes from that func.

cvanelteren:
Ah ok yeah that should be addressed probably happened in the refactor

cvanelteren:
Fixed on branch `fix/geo-toggle-ticks-label-spec-578` in commit `4b93b933`.

[...]

PR #579, opened 2026-02-11
Fix GeoAxes _toggle_ticks label parsing

cvanelteren:
This fixes the GeoAxes tick-toggle parsing mismatch reported in #578.


cvanelteren:
Maintainer note: fix is on this PR from branch fix/geo-toggle-ticks-label-spec-578 (commit 4b93b933).

Behavior before fix:
- _toggle_ticks only handled string patterns and warned for valid shorthand like labels='lb'.

Behavior after fix:
- _toggle_ticks uses normalized label arrays, so bool/string/sequence inputs are accepted consistently with the GeoAxes label docs.
- Added focused regression tests to lock this behavior.

cvanelteren:
Merging as the image difference was not supposed to look like that.

PR #586, opened 2026-02-14
Feature: semantic legend API and geo legend support

cvanelteren:
This PR combines the #570/#571 groundwork into a user-facing semantic legend API on Axes (cat_legend, size_legend, num_legend, geo_legend) and wires defaults through rc so behavior is configurable consistently. It also adds geometry-aware legend handling (including country shorthand, projection/resolution/territory controls, and unified handle sizing), plus tests to cover shorthand loc behavior and geo legend rendering paths.

### DIFF
--- a/ultraplot/legend.py
+++ b/ultraplot/legend.py
@@ -1441,7 +1441,9 @@ class UltraLegend:
         styles.update(_pop_entry_props(styles))
         line = _not_none(line, styles.pop("line", None), rc["legend.cat.line"])
         colors = _not_none(colors, styles.pop("color", None))
-        markers = _not_none(markers, styles.pop("marker", None), rc["legend.cat.marker"])
+        markers = _not_none(
+            markers, styles.pop("marker", None), rc["legend.cat.marker"]
+        )
         linestyle = _not_none(
             linestyle,
             styles.pop("linestyle", None),


ultraplot/legend.py
```diff
@@ -94,6 +129,873 @@ def marker(cls, label=None, marker="o", **kwargs):
...
+):
+    """
+    Build categorical semantic legend handles and labels.
+    """
+    labels = list(dict.fromkeys(categories))
+    palette = _default_cycle_colors()
+    handles = []
+    for idx, label in enumerate(labels):
+        color = _style_lookup(colors, label, idx, default=palette[idx % len(palette)])
+        marker = _style_lookup(markers, label, idx, default="o")
```
gepcel, line 941: Besides color and marker, I suggest cat_legend support all marker kws, as different entries, like ls, ms, lw, ec, markeredgewidth, alpha, etc.

ultraplot/legend.py
```diff
@@ -94,6 +129,873 @@ def marker(cls, label=None, marker="o", **kwargs):
...
+    """
+    values = np.asarray(list(levels), dtype=float)
+    if values.size == 0:
+        return [], []
+    if area:
+        ms = np.sqrt(np.clip(values, 0, None))
+    else:
+        ms = np.abs(values)
+    ms = np.maximum(ms * scale, minsize)
+    labels = [_format_label(value, fmt) for value in values]
```
gepcel, line 1083: Legend label are from values as defult, I suggest it accepct arrays as string as labels. Like: `ax.size_legend([5, 10, 15], labels=['A', 'B', 'C']). I'm thinking if it makes sense with different fc, ec, ...

ultraplot/legend.py
gepcel: After giving it another thought, instead ovload cat_legend, size_legend.., maybe a another universial method, whick can handle different ms, fc, ls, ec, alpha, labels, all kind of kws for a legend marker?

ultraplot/legend.py
gepcel: It will useful for circumstances like `labels=['<0', '10-20', '>20']`

gepcel:
Seems that cat_legend, size_legend, num_legend, geo_legend take `label="Legend"` as legend title. I suggest it only accepts `title= ...`. Because `label` and `labels` easly to get confused. Not easy to tell difference with all arg and it 's plura.

cvanelteren:
Agreed. The paths look a bit funky with the default `lw=1.0` but this is something that happens on the geometry backend.

cvanelteren:
Should the api be `catlegend` or `cat_legend`?

gepcel:
I prefer `catlegend`.

cvanelteren:
I renamed the public API methods:

  - ax.cat_legend(...) -> ax.catlegend(...)
  - ax.size_legend(...) -> ax.sizelegend(...)
  - ax.num_legend(...) -> ax.numlegend(...)
  - ax.geo_legend(...) -> ax.geolegend(...)

cvanelteren:
@gepcel if you are ok with it, we can proceed with a v2 release.

ultraplot/legend.py
cvanelteren: didn't get notified by these comments; will have a final look and try to expand it. Some of the mechanisms are already there

Issue #588, opened 2026-02-17
Compatibility issue with matplotlib-map-utils package


chudlerk:
Hi,

I've recently been making maps using the new(ish) [matplotlib-map-utils](https://geopandas.org/en/latest/gallery/matplotlib_scalebar.html#Using-matplotlib-map-utils) package, referenced in the new geopandas [documentation](https://geopandas.org/en/latest/gallery/matplotlib_scalebar.html#Using-matplotlib-map-utils).

There appear to be some compatibility issues somewhere with ultraplot.

If I run this simple code:
```
from matplotlib_map_utils.core.scale_bar import ScaleBar, scale_bar
import matplotlib.pyplot as plt
import cartopy.crs as ccrs

bounds = [-122.3094136,   41.02776967, -121.65305008,   41.61986174]
c_lon = (bounds[0] + bounds[2]) / 2
c_lat = (bounds[1] + bounds[3]) / 2
plot_crs = ccrs.LambertConformal(central_longitude=c_lon, central_latitude=c_lat)

fig, ax = plt.subplots(subplot_kw={'projection': plot_crs}, dpi=150)
ax.set_extent((bounds[0], bounds[2], bounds[1], bounds[3]), crs=ccrs.PlateCarree())
scale_bar(ax, bar={"projection": plot_crs})
fig.savefig('test.png', dpi=150)
```

The code runs and the figure saves just fine.

However, if I do this with ultraplot instead..
```
from matplotlib_map_utils.core.scale_bar import ScaleBar, scale_bar
import matplotlib.pyplot as plt
import cartopy.crs as ccrs

bounds = [-122.3094136,   41.02776967, -121.65305008,   41.61986174]
c_lon = (bounds[0] + bounds[2]) / 2
c_lat = (bounds[1] + bounds[3]) / 2
plot_crs = ccrs.LambertConformal(central_longitude=c_lon, central_latitude=c_lat)

fig, ax = uplt.subplots(projection=plot_crs, dpi=150)
ax.set_extent((bounds[0], bounds[2], bounds[1], bounds[3]), crs=ccrs.PlateCarree())
scale_bar(ax, bar={"projection": plot_crs})
fig.savefig('C:/Users/kchudler/Downloads/ultraplot_test.png', dpi=150)
```

The numbers and text around the scale bar appear fuzzy in the saved image.
[...]

PR #589, opened 2026-02-17
Scope raster rc defaults (partial fix for map-utils compatibility)

cvanelteren:
This PR partially addresses #588 by reducing global Matplotlib rc side-effects that leak into third-party rasterized artists, notably matplotlib-map-utils ScaleBar.

Changes:
- Stop forcing global image.interpolation to none in UltraPlot defaults.
- Stop forcing global savefig.format to pdf in UltraPlot defaults.
- Keep UltraPlot visual behavior by applying interpolation=none locally in UltraPlot image-producing methods (imshow, pcolorfast image branch, spy image branch), preserving user overrides.
- Add regression tests that creating UltraPlot figures does not force global image.interpolation or savefig.format, and that UltraPlot imshow still defaults to none.

### DIFF
--- a/ultraplot/axes/plot.py
+++ b/ultraplot/axes/plot.py
@@ -7138,6 +7138,7 @@ class PlotAxes(base.Axes):
         %(plot.pcolorfast)s
         """
         x, y, z, kw = self._parse_2d_args(x, y, z, edges=True, **kwargs)
+        user_interpolation = "interpolation" in kw
         kw.update(_pop_props(kw, "collection"))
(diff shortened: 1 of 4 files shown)


cvanelteren:
See the 2 proposed fixes in #588 -- we can also force the interpolation and them fix the rasterization on their end. We are still leaking the environment which may not be wanted --> not sure how we wouldd clean up ultraplot if matplotlib is imported, i.e. we are permanently changing the RC for them too.

beckermr (approved):

PR #593, opened 2026-02-23
Fix inset colorbar frame reflow for refaspect

### LINKED ISSUE #592: Inset colorbar's title is out of the frame with refaspect
I am new to UltraPlot and am not sure if this is a bug or an intended behavior.

The inset colorbar's title is not included with the frame if refaspect is set for the figure.

```
import numpy as np
import ultraplot as uplt

# fig, axs = uplt.subplots(refwidth=4)  # ok
fig, axs = uplt.subplots(refaspect=2)  # fail

axs.pcolormesh(
    np.random.rand(20, 30), vmin=0, vmax=1,
    colorbar='ul',    
    colorbar_kw=dict(
        label='title',
    ),
)
```

Image

UltraPlot version: '2.0.1'
JupyterLab version: 4.5.4 on MS Edge 145.0.3800.70, Windows11
Jupyter server on Ubuntu 24.04.4

cvanelteren:
Closes #592 
With refaspect, the inset colorbar geometry is finalized during draw. Previously, we were reflowing the inset frame too late (effectively after draw scheduling), so some render paths used stale bounds and the frame no longer wrapped the label correctly.

The fix moves frame reflow into the actual draw path and runs it before artists are painted, using the active renderer. We also mark inset colorbars as needing reflow when created, so the frame is recomputed at the right time. That makes the frame/label layout deterministic across backends and preserves correct wrapping with refaspect.

A regression test was added to lock this behavior.

### DIFF
--- a/ultraplot/gridspec.py
+++ b/ultraplot/gridspec.py
@@ -1988,9 +1988,11 @@ class SubplotGrid(MutableSequence, list):
         elif not isinstance(objs, list):
             objs = [objs]
 
+        # Spanning subplots can appear more than once in the sliced slot grid.
+        # De-duplicate while preserving order so method dispatch does not repeat.
+        objs = list(dict.fromkeys(objs))
         if len(objs) == 1:
             return objs[0]
-        objs = [obj for obj in objs if obj is not None]
         return SubplotGrid(objs)
 
     def __setitem__(self, key, value):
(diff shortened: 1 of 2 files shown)


cvanelteren:
Fix is relatively minor; just aligning already present machinery to work for the inset properly -- merging without review @beckermr

PR #598, opened 2026-02-26
Bugfix: Deduplicate spanning axes in SubplotGrid slicing

### LINKED ISSUE #597: Duplicate legend entries when plotting on sliced spanned axes with `legend=True`
When creating a figure with a spanned subplot layout using a 2D array, slicing the axes grid to access the spanned subplot (e.g., `axs[0, :2]`) and calling `.scatter()` with `legend=True` results in duplicate legend entries.

**Reproducible Example:**

```python
import numpy as np
import ultraplot as uplt

array = np.array(
    [[0.71510417, 0.18342202],
    [0.63857324, 0.49007551],
    [0.34293048, 0.81424364],
    [0.93123847, 0.54010325],
    [0.12053916, 0.67823549]]
)

fig, axs = uplt.subplots(np.array([[1, 1, 2], [3, 4, 5]]))
ax = axs[0, :2] 

ax.scatter(
    x=array[:, 0], y=array[:, 1], c="grey", label="data", legend=True
)

```

Image

A quick workaround is to replace `ax = axs[0, :2]` with `ax = axs[0]`. However, I'm not sure if this duplicate behavior is intended, or if the usage of `axs[0, :2]` is simply discouraged in ultraplot.

cvanelteren:
Closes #597 

Dedups the indexing from gridspec to prevent calling legend multiple times.

### DIFF
--- a/ultraplot/gridspec.py
+++ b/ultraplot/gridspec.py
@@ -1990,20 +1990,7 @@ class SubplotGrid(MutableSequence, list):
 
         # Spanning subplots can appear more than once in the sliced slot grid.
         # De-duplicate while preserving order so method dispatch does not repeat.
-        try:
-            objs = list(dict.fromkeys(objs))
-        except TypeError:
-            # Fallback for any unhashable custom axes-like objects.
-            seen = set()
-            unique = []
-            for obj in objs:
-                obj_id = id(obj)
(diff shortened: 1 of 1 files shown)


cvanelteren:
This would work but is rather uggly. Looking for another fix.

PR #607, opened 2026-03-07
Feature: Add AstroPy Support

### LINKED ISSUE #606: Feature request: Handle WCS projections for plotting astronomical images.
<div align="justify">

Currently, it seems that plotting with WCS does not work in UltraPlot. Here is a minimal working example:

```python
import ultraplot as uplt
from astropy.coordinates import SkyCoord

coords = SkyCoord("04h30m49.1s +64d50m53.0s")

fig = uplt.figure()
ax = fig.add_axes([0.1,0.1,0.85,0.85], projection=mosaic.wcs)
ax.plot(
    coords.ra,
    coords.dec,
    ".",
    transform=ax.get_transform("icrs"),
    mec="b",
    mfc="none",
)
uplt.show()
```

This raises the following error:

```bash
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
Cell In[13], line 11
      5 fig = uplt.figure()
      6 ax = fig.add_axes([0.1,0.1,0.85,0.85], projection=mosaic.wcs)
      7 ax.plot(
      8     coords.ra,
      9     coords.dec,
     10     ".",
---> 11     transform=ax.get_transform("icrs"),
     12     mec="b",
     13     mfc="none",
     14 )
     15 uplt.show()

TypeError: Artist.get_transform() takes 1 positional argument but 2 were given
```

Using Matplotlib, this works:

```python
from astropy.coordinates import SkyCoord

coords = SkyCoord("04h30m49.1s +64d50m53.0s")

fig = plt.figure(1)
ax = fig.add_axes([0.1,0.1,0.85,0.85], projection=mosaic.wcs)
ax.plot(
    beam.ra,
    beam.dec,
    ".",
    transform=ax.get_transform("icrs"),
    mec="b",
    mfc="none",
)
plt.show()
```

I am assuming that UltraPlot cannot handle WCS projections, hence this issue is framed as a feature request. As far as I am aware, UltraPlot has a lot of features focused on plotting geographical data. It would be great if features related to plotting astronomical data and images could also be added, especially since many of the plots we make are quite similar to ones made for geographical data.

</div>

cvanelteren:
Closes #606 
Issue #606 is that UltraPlot does not correctly handle Astropy WCS projection objects passed as `projection=mosaic.wcs`.

  In practice, this shows up when code like `ax.get_transform("icrs")` fails with:

  ```
TypeError: Artist.get_transform() takes 1 positional argument but 2 were given
```

The underlying problem is:

  1. UltraPlot was not preserving/wrapping non-string projection objects like WCS during axes creation.
  2. The external-axes container then exposed the wrong get_transform() method, so WCS-specific transform lookups were
     masked by the base Matplotlib artist API.


cvanelteren:
Integration requires more effort see #606 -- don't have time to work on this atm.

cvanelteren:
@astrogewgaw can you test this?

astrogewgaw:
<div align="justify">

@cvanelteren I tested this using the minimal working example I shared earlier (the code for which has been added to the issue now as a comment [**here**](https://github.com/Ultraplot/UltraPlot/issues/606#issuecomment-4044218142); do have a look), and it works! I will see if I can test it with some FITS images as well. Also, maybe @beckermr can also have a look at this if he has some time? Just a thought.

</div>

cvanelteren:
Need to do some more testing and quality check the implementation; Astropy seems like a serious undertaking. We should also reach out to add us on the list of packages if this goes through.

cvanelteren:
Key thing is that supporting other axis implementations is that it will become harder for me to keep up if the backend changes on their end; if we keep this in house and it changes significantly this would require me to keep up-to-date with their changes. Since I am not an astro-person, this would be harder to catch than users that use it every day. Naturally, I can write tests but this will not be perfect.

astrogewgaw:
@cvanelteren Makes sense. I can help with some of the testing, but I don't deal with that many astronomical images in my line of work, I am more of a time-domain astronomy guy. Do let me know if and when this makes to a future release of UltraPlot; until then, I can point my package to the latest commit on the `main` branch.

cvanelteren:
Putting this on the back burner as it is nicer to add this after a refactor happens on `figure.py` that will design contracts as to how we interface with other axes in clean way so I don't have to repeat my steps.

cvanelteren:
This needs review preferably from someone in the field -- otherwise it is ready to go.

PR #616, opened 2026-03-17
Honor cmap for numeric scatter colors

### LINKED ISSUE #615: ax.scatter Ignores "cmap" Argument When 'c' Argument is a List of Values Between 0 and 1
Hello,

When I run the following simple script:
```
import ultraplot as uplt
fig,ax=uplt.subplots()
ax.scatter(
    [1.0, 2.0, 3.0, 4.0],
    [1.0, 2.0, 3.0, 4.0],
    c=[.1, .2, .3, .4], 
    cmap='turbo',
    )
```

I get the following warning:
`UltraPlotWarning: Ignoring unused keyword arg(s): {'cmap': 'turbo'}`
and the resulting scatter plot does not use the specified colormap.

If I do the same thing with matplotlib, it works fine:
```
fig,ax=plt.subplots()
ax.scatter(
    [1.0, 2.0, 3.0, 4.0],
    [1.0, 2.0, 3.0, 4.0],
    c=[0.1, 0.2, 0.3, 0.4],
    cmap='turbo',
    )
```

I am using the latest version of ultraplot (2.1.3)

cvanelteren:
This fixes a Matplotlib-compatibility bug in `ax.scatter(...)` when `c` is a 1D numeric array. UltraPlot was treating short numeric sequences like `[0.1, 0.2, 0.3, 0.4]` as literal colors too early, which caused `cmap` to be dropped as unused instead of applying colormapping. The fix keeps explicit `N x 3` / `N x 4` RGB(A) handling intact, but treats 1D numeric arrays matching the point count as scalar data for colormapping, which is the behavior users expect from Matplotlib. A regression test covering the reported `c=[...], cmap="turbo"` case is included.

Closes #615.

### DIFF
--- a/ultraplot/axes/plot.py
+++ b/ultraplot/axes/plot.py
@@ -11,7 +11,7 @@ import re
 import sys
 from collections.abc import Callable, Iterable
 from numbers import Integral, Number
-from typing import Any, Iterable, Mapping, Optional, Sequence, Union
+from typing import Any, Iterable, Mapping, Optional, Sequence, TypeAlias, Union
 
 import matplotlib as mpl
 import matplotlib.artist as martist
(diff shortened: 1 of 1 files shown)


cvanelteren:
Reminder: review this later today.

PR #618, opened 2026-03-17
Add manual text staggering for overlapping labels

### LINKED ISSUE #617: Add text staggering for overlapping labels and annotations
Overlapping text boxes and annotations are a common pain point in dense figures. UltraPlot should provide an explicit helper to stagger existing text() and annotate() artists without changing their default behavior.

A useful API would support both subplot-level and figure-level usage, apply offsets independently within each subplot, and allow horizontal, vertical, or two-dimensional staggering so users can reduce collisions without manually tweaking every label.

cvanelteren:
This adds an explicit text-staggering helper for dense figures without changing the default behavior of text placement. `Axes.stagger_text()` can now offset existing `text()` and `annotate()` artists horizontally, vertically, or in two dimensions to reduce overlap, and `Figure.stagger_text()` applies the same behavior across selected subplots while keeping each subplot independent. The PR also adds targeted regressions for text, annotation, idempotency, and figure-level selection behavior. Closes #617.

### DIFF
--- a/ultraplot/axes/base.py
+++ b/ultraplot/axes/base.py
@@ -4051,19 +4051,50 @@ class Axes(_ExternalModeMixin, maxes.Axes):
     @staticmethod
     def _pad_stagger_bbox(
         bbox: Optional[mtransforms.Bbox],
-        pad_px: float,
+        pad_x_px: float,
+        pad_y_px: float,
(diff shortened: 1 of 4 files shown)


cvanelteren:
#754 has implemented this partially

PR #630, opened 2026-03-19
Fix duplicate shared boxplot tick labels

### LINKED ISSUE #628: Boxplots have duplicate xtick labels when enabling x axis sharing
When sharex in the following code is set to 2 or above, there are duplicated xticklabels. This issue also shows up with xrotation = 0 because the ticklabels appear bold (double rendering). 

```
import numpy as np
import ultraplot as uplt
data = [np.random.random(size=j * 50) for j in range(1, 11)]

f, axs = uplt.subplots(
    nrows=2,
    ncols=2,
    xrotation=45,
    figwidth="6in",
    figheight="4in",
    xminorlocator="null",
    grid=False,
    sharey = False, sharex = 2,
    xlabel = 'Set Index', ylabel = 'Value',
    suptitle = 'Boxplot of random numbers between 0 and 1 with increasing set size'
)
for idx, ax in enumerate(axs, 1):
    ax.format(title=f"Panel Number {idx}")
    ax.boxplot(data, showfliers=False, lw = 0.5)
```

I don't know how to look at the internals of ultraplot, but I can attach the two figures for different values of sharex here.

Image
Image

cvanelteren:
This fixes the shared-axis boxplot regression where repeated `boxplot` calls on axes with `sharex >= 2` could append the same fixed tick labels multiple times, making the x tick labels appear duplicated or bold. The boxplot wrapper now avoids Matplotlib's appending tick manager on already shared axes and applies the intended tick labels once, with a regression test covering the reported case. Closes #628.

### DIFF
--- a/ultraplot/tests/test_plot.py
+++ b/ultraplot/tests/test_plot.py
@@ -317,7 +317,7 @@ def test_boxplot_shared_x_axes_do_not_duplicate_tick_labels():
     for ax in axs:
         ax.boxplot(data, showfliers=False, lw=0.5)
 
-    expected = [str(i) for i in range(10)]
+    expected = [str(i) for i in range(len(data))]
     for ax in axs:
         labels = [tick.get_text() for tick in ax.xaxis.get_ticklabels()]
         assert labels == expected


cvanelteren:
Manual inspection looks good. Merging.

PR #634, opened 2026-03-19
feat: add 'plot.sticky' RC flag to toggle sticky edges

### LINKED ISSUE #631: RC flag to disable sticky edges for line plots
Quoting from this page,
https://ultraplot.readthedocs.io/en/latest/1dplots.html#Line-plots

> Also, the default x bounds for lines drawn with [plot()](https://ultraplot.readthedocs.io/en/latest/api/ultraplot.axes.PlotAxes.html#ultraplot.axes.PlotAxes.plot) and y bounds for lines drawn with [plotx()](https://ultraplot.readthedocs.io/en/latest/api/ultraplot.axes.PlotAxes.html#ultraplot.axes.PlotAxes.plotx) are now “sticky”, i.e. there is no padding between the lines and axes edges by default.

Is there a way to globally disable sticky edges for line plots in the rc file, instead of having to type ax.disable_sticky_edges = False in every axis individually?

Deepak8858:
Fixes #631

Added a new 'plot.sticky' configuration setting (default True) that allows globally disabling sticky edges for line and area plots via the rc file or rc context.

### DIFF
--- a/ultraplot/axes/plot.py
+++ b/ultraplot/axes/plot.py
@@ -3497,6 +3497,8 @@ class PlotAxes(base.Axes):
         Fix sticky edges for the input artists using the minimum and maximum of the
         input coordinates. This is used to copy `bar` behavior to `area` and `lines`.
         """
+        if not rc["plot.sticky"]:
+            return
         for array in args:
             min_, max_ = inputs._safe_range(array)
             if min_ is None or max_ is None:
--- a/ultraplot/internals/rcsetup.py
+++ b/ultraplot/internals/rcsetup.py
@@ -1050,6 +1050,12 @@ _rc_ultraplot_table = {
         _validate_float,
         "Vertical margin around ribbon diagrams (axes-relative units).",
     ),
+    # Plot settings
+    "plot.sticky": (
+        True,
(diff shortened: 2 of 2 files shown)


ultraplot/internals/rcsetup.py
cvanelteren, line 1: Needs a unit smoke test

ultraplot/axes/plot.py
```diff
@@ -3497,6 +3497,8 @@ def _fix_sticky_edges(self, objs, axis, *args, only=None):
         Fix sticky edges for the input artists using the minimum and maximum of the
         input coordinates. This is used to copy `bar` behavior to `area` and `lines`.
         """
+        if not rc["plot.sticky"]:
```
cvanelteren, line 3500: not sure if this is the correct place here. This would affect fill, and area plots too. We could add this checker in the line only for now (as I don't see the point for adding it for fills and areas). We may also add it to the plot kwargs to allow to turn it off per subplot.

cvanelteren (changes requested):

cvanelteren:
Closing due to inactivity.

PR #644, opened 2026-03-20
Improve gallery widget and thumbnail backgrounds

cvanelteren:
Update the CS to make the gallery pop more on the background.


cvanelteren:
image

PR #647, opened 2026-03-20
fix(tests): stabilize test_ignored_keywords across CI

### LINKED ISSUE #641: Stabilize the ignored-keyword warning test across CI
`ultraplot/tests/test_format.py` contains a skipped `test_ignored_keywords()` with the note that it fails on GitHub but not locally. The warning paths themselves are real, but the test currently looks brittle.

This issue is intentionally scoped for newcomers to the project. It is low complexity, touches a single test, and is a good way to get familiar with UltraPlot's warning behavior without changing core plotting logic.

Suggested scope:
- inspect why the current test is environment-sensitive
- replace raw warning-count assertions with more specific checks on the relevant warning messages
- keep the test focused on `gridspec_kw`, `subplotpars`, and `fig.subplots_adjust()` behavior

Acceptance criteria:
- the skip marker is removed
- the test passes locally and in CI
- the assertions are specific enough that unrelated warnings do not cause failures

Bingtagui404:
## Summary

- Remove `@pytest.mark.skip` from `test_ignored_keywords`
- Replace brittle `len(record) == 3` count assertions with specific message-content checks
- Filter warnings by `UltraPlotWarning` category using `issubclass` instead of string matching
- Add `warnings.simplefilter("always")` to ensure warnings are captured regardless of environment
- Each assertion now checks that the expected keyword name appears in a warning message

closes #641

### DIFF
--- a/ultraplot/axes/plot.py
+++ b/ultraplot/axes/plot.py
@@ -3509,7 +3509,9 @@ class PlotAxes(base.Axes):
                 edges.extend(convert((min_, max_)))
 
     @staticmethod
-    def _fix_patch_edges(obj, edgefix=None, **kwargs):
+    def _fix_patch_edges(
+        obj, edgefix=None, default_linewidth: float | None = None, **kwargs
+    ):
         """
         Fix white lines between between filled patches and fix issues
         with colormaps that are transparent. If keyword args passed by user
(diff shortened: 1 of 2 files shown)


cvanelteren:
Hi thanks @Bingtagui404, not to offend but was this generated with an agent? We don't have an active policy (yet) on AI but generally prefer the good first issues to be solved by humans to allow for getting familiar with the codebase.

Bingtagui404:
Hi @cvanelteren, thanks for asking — yes, this was largely AI-assisted (Claude Code).

If you'd prefer good first issues to be solved by humans, I completely understand. I'm happy to close this PR so it doesn't take up your review time. Just let me know!

cvanelteren:
Thanks for the clarification.  Will close for now.

cvanelteren:
Will draft some instructions to prevent this in the future. For more complex tasks I'm not against it, but what we need right now is more hands developing, so I become useless in the end 😁

Issue #648, opened 2026-03-20
rc entry for legend box linewidth


anirudhkrishnan2718:
Hi,

In matplotlib, setting 'patch.linewidth' can be used to change the legend box linewidth indirectly.
Add a direct rc entry to set this, in the same vein as legend.geo.linewidth.

cvanelteren:
You can control rc parametrs from matplotlib directly, but after a brief test I see that for some reason patch.linewidth is not responding on line plots but seems to be working for other artists.

cvanelteren:
In short, all matplotlib rcs are also ultraplot rcs so we are backwards compatible. We just add to matplotlib in a meaningful manner.

cvanelteren:
Ok so I looked in the wrong direction. `patch.linewidth` would control the linewidth for all 2D artists. I think adding these for legends is great. 

We could add a general `legend.linewidth` to control the frame width.

cvanelteren:
One would need to modify `_parse_frame()` and add the rc to the `rcsetup.py`

cvanelteren:
@K-Mirembe-Merc how bout this^

K-Mirembe-Mercy:
Hi! I’d like to work on this by adding a "legend.linewidth" rc parameter and updating "_parse_frame()" to use it. Let me know if that approach sounds good

cvanelteren:
@K-Mirembe-Mercy yep

K-Mirembe-Mercy:
Hi! I’ve now opened a PR implementing the "linewidth" and "lw" support in "_parse_frame()" as discussed.

Please let me know if any changes are needed — happy to adjust!

PR #649, opened 2026-03-20
Honor patch linewidth rc for edgefix

cvanelteren:
Preserve `rc['patch.linewidth']` for patch-style 2D artists when edgefix is active, and add targeted regressions for bar, hist, pie, and fill_between while leaving collection edgefix behavior unchanged.

 Not related to #648  but still a bug.


cvanelteren:
Manual inspection looks good and expected. Merging.

Issue #651, opened 2026-03-21
Feature request: Shared title for axes slicing


cvanelteren:
We need to add a feature where slices are meaningful so we add shared titles. See #650

PR #652, opened 2026-03-21
Use format() for shared subplot slice titles

cvanelteren:
We should make more use of the slicing feature to format a subset and assume the user aims to share the titles. This is a move in making the `format` function more intuitive.

Closes #650 #651


cvanelteren:
Need to fix a small bbox alignment but other than that it works.

gepcel:
The fix messed up the spaces between rows of axes, when using shared titles with subset of axes.

cvanelteren:
@gepcel  do you have an example?

cvanelteren:
```python
import ultraplot as uplt

# Test whether ignored for bar plots
fig, ax = uplt.subplots(ncols=2, nrows=2)
ax[0, :].format(title="True")
ax[1, :].format(title="True")
uplt.show(block=1
```
Looks good I think

cvanelteren:
image

gepcel:
```python
fig,ax=pplt.subplots(ncols=4, nrows=2, refwidth=1)
ax[:,:3].format(title="A test title")
```

image

cvanelteren:
Thanks! Funnily enough the code works fine when index to the first row. Will investigate.

cvanelteren:
Found the issue; it was feeding back into tight layout causing the figure to act as if the text was still there, overinflating the hspace.

cvanelteren:
Good catch @gepcel

gepcel (approved):

beckermr (approved):

gepcel (approved):

PR #653, opened 2026-03-21
Add support for linewidth in legend frame parsing

K-Mirembe-Mercy:
This PR adds support for `linewidth` and `lw` as aliases for controlling the legend frame edge width.

Previously, only `edgewidth` was supported. This improves usability and aligns behavior with matplotlib conventions.

Changes:
- Updated `_parse_frame()` to accept `linewidth` and `lw`.

### DIFF
--- a/ultraplot/tests/test_legend.py
+++ b/ultraplot/tests/test_legend.py
@@ -22,6 +22,13 @@ def test_auto_legend(rng):
     ix = ax.inset_axes((-0.2, 0.8, 0.5, 0.5), zoom=False)
     ix.line(rng.random((5, 2)), labels=list("pq"))
     ax.legend(loc="b", order="F", edgecolor="red9", edgewidth=3)
+    # Test lw
+    leg = ax.legend(frameon=True, lw=5)
+    assert leg.get_frame().get_linewidth() == 5
+
+    # Test lw alias
+    leg2 = ax.legend(frameon=True, lw=3)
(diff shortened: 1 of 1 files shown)


cvanelteren:
Hi @K-Mirembe-Mercy.  Good start but this doses not solve the issue yet. While parseframe now accepts the new keywords, we need to update the parts of the code so that it updates the frame. 

You can have a look how that is performed for `legend.geo`. After the code is implemented (or before), we need to add unit tests that will ensure correctness of to e added code. 
Hope this helps.

cvanelteren:
The commit marked as 0 is also a bit curious. Could you restore it to the one before.

K-Mirembe-Mercy:
Thanks for the clarification! I’ve updated the handling so that "kw_frame" is no longer mutated when extracting specific kwargs. Frame properties like "linewidth" are now preserved and correctly applied via "legendPatch.update".

cvanelteren:
@K-Mirembe-Mercy the history is dirty here, you need to rebase from main and include the tests in this PR -- no separate PR is necessary.

K-Mirembe-Mercy:
Thanks for the feedback! I’ve rebased from main and added tests to this PR.

I’ll resolve the remaining conflicts and update the branch shortly.

K-Mirembe-Mercy:
Closing this PR due to merge conflicts. I’ve opened a new clean PR with the same changes for easier review.

PR #657, opened 2026-03-23
Feature/legend title kw clean

cvanelteren:
Somehow this was not implemented.

Continuation of #656  after messing with claude a bit (new to it).


cvanelteren:
Regression is not related to the PR. I will need to think if this is a correct fixas we do list this property in the docs but it did not exist since ProPlot times. I ran into this problem when wanting to align the title, but this can be achieved through `align`

PR #659, opened 2026-03-25
Add typing block for inspection

cvanelteren:
This is a bit uncharted territory for me, but I believe the typing block can be used to ensure that we still have quick load time but type hinting can look at the definition of our module and functions.

New public api would need to be explicitly added here to ensure that it does not break inspection.  

Closes #655

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -63,7 +63,6 @@ ignore = ["I001", "I002", "I003", "I004"]
 
 [tool.basedpyright]
 exclude = ["**/*.ipynb"]
-stubPath = ""
 
 [tool.pytest.ini_options]
 filterwarnings = [


cvanelteren:
@anirudhkrishnan2718 this should work I think.

anirudhkrishnan2718:
Once these changes are merged, can I install the current version of the repository or do I need to wait for an official release?
I am guessing there is some way to install a 'nightly' version of ultraplot that does not require a new version to be released onto conda-forge.

cvanelteren:
You can pip install directly any branch from git from pip -- or clone it and do a `pip install -e`

pyproject.toml
```diff
@@ -63,6 +63,7 @@ ignore = ["I001", "I002", "I003", "I004"]
 
 [tool.basedpyright]
 exclude = ["**/*.ipynb"]
+stubPath = ""
```
cvanelteren: Don't think this is necessary: first tried generating stubs with pyright but that did not work.

ultraplot/tests/test_imports.py
```diff
@@ -139,6 +141,121 @@ def test_internals_lazy_attrs():
...
+    missing = [name for name in declared if not hasattr(ultraplot, name)]
+    assert not missing, (
+        "Names declared in the TYPE_CHECKING block are not accessible at runtime.\n"
+        "Either remove them from the TYPE_CHECKING block or expose them via the "
+        "lazy loader / _LAZY_LOADING_EXCEPTIONS:\n"
+        + "\n".join(f"  ultraplot.{n}" for n in missing)
+    )
+
+
+def test_type_checking_block_covers_public_all():
```
cvanelteren: This should protect us from adding public api without adding it to the typehinting block.

ultraplot/__init__.py
```diff
@@ -7,10 +7,151 @@
...
+if TYPE_CHECKING:
+    # These imports are never executed at runtime, so they have zero effect on
+    # import performance. They exist solely so that type checkers (pyright, mypy)
+    # can resolve names that are otherwise provided by the lazy loader at runtime.
+    #
+    # Keep this block in sync with _LAZY_LOADING_EXCEPTIONS and every submodule's
+    # __all__ — that is the only maintenance burden.
+    import matplotlib.pyplot as pyplot
+
+    from .axes import Axes as Axes
```
cvanelteren: `as Name` is necessary(!).

cvanelteren:
Me neither. This is what I gathered from stackoverflow. Normally these packages would just import things or provide a .pyi file but we can't do that as it is hidden behind when code is actually used and imported.

PR #661, opened 2026-03-25
Attempt fix missing logo

cvanelteren:
On iOS there is a missing logo -- which bugs me. I think this is an issue with the div not being rendered properly.
Will check if this fixes it.


cvanelteren:
Doesn't fix it. Will try again later.

cvanelteren:
Closing as I am not sure what the fix would be but this isn´t it.

PR #662, opened 2026-03-25
docs: add AI contribution policy

cvanelteren:
We are seeing an uptick of AI-aided PRs submitted to UltraPlot. While I have no issue with the use of AI per se, I am a bit concerned that it will not help flow through from novice to more adapt for people who do want to get involved. The issues marked as "Good first issue" are designed for others to pick-up. These are usually lower in complexity and are genuinely intended for people to pick up as they are small enough in scope. However, if we let AI execute on them, the aim is gone: people won't get familiar with the codebase, but the issue does get fixed.

I therefore think it is time to get an AI explicit policy going.

### DIFF
--- a/docs/contributing.rst
+++ b/docs/contributing.rst
@@ -241,6 +241,39 @@ Note that you can create the pull request before you're finished with your
 feature addition or bug fix. The PR will update as you add more commits. UltraPlot
 developers and contributors can then review your code and offer suggestions.
 
+.. _contrib_ai:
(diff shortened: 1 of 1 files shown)


cvanelteren:
Would love to have your input here on this @beckermr @gepcel

gepcel:
Although AI is rapidly developing and evolving, at the current stage of AI development, I agree with your decision.

PR #680, opened 2026-04-13
Fix: axes aspect shifting on pixel snapping after drawn

### LINKED ISSUE #679: Unaligned subplot axes using GridSpec geometry
## Summary
I found the previously correct behavior for subplots with gridspec geometry.
When using `[[1, 2], [1, 3]]` geometry for subplots and setting the left axis's aspect ratio to "equal", the right axes slightly stick out both up and down.
Does this seem to be a regression bug?
It appears after v2.0.1.

## Minimal reproducer

```python
import numpy as np
import ultraplot as uplt

x = np.linspace(0, 1)
y = x * x

fig, axs = uplt.subplots([[1, 2], [1, 3]])
axs[0].plot(x, y)
axs[0].format(aspect="equal")  # if "auto" is specified, it seems fine.
```
In v1.72.0
Image

In v2.0.1
Image

cvanelteren:
Closes #679 by restoring alignment for mixed-span subplot layouts when one one of the axes uses a fixed aspect ratio. The regression showed up in arrangements like [[1, 2], [1, 3]], where an equal-aspect axis on the left would shrink inside its gridspec slot but the stacked axes on the right would keep their full vertical extent and visibly stick out above and below it. This change teaches the figure layout pass to propagate the aspect-constrained bounds across the neighboring subplots that share the same span, and adds a regression test for both the legacy and UltraLayout code paths so the layout stays visually consistent going forward.

### DIFF
--- a/ultraplot/figure.py
+++ b/ultraplot/figure.py
@@ -1164,6 +1164,85 @@ class Figure(mfigure.Figure):
                 which="both",
             )
 
+    def _align_aspect_constrained_axes(self, *, tol: float = 1e-9) -> None:
+        """
+        Propagate aspect-constrained spanning axes boxes across sibling rows/columns.
+
+        When a fixed-aspect subplot spans multiple rows or columns, matplotlib shrinks
+        just that axes inside its gridspec slot. In layouts like ``[[1, 2], [1, 3]]``
+        this leaves the adjacent stack slightly taller or wider than the spanning axes.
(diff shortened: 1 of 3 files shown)


cvanelteren:
need to fix the code patch and refactor some of this but otherwise LGTM.

cvanelteren:
Regression is not caused by this PR -- merging.

PR #688, opened 2026-04-15
Feature/span cbar slot based

### LINKED ISSUE #678: Feature request for colorbar support on geographic axes
Hello,

I am using the Ultraplot library to create a 2×2 grid of subplots. The first row consists of geographic axes, while the second row contains regular (non-geographic) axes.

I created a colorbar for the geographic plots, but I would like to position the colorbar between the two rows of subplots, spanning across both columns.

However, in my attempts, the colorbar always appears below the first subplot instead of being centered between the two rows.

How can I correctly place the colorbar in the middle between the two rows and have it span across the columns?

This issue is similar to #392.

```
import ultraplot as uplt
import numpy as np

fig, axs = uplt.subplots(nrows=2, ncols=2, proj=["npstere", "npstere", None, None])

rng = np.random.default_rng()
data = rng.random((100, 100))

cm = axs[0, 0].pcolormesh(data)
fig.colorbar(cm, loc="b", ax=axs[0, :], span=(1, 2))
```

Image

cvanelteren:
Continuation of #681
Should close #678  properly

### DIFF
--- a/ultraplot/axes/base.py
+++ b/ultraplot/axes/base.py
@@ -2482,9 +2482,7 @@ class Axes(_ExternalModeMixin, maxes.Axes):
                         oss = oss.get_topmost_subplotspec()
                         if oss.get_gridspec() is not gs:
                             continue
-                        o_r1, o_r2, _, _ = oss._get_rows_columns(
-                            ncols=gs.ncols_total
-                        )
+                        o_r1, o_r2, _, _ = oss._get_rows_columns(ncols=gs.ncols_total)
                         if o_r1 >= row1 and o_r2 <= row2:
                             opos = other.get_position()
                             vmin = min(vmin, opos.y0)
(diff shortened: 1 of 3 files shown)


gepcel:
Doesn't seem to get the right height of GeoAxes. Try the following code:

```
import ultraplot as uplt
import numpy as np

fig, axs = uplt.subplots(nrows=2, ncols=2, proj=["npstere", "npstere", 'cyl','cyl'])
# fig, axs = uplt.subplots(nrows=2, ncols=2)

rng = np.random.default_rng()
data = rng.random((100, 100))

cm = axs[0, 0].pcolormesh(data)
fig.colorbar(cm, loc="r", ax=axs[0], ref=axs[1,0])
fig.colorbar(cm, loc="r", ax=axs[0], ref=axs[:,1])
```

cvanelteren:
You mean this is not looking correct:
test

Are you referring to how the bottom ones are reshaping to a different aspect?

gepcel:
image

As ot mark ① has the same height of axs[1,0] when colorbar doesn't has span. but ② doesn't has the same height, with span.

cvanelteren:
I think this works. With the latest commit:

`["npstere", "cyl", "cyl", "cyl"]`: would adjust but
`["npstere", "cyl", "cyl", None]`: will not as the last line would have a plot that is longer.

cvanelteren:
Worst case we have this which I am ok with:
image

gepcel (approved):

cvanelteren:
Will let this ruminate a bit to see if I can improve this -- it is a bit janky imo.

PR #707, opened 2026-04-29
Feature: figure semantic legends

### LINKED ISSUE #706: Figure wide semantic legends
I think that the semantic legends are a really great addition, making some otherwise “tedious” patterns much easier. What would make the semantic legend even better is if they were available figure wide, e.g. `fig.catlegend(...)` to create a single set of legend entries that apply to all subplots in a figure.

cvanelteren:
Closes #706 

This exposes the public API for adding semantic legends to the figure class. This PR is a shallow pass however.

Currently UltraPlot has an awkward split on the semantic legends under `UltraLegend` and the actual thin matplotlib wrapper for the implementation of the legend under `Legend`. A future refactor would need  to integrate these two either through composition to reduce the overhead. It ties closely into a general refactor for the `Figure` class in general but is important to note regardless.

A follow-up refactor should separate semantic legend content generation from legend placement and rendering. The likely direction is a composition-based design with target-agnostic builders that return `(handles, labels)`, and thin `Axes` and `Figure` adapters that place those legends through their respective public APIs.

### DIFF
--- a/docs/colorbars_legends.py
+++ b/docs/colorbars_legends.py
@@ -477,13 +477,19 @@ axs.format(xlabel="xlabel", ylabel="ylabel", suptitle="Legend formatting demo")
 #
 # Legends usually annotate artists already drawn on an axes, but sometimes you need
 # standalone semantic keys (categories, size scales, color levels, or geometry types).
-# UltraPlot provides helper methods that build these entries directly:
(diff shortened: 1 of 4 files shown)


gepcel (approved):

cvanelteren:
Merging this as the feature is small. Hope you are still interested to maintaining this package long-term @beckermr.

PR #709, opened 2026-04-30
Fix scaling of title

### LINKED ISSUE #708: Center-aligned title's font size doesn't increase.
# Summary
The title's font size doesn't increase at some point when it is center-aligned.
This behavior may be correct for aligning an ABC label, but I'll report it because it differs from what the user expects.

# Reproducer

When I set the title to be center-aligned,
```python
fig, ax = uplt.subplots()
ax.format(
    abc=True,
    ctitle="title longlonglonglonglonglonglonglong",  # it is same as "title=..."
    fontsize=20,
)
```
Image

When I set the title to be left-aligned,
```python
fig, ax = uplt.subplots()
ax.format(
    abc=True,
    ltitle="title longlonglonglonglonglonglonglong",
    fontsize=20,
)
```
Image

cvanelteren:
Closes #708 

Adjusts title scaling through a combination of scaling and moving when abc is present.

### DIFF
--- a/ultraplot/axes/base.py
+++ b/ultraplot/axes/base.py
@@ -2927,10 +2927,18 @@ class Axes(_ExternalModeMixin, maxes.Axes):
                         if base_x >= ax1 + abc_title_sep:
                             max_width = base_x - (ax1 + abc_title_sep)
                     elif ha == "center":
+                        # Keep the requested font size for centered titles and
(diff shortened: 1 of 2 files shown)


cvanelteren:
An idea would be to have an autowrap feature. TODO for the feature I guess.

PR #711, opened 2026-05-01
Fix bar tick labels for xarray DataArray with string coordinate

kinyatoride:
## Summary

`ax.bar(da)` on a 1-D `xarray.DataArray` with a string coordinate renders the full multi-line `xarray` repr as each x-tick label instead of the underlying string values.

## Reproducer

```python
import xarray as xr
import ultraplot as uplt

gm = xr.DataArray(
    [1.0, 2.0, 3.0],
    coords={"ens": ["A", "B", "C"]},
    dims=["ens"],
    name="pr",
    attrs={"long_name": "precipitation", "units": "mm/day"},
)

fig, ax = uplt.subplots(refwidth=4)
ax.bar(gm)
print([t.get_text() for t in ax.get_xticklabels()])
```

**Before:** each tick label is `"<xarray.DataArray 'ens' ()> Size: 4B\narray('A', dtype='<U1')\nCoordinates:\n    ens      <U1 4B 'A'"` (and similarly for `'B'`, `'C'`).

**After:** `['A', 'B', 'C']`.

## Cause

In `ultraplot/internals/inputs.py::_meta_coords`, `labels = list(map(str, data))` did not unwrap the `xarray.DataArray` first. `_parse_1d_format` passes the coordinate through as a `DataArray` (not a numpy array); iterating a 1-D `DataArray` yields scalar `DataArray`s whose `str()` is the multi-line repr.

## Fix

Wrap `data` with `_to_numpy_array` (already defined in the same file) before stringifying, so iteration yields the raw element values for both `xarray.DataArray` and `pandas` inputs.

## Test

Added `test_meta_coords_xarray_string_coord` in `ultraplot/tests/test_inputs_helpers.py` (skipped if `xarray` is not installed).

Confirmed broken on `2.2.0`; fix verified locally with `pytest ultraplot/tests/test_inputs_helpers.py` and an end-to-end `ax.bar(xr.DataArray)` smoke check.

### DIFF
--- a/ultraplot/internals/inputs.py
+++ b/ultraplot/internals/inputs.py
@@ -686,7 +686,7 @@ def _meta_coords(*args, which="x", **kwargs):
         if data.ndim > 1:
             raise ValueError("Non-1D string coordinate input is unsupported.")
         ticks = np.arange(len(data))
-        labels = list(map(str, data))
+        labels = list(map(str, _to_numpy_array(data)))
         kwargs.setdefault(which + "locator", Locator(ticks))
         kwargs.setdefault(which + "formatter", Formatter(labels, index=True))
         kwargs.setdefault(which + "minorlocator", Locator("null"))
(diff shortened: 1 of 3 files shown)


cvanelteren:
I will fix the zenodo stuff; it's not related to this PR

cvanelteren:
Error will fix itself. Thanks @kinyatoride

PR #713, opened 2026-05-04
Fix PolarAxes.format() silently dropping xlabel/ylabel

kinyatoride:
## Summary

`ax.format(xlabel=..., ylabel=...)` had no effect on polar axes (e.g. for Taylor-diagram-style figures), even though `ax.set_xlabel(...)` worked. `PolarAxes.format()` did not declare `xlabel`/`ylabel` parameters and `base.Axes.format()` does not handle them — only `CartesianAxes.format()` did.

This PR:

- Adds an `_update_labels` helper to `PolarAxes` mirroring the `CartesianAxes` one, and wires `xlabel`, `ylabel`, `xlabel_kw`, `ylabel_kw` through the existing `(x, y)` loop so they are forwarded to matplotlib's `set_xlabel`/`set_ylabel`.
- Fixes a dispatcher limitation in `Figure.format()` exposed by the change: `_pop_params` previously popped each kwarg into the first matching class signature, so when called via `SubplotGrid.format` → `Figure.format` on a polar axes, `xlabel` was routed to `kws[CartesianAxes]` and never applied. Switched to a non-popping collector so shared params reach every matching class signature; the unused-warning is updated so shared keys consumed by any matched class don't trigger a false warning.
- Adds a regression test in `test_projections.py`.

## Test plan

- [x] `pytest ultraplot/tests/test_projections.py -k polar` — new and existing polar tests pass
- [x] `pytest ultraplot/tests/test_format.py` — no regressions
- [x] Full suite: 523 passed (1 unrelated failure in `test_docs_fetch_releases`)
- [x] Smoke test: `ax.format(xlabel='Radius', ylabel='Angle', labelpad=15, labelcolor='red', xlabel_kw={'weight': 'bold'})` renders correctly with per-axis-kw override

### DIFF
--- a/ultraplot/axes/polar.py
+++ b/ultraplot/axes/polar.py
@@ -82,6 +82,13 @@ thetalabels, rlabels : optional
 thetaformatter_kw, rformatter_kw : dict-like, optional
     The azimuthal and radial label formatter settings. Passed to
     `~ultraplot.constructor.Formatter`.
+xlabel, ylabel : str, optional
+    The x and y axis labels. Applied with `~matplotlib.axes.Axes.set_xlabel`
+    and `~matplotlib.axes.Axes.set_ylabel`.
+xlabel_kw, ylabel_kw : dict-like, optional
+    Additional axis label settings applied with `~matplotlib.axes.Axes.set_xlabel`
+    and `~matplotlib.axes.Axes.set_ylabel`. See also `labelpad`, `labelcolor`,
+    `labelsize`, and `labelweight`.
 color : color-spec, default: :rc:`meta.color`
     Color for the axes edge. Propagates to `labelcolor` unless specified
     otherwise (similar to :func:`~ultraplot.axes.CartesianAxes.format`).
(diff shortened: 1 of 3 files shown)


cvanelteren:
Out of town atm can look at this when I'm back.

cvanelteren:
My initial read on this was that xlabel and ylable makes no sense for `PolarAxes`but as I understand it you want to work towards making a Taylor diagram directly which you would then have x and ylabels which are currently being filtered out. Am I correctin this?

cvanelteren:
If yes then my issue would be that Taylor diagrams probably need a new axis type as it does not make sense to have x and y labels for polar axes (as they don't exist).

kinyatoride:
Yeah, I agree. I should not have included this in the same PR. I originally started from xlabel/ylabel and realized later rlabel/thetalabel make more sense. So, I am okay dropping this. I think we can suggest to use rlabel/thetalabel if users pass xlabel/ylabel.

cvanelteren:
Closeing this as we need to implement the specific axis and this does not pertain to polar.

PR #715, opened 2026-05-06
Suppress sharing warnings when no sharing is possible

kinyatoride:
# Suppress sharing warnings when no sharing is possible

## Problem

`UltraPlotWarning`s about axis sharing fired in cases where the user did not request sharing or where sharing was not possible at all:

- `uplt.subplots(proj='polar', ncols=2, nrows=3, share=0)` — the user explicitly disabled sharing, but construction and every subsequent `draw` emitted `Skipping incompatible y-axis sharing for PolarAxes and PolarAxes: different axis families.` and `Tick label sharing not implemented for <class 'ultraplot.axes.polar.PolarAxes'> subplots.`
- `uplt.subplot(proj='polar')` / `uplt.subplots(proj='polar')` — a single polar axis has nothing to share with, but `Tick label sharing not implemented for <PolarAxes> subplots.` still fired four times per draw.

## Fix

Two unrelated paths needed gating:

- **`Axes._apply_auto_share`** (`ultraplot/axes/base.py`) called `Figure._partition_share_axes` for x and y unconditionally. Wrapped the external-share blocks in `if self.figure._sharex > 0:` / `if self.figure._sharey > 0:`, mirroring the existing panel-share guards just above.

- **`Figure._share_ticklabels`** (`ultraplot/figure.py`) ran `_compute_baseline_tick_state` (which emits the `Tick label sharing not implemented` warning for unsupported axes types) before checking whether the group could actually share. Skip the group up front when it has fewer than two axes or when no axes have effective share level >= 3.

Explicit `share='all'` (or any level >= 1) on multiple polar / mixed cart-polar grids still warns as before — that case is informative.

## Tests

Added regression tests in `ultraplot/tests/test_figure.py`:

- `test_share_zero_polar_emits_no_warnings` — `share=0` on an all-polar grid.
- `test_share_zero_mixed_cartesian_polar_emits_no_warnings` — `share=0` on a mixed cart/polar grid.
- `test_share_default_single_polar_emits_no_warnings` — default share on a single-axis polar figure via `uplt.subplots`.
- `test_share_default_single_polar_subplot_singular_emits_no_warnings` — same via `uplt.subplot` (singular).

Existing `test_explicit_share_warns_for_mixed_cartesian_polar` continues to pass, confirming the explicit-share warning path is preserved. All 23 share-related tests pass; broader `test_figure.py` / `test_subplots.py` / `test_projections.py` (139 tests) also pass.

### DIFF
--- a/ultraplot/axes/base.py
+++ b/ultraplot/axes/base.py
@@ -1347,22 +1347,24 @@ class Axes(_ExternalModeMixin, maxes.Axes):
 
         # External axes sharing, sometimes overrides panel axes sharing
         # Share x axes within compatible groups
-        axes_x = self._get_share_axes("x")
-        for group in self.figure._partition_share_axes(axes_x, "x"):
-            if not group:
-                continue
-            parent, *children = group
-            for child in children:
-                child._sharex_setup(parent)
+        if self.figure._sharex > 0:
+            axes_x = self._get_share_axes("x")
+            for group in self.figure._partition_share_axes(axes_x, "x"):
+                if not group:
+                    continue
+                parent, *children = group
+                for child in children:
+                    child._sharex_setup(parent)
 
         # Share y axes within compatible groups
-        axes_y = self._get_share_axes("y")
-        for group in self.figure._partition_share_axes(axes_y, "y"):
-            if not group:
-                continue
-            parent, *children = group
-            for child in children:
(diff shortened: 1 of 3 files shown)


cvanelteren:
Thanks!

PR #721, opened 2026-05-12
Fix tick visibility leaking from styles in alternative axes

cvanelteren:
When applying custom styles like dark_background, matplotlib's styling dictionaries often include explicit boolean flags for tick visibility on all sides (e.g. left, right, top, bottom). Previously, UltraPlot queried these style settings using _get_tickline_props and inadvertently reapplied them to the axes, completely overriding UltraPlot's internal locators. This caused bugs such as the left axis displaying right ticks and the right axis displaying left ticks when a style was active. By aggressively popping visibility keys out of the retrieved styling dictionary, we ensure that themes strictly govern visual appearances (color, width, padding) without hijacking the structural tick visibility correctly managed by the format methods.

### DIFF
--- a/ultraplot/tests/test_axes_alt_styles.py
+++ b/ultraplot/tests/test_axes_alt_styles.py
@@ -1,0 +1,45 @@
+import pytest
+import ultraplot as uplt
+
+
+def test_alt_axes_styling_dark_background():
+    """
+    Test that applying dark_background style does not leak tick visibility
+    settings and correctly preserves alternative axes tick locations.
+    """
+    with uplt.rc.context(style="dark_background"):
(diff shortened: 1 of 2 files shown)


ultraplot/tests/test_axes_alt_styles.py
munechika-koyo, line 1: Should we check whether each axis's color is correct, not just visibility?
Checking `t.get_color() == #xxxxxx`, `ax.yaxis.label.get_color() == #xxxxxx`, `ax.spines["left"].get_edgecolor() == (x, x, x, 1.0)`,  etc. should be useful.

Issue #726, opened 2026-05-14
Proposal: Unify the parameters for semantic legend functions like `catlegend` and `sizelegend`...


gepcel:
I've been using `sizelegend` and `catlegend` recently and ran into a few issues. I initially intended to fix them myself and submit a PR, but as I continued testing, I became increasingly confused. So, I'm writing this as a suggestion for your reference instead.

The parameters for `sizelegend` and `catlegend` are currently inconsistent, mixing singular and plural forms. For example, `catlegend` accepts both `colors` and `color`: the former takes a list (mapping to every entry), while the latter accepts a single color applied to all entries. However, it only accepts `facecolor` (whose value can be a list for every entry) but does not accept the plural form `facecolors`.  And `sizelegend` only accept `color` but not `colors`. I tested a portion of this and did some source code digging, but I quickly got lost in the value parser and argument check functions.

If my explanation isn't clear enough, I suggest you test the following parameters separately for `catlegend` and `sizelegend`:
`color`, `colors`, `facecolor`, `facecolors`, `markerfacecolor`, `markerfacecolors`, `edgecolor`, `markeredgecolor`, `markeredgewidth`, etc.

**Suggestions:**

1. There is already function like `_style_lookup` that matches a list (or dict) of values to every entry. I suggest using strictly singular parameters and dropping the plural forms like `colors`, `sizes`.
2. Semantic legend functions support all marker arguments, such as `size`,  `markersize`, `color`, `facecolor`, `edgecolor`, `markerfacecolor`, `markeredgecolor`, `markeredgewidth`, `alpha`, etc.
3. Support all argument aliases, such as: `ms`, `c`, `mfc`, `mec`, `fc`, `ec`, `mew`, etc
4. Ensure all arguments about marker styles support both (now some args like labels and colors does, and some like edgecolor doesn't): 
    * a single value applied to all entries, 
    * a list (or dict) of values applied to each individual entry.

gepcel:
Today, I have some spare time, so I’ll try to put something together and submit a pull request for your consideration. Please keep in mind that I’m not a professional programmer; I just use Python for data analysis. So, I completely understand if you find my submission unacceptable. However, I would be honored if you could at least consider it and help me get the PR into shape.

gepcel:
An example with pr #727 

<img width="421" height="421" alt="Image" src="https://github.com/user-attachments/assets/70b01579-4400-4710-baf1-fa2112da253b" />

```python
[...]

PR #727, opened 2026-05-18
Unify semantic legend params.

gepcel:
Addresses #726


gepcel:
Try something.  

pre-commit.ci autofix

gepcel:
There are at least two existing issues with this PR:

1. Arguments can be passed through full names, aliases, or within handle_kw, kwargs and other parameters, which occasionally causes argument value overrides.
2. Regarding color-related parameters such as color, fc, ec and so on, I have not yet figured out a clear judgment logic. It should determine whether a value is a list of color values lthat should be applied separately to each entry, or whether list-type tuple values need to be converted into a single unified value. I suspect Ultraplot already has built-in logic to handle this judgment, but I have not located the relevant code implementation yet.

gepcel:
call for 

pre-commit.ci autofix

cvanelteren:
Thanks for the initiative @gepcel. Took a pass at tightening this up. First commit is the substantive cleanup of the legend rewrite — the stray print() in _is_color_like is gone, lists of three or four floats now work as a single color the same way tuples do (so color=[0.5, 0.5, 0.5] stops crashing), and a dead branch in _style_lookup that returned the same value either way is collapsed. The eight numbered steps inside _pop_entry_props are split into three small helpers that _pop_num_props reuses, the # without this, line 645 of test_legend.py won't pass comment is replaced with one that actually explains what it's guarding, and the misleading ec/fc lines come out of _LINE_ALIAS_MAP since they already work through _pop_props. The second commit is just renaming test_sematic_legend.py to fix the typo. The third commit adds real docs: a few shared snippets in legend.py feed numpy-style docstrings on the five public Axes.*legend methods, so the aliasvocabulary and per-entry rules live in one place instead of five. 

I hope this helps you on their way, and I think this addresses most of the issues you outlined. Let me know!

gepcel:
I agree.

cvanelteren (approved):

cvanelteren:
Leave it up to you if we can merger @gepcel

gepcel:
> Leave it up to you if we can merger @gepcel

I think it's ready to merge.

PR #738, opened 2026-06-19
Allow colorbars to attach to inset axes

### LINKED ISSUE #737: [Feature request] inset colorbars
Colorbars are currently not supported on insets; we should change that!

cvanelteren:
Closes #737 
Colorbars requested on inset axes with side locations now stay attached to the inset instead of trying to allocate a subplot panel, which fails for axes without a subplot spec. The change remaps side requests to inset colorbar placements with side-appropriate default orientations and adds a regression test covering both direct inset-axis calls and figure-level calls that target an inset axis.

### DIFF
--- a/ultraplot/axes/base.py
+++ b/ultraplot/axes/base.py
@@ -726,11 +726,12 @@ class _TransformedBoundsLocator:
 class _SideColorbarLocator:
     """Position a side colorbar beyond its parent axes decorations."""
 
-    def __init__(self, parent, side, bounds, pad):
+    def __init__(self, parent, side, bounds, pad, previous=()):
         self._parent = parent
         self._side = side
         self._bounds = bounds
         self._pad = pad
+        self._previous = tuple(previous)
 
     def __call__(self, ax, renderer):
         parent = self._parent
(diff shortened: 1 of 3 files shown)


cvanelteren:
test

Example

```python
import numpy as np
import ultraplot as uplt

data = np.random.rand(10, 10)
x = 0.35
fig, ax = uplt.subplots()
inax = ax.inset_axes([0.55, 0.45, x, x])
inax.pcolormesh(
    data,
    colorbar="t",
    colorbar_kw=dict(width=0.85),
)
```

gepcel (approved):
I haven't had time to test this yet—don't let me hold up the process.

cvanelteren:
There is no rush @gepcel

cvanelteren:
I will merge it by Friday if I don't here anything fyi @gepcel if you have time before to test stuff then would love to get your feedback.

ultraplot/axes/base.py
```diff
@@ -671,6 +671,58 @@ def _align_bbox(align, length):
     return mtransforms.Bbox(bounds)
 
 
+def _get_side_colorbar_ticklocation(side, orientation, tickloc, ticklocation):
+    """Return the outward-facing tick location for a side colorbar."""
+    if orientation == "horizontal":
+        default = "top" if side == "top" else "bottom"
```
gepcel, line 681: If orientation is horizontal, I think the default should be "bottom".

gepcel: Because both ultraplot and matplotlib use bottom as default position for horizontal colorbar. Other than this, I'm ok with this.

cvanelteren: ah good catch will do that.

gepcel: You can merge when you think it's ready. This is an execellent feature.

cvanelteren:
Thanks @gepcel

PR #749, opened 2026-07-02
Automatic inference for semantic legends for marker/scatter size

cvanelteren:
How about this @gepcel ?

This PR makes sizelegend line up with scatter marker sizing without forcing users to manually reconstruct UltraPlot’s size-scaling rules. Scatter collections now retain their effective size-scale metadata, and sizelegend() can reuse that information by default to infer matching legend marker sizes from an existing compatible scatter plot, while still allowing explicit overrides and preserving direct area=True / area=False behavior when requested. The change also adds regression coverage for absolute- size, scaled-size, compatibility fallback, and explicit override cases so the legend behavior stays predictable.

### DIFF
--- a/ultraplot/axes/base.py
+++ b/ultraplot/axes/base.py
@@ -3835,6 +3835,27 @@ class Axes(_ExternalModeMixin, maxes.Axes):
             Treat ``levels`` as marker areas (``True``, default) or
             diameters (``False``). Areas are converted with
             ``ms = sqrt(level) * scale``. Falls back to :rc:`legend.size.area`.
+        values : array-like, optional
+            Full scatter-size data used to infer the scaling range for
+            ``levels``. When provided, or when any of ``vmin``, ``vmax``,
+            ``smin``, ``smax``, ``area_size``, or ``absolute_size`` are
+            provided, ``levels`` are transformed with the same size scaling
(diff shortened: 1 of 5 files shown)


cvanelteren:
@gepcel 

`ax.plot(..., m='.', ms=50)` creates a Line2D point with marker '.' and markersize=50, but your `sizelegend([50, 25], area=False)` is building semantic legend entries with the default marker 'o'. Numerically the first legend handle is 50 pt, but it is marker='o', not marker='.', so it renders differently and looks wrong for that plot.

gepcel:
> @gepcel
> 
> `ax.plot(..., m='.', ms=50)` creates a Line2D point with marker '.' and markersize=50, but your `sizelegend([50, 25], area=False)` is building semantic legend entries with the default marker 'o'. Numerically the first legend handle is 50 pt, but it is marker='o', not marker='.', so it renders differently and looks wrong for that plot.

Oh, despite using matplotlib for a long time, I didn't realize '.' and 'o' are different markers. I mistakenly thought they were the same. My bad! Sorry to bother.

cvanelteren:
No bother, but is this PR more what you had in mind? Then can proceed to merge and draft a release.
EDIT: I am not that opinionated in this area as I don't really use these kind of legends, so I take your feedback to hearth.

gepcel (approved):

PR #754, opened 2026-07-13
[Feature] Text alignment

### LINKED ISSUE #753: Overlapping elements
I run into an issue a lot where annotation are overlapping a lot. This creates a loop where I make small edits unless it "looks" right. I think UltraPlot should add a "looks good by default" option for plots with lots of texts to prevent overlaps while still allowing manual overrides when the users wants to put at exactly position x,y etc.

cvanelteren:
Closes #753 for text alignment

I implemented a KD tree approach to determine the closest neighbor between text elements to prevent overlap. How this works is that by default it would keep track of texts and aligns them on the final draw. Nothing is registered unless you pass `avoid_overlap=True` (or flip `rc['text.align']`), so the solver never even imports on an ordinary figure. The naive approach would be to do this with O(n^2) pairwise comparison but that scales poorly and is not necessary.

A plot with 30 labeled texts with run in about 80ms compared to 50ms natively without any draws. This approach will scale not properly for large number of labelled elements (say > 500) but I don't think that will ever occur for plots as it will be unreadable anyway.


cvanelteren:
~~TODO: add tests~~

gepcel (approved):

cvanelteren:
Merging this on main to test it out for the next few days. It seems fine as is but still a bit hesitant. Have a few papers coming up where this would have been a god send.

Issue #758, opened 2026-07-14
Remove dead imports in figure.py left over from SubplotManager extraction


cvanelteren:
Trivial cleanup left over from #698. Three imports in `ultraplot/figure.py` are now referenced only by their own import statements — the code that used them moved to `ultraplot/_subplots.py`:

- `from numbers import Integral` (line 9)
- `import matplotlib.gridspec as mgridspec` (line 20)
- `import matplotlib.projections as mproj` (line 21)

```console
$ grep -n '\bIntegral\b\|\bmgridspec\b\|\bmproj\b' ultraplot/figure.py
9:from numbers import Integral
20:import matplotlib.gridspec as mgridspec
21:import matplotlib.projections as mproj
```

Safe to delete.

Issue #757, opened 2026-07-14
SubplotManager is not standalone: bidirectional coupling with Figure


cvanelteren:
Design follow-up from #698 / #677. Not a bug — the refactor is behaviour-preserving and the full suite is green. This is about whether the seam is where the plan says it is.

The manager sequencing in #677 positions `SubplotManager` as the leaf node: *"SubplotManager: (standalone, only needs gridspec)"*. As extracted, it isn't standalone.

### The dependency is bidirectional

`SubplotManager` reaches back into `Figure` for:

- `fig._layout_dirty`
- `fig._format_signature`
- `fig.add_subplot(...)`
- `fig.format(...)`
- a function-local `from .figure import Figure` (`_subplots.py:256`) needed to call `super(Figure, fig).add_subplot(...)` — a deferred import that exists purely to break an import cycle.

And `Figure`'s collaborators now call *back* through the figure to reach the manager: `gridspec.py` and `axes/base.py` use `fig._get_subplot` / `fig._iter_subplots`.

There's also a round trip: `SubplotManager.add_subplots` calls `fig.add_subplot(...)` (the public, decorated method) rather than `self.add_subplot(...)`, so every subplot goes manager → figure → manager just to pick up `@_clear_border_cache`.

Net: the code moved to a new file, but the coupling didn't shrink. That's fine as a first step — it just means the "standalone" edge in the #677 dependency graph isn't real yet, and the managers that are supposed to depend *on* this one will inherit the cycle.

### Bonus: `parse_proj` doesn't belong here

`SubplotManager.parse_proj` never uses `self` except to call `self.parse_backend(...)`, which is itself a `@staticmethod`. Projection parsing is a pure function of its arguments, and the class docstring already admits it's a third concern (*"subplot creation, gridspec ownership, and projection parsing"*).

Making it a `@staticmethod` or a module-level function would be a strict simplification — and it also makes the fix for the `ui.py` signature coupling (filed separately) trivial.

Issue #756, opened 2026-07-14
Figure.clear() leaves stale subplot state (subplotgrid returns dead axes)


cvanelteren:
**Pre-existing bug — reproduces identically on `main`, not caused by #698.** Filing it because #698 is what gives this state a named owner (`SubplotManager`), which makes it the natural place to fix.

### Repro

```python
import ultraplot as uplt
fig, axs = uplt.subplots(nrows=1, ncols=2)
fig.clear()

print(fig.axes)                          # []          <- matplotlib cleaned up
print(list(fig._subplots.subplot_dict))  # [1, 2]      <- stale, dead axes
print(len(fig.subplotgrid))              # 2           <- hands them back to the user
print(fig.gridspec)                      # GridSpec(nrows=1, ncols=2)  <- stale
print(fig._subplots.counter)             # 2           <- never reset
```

`Figure.clear()` goes through matplotlib and empties `fig.axes`, but nothing resets the subplot bookkeeping. So after a `clear()`:

- `fig.subplotgrid` returns axes that are no longer in the figure,
- `fig._get_subplot(n)` resolves to a dead axes,
- `fig.gridspec` still points at the old gridspec,
- the label counter keeps climbing.

(On `main` the same thing happens via `fig._subplot_dict` / `fig._gridspec` / `fig._subplot_counter` — this is a faithful description of long-standing behaviour, not something the refactor introduced.)

### Suggested fix

Give `SubplotManager` a `reset()` that clears `subplot_dict`, zeroes `counter`, and drops `_gridspec`, and have `Figure.clear()` call it. Add a regression test — there is currently no test covering figure state after `clear()`, which is why this survived.

Issue #755, opened 2026-07-14
ui.subplot silently mis-routes projection kwargs if Figure._parse_proj's signature drifts


cvanelteren:
Found while reviewing #698 (SubplotManager extraction). **Not a regression — the signatures are correct today.** This is about a trap that has already fired once.

### The coupling

`ultraplot/ui.py` uses `_pop_params` to decide which keyword arguments belong to the subplot and which belong to the figure, and it does so by **introspecting the signature of two `Figure` methods**:

```python
# ui.py:185
kwsub.update(_pop_params(kwargs, pfigure.Figure._parse_proj))
# ui.py:230
kwsubs.update(_pop_params(kwargs, pfigure.Figure._add_subplots))
```

After #698, `Figure._parse_proj` and `Figure._add_subplots` are pure delegators to `SubplotManager`. Their bodies do nothing but forward. Their *only* remaining job is to carry a parameter list that mirrors the manager's — a contract enforced by nothing but a docstring `NOTE:`.

### Why this is not hypothetical

It already broke. During #698, an earlier commit collapsed `_parse_proj` to `(*args, **kwargs)`. `_pop_params` then saw no `proj`/`projection`/`proj_kw` parameters, routed `proj=` to the **figure** instead of the subplot, and `uplt.subplot(proj='polar')` silently raised from `Figure.set()`. Commit `537f794` fixed it by spelling the parameters back out.

So the failure mode is: a reasonable-looking cleanup of a method that appears to be a trivial pass-through silently breaks projection handling, with no type error and no local test failure.

### Suggested fix

Point `_pop_params` at the functions that actually own the parameters, so the signature and its meaning live in one place:

```python
from ._subplots import SubplotManager
kwsub.update(_pop_params(kwargs, SubplotManager.parse_proj))
kwsubs.update(_pop_params(kwargs, SubplotManager.add_subplots))
```

`Figure._parse_proj` / `._add_subplots` can then go back to thin `(*args, **kwargs)` delegators and the drift hazard disappears entirely.

This matters more as the rest of the manager split in #677 lands — each future extraction is exactly the kind of change that steps on this.

PR #759, opened 2026-07-14
Chore: refactor figure with new subplot manager (#698)

cvanelteren:
Move subplot creation, gridspec ownership, and projection parsing out of `Figure` and into a dedicated `SubplotManager` (`ultraplot/_subplots.py`), a first step toward treating `Figure` as an interface that delegates to focused collaborators rather than a monolith (partially addresses #677). The public API is unchanged: the former `_subplot_dict`, `_subplot_counter` and `_gridspec` attributes now live on the manager and are reached through the new `Figure._get_subplot()` / `Figure._iter_subplots()` accessors and the existing `Figure.gridspec` property, with call sites in `gridspec.py` and `axes/base.py` going through those instead of touching figure internals. Also fixes projection keywords being dropped by `uplt.subplot()`: `ui.subplot` introspects the signature of `Figure._parse_proj` to decide which keywords belong to the subplot rather than the figure, so that signature must stay spelled out rather than collapsing into `**kwargs`. Adds `ultraplot/tests/test_subplot_manager.py` covering subplot creation, the gridspec setter, projection resolution (native, external-container, and geographic), per-axes projection arguments, and the `ui.subplot`/`subplots` keyword routing. Follow-ups filed: #755 (signature coupling in `ui.py`), #756 (`Figure.clear` leaves stale subplot state), #757 (`SubplotManager`/`Figure` coupling), #758 (dead imports).

### DIFF
--- a/ultraplot/_subplots.py
+++ b/ultraplot/_subplots.py
@@ -1,0 +1,386 @@
+"""
+Subplot creation and management for ultraplot figures.
+"""
+
+from numbers import Integral
+from typing import TYPE_CHECKING
+
+import matplotlib.axes as maxes
+import matplotlib.gridspec as mgridspec
+import matplotlib.projections as mproj
+import numpy as np
+
+from . import axes as paxes
+from . import constructor
+from . import gridspec as pgridspec
+from .internals import _not_none, _pop_params, warnings
+
+if TYPE_CHECKING:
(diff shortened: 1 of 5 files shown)


cvanelteren:
Closing the v3.0 branch as it only contained this change anyway.

PR #765, opened 2026-07-15
Reorder visual hierarchy when side-cap labels are given

### LINKED ISSUE #764: Side labels create unclear visual hierarchy
When a spanning y-axis label is used together with categorical row labels, UltraPlot places the shared y-axis label between the row labels and the subplot grid. This pushes labels such as “Cropland,” “Forest,” “Grassland,” and “Savanna” farther away from their corresponding rows, creating an unclear visual hierarchy. The shared y-axis label should preferably be outermost, while the categorical row labels should remain closer to their panels. This behavior is reproducible in plain UltraPlot and is unaffected by the order of the formatting calls.

Minimal Reproducible Example
```
import ultraplot as uplt

fig, axs = uplt.subplots(
    nrows=4,
    ncols=4,
    sharex=True,
    sharey=True,
)

axs.format(
    xlabel="Observed value",
    ylabel="Predicted value",
)

fig.format(
    rowlabels=[
        "Cropland",
        "Forest",
        "Grassland",
        "Savanna",
    ],
)

plt.show()

```

_Originally posted by @lwq-star in https://github.com/Ultraplot/UltraPlot/discussions/747#discussioncomment-17643884_

cvanelteren:
Closes #764 
This PR changes the behavior of sidelabels. Now, the side caption labels sit closest to the y labels so that a shared ylabel now has a clearer visual hierarchy. During this process an additional bug was found relating to the ylabels and dpi; this is now figure relative.

### DIFF
--- a/docs/subplots.py
+++ b/docs/subplots.py
@@ -395,6 +395,14 @@ for (name1, equal), (name2, group) in zip(equals, groups):
 #    Explicit sharing levels still force sharing attempts and may warn when
 #    incompatible axes are encountered.
 #
+# When figure-level row or column labels and a shared spanning axis label use
+# the same side, the row or column labels are placed nearer the axes and the
+# spanning label is placed outside them. Control this gap with the
+# :rc:`leftlabel.sharedpad`, :rc:`rightlabel.sharedpad`,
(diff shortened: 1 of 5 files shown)


cvanelteren:
Example: 
image

PR #766, opened 2026-07-15
Fix centering of titles

cvanelteren:
Fix an issue where the added abc logic was interferring with a shift in the title.


AbsoluteVacuum:
Thank you so much! I had such a huge issue with this. Happy that it's finally fixed!

PR #767, opened 2026-07-15
[Feature] Abc anchor and Geo aspect

cvanelteren:
Mixed plots with GeoAxes and normal cartesian plots are a bit cumbersome to deal with. The major issue is that cartopy will adjust the subplot dimension such that it matches the aspect ratio of the plot. This makes the subplot look good, but runs into issues when we have multiple subplot kinds within one figure. This PR will introduce 2 new options.

With `abcanchor`  users can now override the default behavior of putting the abc in the top left corner, by stating the alignment point.  For example, setting `abcanchor="slot"` will set the anchor mode to be within the "slot" of the layout rather than the axes as a reference point. 

image

In addition, aspect can now be set on geo axes  (i.e. `aspect = "auto"`); this aids in forcing aspect to match the rest of the plot rather than the reshaping performed by cartopy.

I think these features together will give users within geoscience more power and flexibility. 

```
import ultraplot as uplt

layout = [[1, 2], [3, 4]]
fig, axs = uplt.subplots(
    layout,
    proj={4: "cyl"},
    share=False,
)
axs.format(abc = True)
axs.format(land = True, landcolor = "k", oceancolor = 'ocean blue', ocean = True, abcanchor = "slot")
```

</details>

### DIFF
--- a/docs/projections.py
+++ b/docs/projections.py
@@ -517,6 +517,60 @@ fig.format(
 
 
 # %% [raw] raw_mimetype="text/restructuredtext"
+# .. _ug_geo_layout:
+#
+# Geographic axes in mixed subplot layouts
+# -----------------------------------------
+#
+# Geographic axes preserve the aspect ratio required by their projection. This is
+# important for interpreting distances and shapes correctly, but it creates a layout
+# choice when a map shares a GridSpec with Cartesian axes or maps that have a different
+# natural aspect ratio. A fixed-aspect map cannot both fill a differently shaped subplot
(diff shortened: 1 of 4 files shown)


gepcel (approved):

munechika-koyo:
I couldn't handle this review because of recent busy weeks.
I sincerely apologize for not reviewing this PR.

cvanelteren:
> I couldn't handle this review because of recent busy weeks.
> 
> I sincerely apologize for not reviewing this PR.
 

No worries!

PR #771, opened 2026-07-17
[Feature] Add Hawkeye option

### LINKED ISSUE #770: Suggestion: improve inset positioning for geoaxes
### Why this?

Inset axes are widely used in plots, both for Cartesian Axes and for GeoAxes. In map contexts, this is sometimes called a *hawkeye map*.

While `ax.inset` works well for Cartesian axes, it does not handle geoaxes correctly. The reason is that geoaxes enforce a fixed aspect ratio determined by the map projection. As a result, the standard inset positioning logic fails, as reported in issue [#751](https://github.com/Ultraplot/UltraPlot/issues/751).

I have a small helper function that works around this for personal use, but I'm not sure whether this approach fits UltraPlot's design philosophy. If it does, **Could you take a look? And see how it might be able to integrated properly, following UltraPlot's code style.**

### How

1. Ignore the explicitly set width and height for geoaxes (though maybe there is a way to enforce at least a minimum width that I am unaware of).
2. Recalculate the position, width, and height from the desired anchor point, and then reset the inset position.

Here is a minimal example using the helper function:

```python
import cartopy.crs as ccrs

fig, ax = plt.subplots(1, 1, subplot_kw={'projection': ccrs.PlateCarree()})
ax.coastlines()

# iax.set_position doesn't work for inset_exes. Don't know why.
# iax = ax.inset_axes([0.8, 0.8, 0.2, 0.2], projection=ccrs.PlateCarree())
iaxs = [fig.add_axes([0.8, 0.8, 0.1, 0.2], projection=ccrs.PlateCarree()) for i in range(10)]

# for the first 9 iax, use loc string as anchor:
for iax, pos, anchor in zip(iaxs[:-1],
    [(0,1), (.5, 1), (1,1), (0,.5), (.5,.5), (1,.5), (0,0), (.5, 0), (1, 0)],
    ['ul', 'uc', 'ur', 'cl', 'c', 'cr', 'll', 'lc', 'lr']):
    iax.set_extent([68, 70, 22, 24])
    iax.coastlines()
    iax.spines['geo'].set_edgecolor('red')
    iax.text(.5, .5, anchor, ha='center', va='center', fontsize=24, color='r', transform=iax.transAxes)
    update_inset_position(pos[0], pos[1], ax, iax, anchor=anchor, anchor_transform='data')

# For the last iax, use (x,y) as for anchor:
iaxs[-1].set_extent([68, 70, 22, 24])
iaxs[-1].coastlines()
iaxs[-1].spines['geo'].set_edgecolor('blue')
iaxs[-1].spines['geo'].set_linewidth(2)
update_inset_position(68, 22, ax, iaxs[-1], transform='data', anchor=(68, 22), anchor_transform='data')
iaxs[-1].text(.5, .5, '(68,22)', ha='center', va='center', fontsize=16, color='b', transform=iaxs[-1].transAxes)
plt.show()
```

And here is the resulting image:

Image

### Future enhancements which the helper func doesn't include.

1. **Indicator rectangle** – automatically place a rectangle on the main axes (or the inset) to indicate the region that is magnified, depending on which of the two is zoomed in.
2. **Indicator lines** – add connecting lines between the inset and the main axes, if necessary.

### The helper function:

```python
# Copy the method from ultraplot here, to work with matplotlib and cartopy.
def _get_transform(ax, transform, default="data"):
    """
    Translates user input transform. Also used in an axes method.
    """
    # TODO: Can this support cartopy transforms? Seems not when this
    # is used for inset axes bounds but maybe in other places?
    from ultraplot.internals import _not_none
    from matplotlib import transforms as mtransforms
    from cartopy.crs import CRS, PlateCarree
    transform = _not_none(transform, default)
    if isinstance(transform, mtransforms.Transform):
        return transform
    elif CRS is not object and isinstance(transform, CRS):
        return transform
    elif PlateCarree is not object and transform == "map":
        return PlateCarree()
    elif transform == "data":
        return ax.transData
    elif transform == "axes":
        return ax.transAxes
    elif transform == "figure":
        return ax.figure.transFigure
    elif transform == "subfigure":
        return ax.figure.transSubfigure
    else:
        raise ValueError(f"Unknown transform {transform!r}.")


def update_inset_position(x, y, ax, iax, transform='axes', anchor='ur', anchor_transform='iax'):
    '''
    Update the position of the inset axes based on position(x, y) and anchor.
    Place anchor point of iax in the position (x, y) of ax.

    Args:
      x, y: float, point position to put iax on ax.
      transform: str, 'axes' or 'data', default 'axes'.
      anchor: str or tuple of float.
              If str, must be one of: ul, ur, ll, lr, c, uc, lc, cl, cr, or "upper right", "upper left", "lower right", "lower left", "center", "center left", "center right", "upper center", "lower center".
              If tuple, must be (x, y) of iax.
      trnasform: str or matplotlib transform object, the transform of x, y, default to "axes", aka: ax.transAxes
      anchor_transform: str or matplotlib transform object, the transform of anchor point, default to "iax", aka: iax.transAxes
    '''
    plt.draw()
    anchor_position_loc = {
        'ul': (0, 1), 'upper left': (0, 1),
        'ur': (1, 1), 'upper right': (1, 1),
        'll': (0, 0), 'lower left': (0, 0),
        'lr': (1, 0), 'lower right': (1, 0),
        'c': (0.5, 0.5), 'center': (0.5, 0.5),
        'uc': (0.5, 1), 'upper center': (0.5, 1),
        'lc': (0.5, 0), 'lower center': (0.5, 0),
        'cl': (0, 0.5), 'center left': (0, 0.5),
        'cr': (1, 0.5), 'center right': (1, 0.5),
    }
    # get position of iax
    if isinstance(anchor, str):
        if anchor not in anchor_position_loc:
            raise ValueError(f'anchor must be one of {list(anchor_position_loc.keys())}, got {anchor}')
        anchor = anchor_position_loc[anchor]
        # if anhcor is string, ignore anchor_transform.
    else:
        anchor = tuple(anchor) # expect (x, y)

        # for anchor_transform, The 'iax' shortcut means "use the inset axes' own transAxes".
        # Since _get_transform does not recognize 'iax', we map it to 'ax'
        # and pass the inset axes (iax) as the axes argument.
        # This yields iax.transAxes, which places the anchor in
        # the inset axes' normalized coordinate system (0-1).
        anchor_transform = 'axes' if anchor_transform == 'iax' else anchor_transform
        anchor_transform = _get_transform(iax, anchor_transform)
        anchor = iax.transAxes.inverted().transform(anchor_transform.transform(anchor))

    position_ax = ax.get_position()
    position_iax = iax.get_position()
    invert_fig_trans = ax.figure.transFigure.inverted()
    x, y = _get_transform(ax, transform).transform((x, y))
    x, y = invert_fig_trans.transform((x, y))
    x -= anchor[0] * position_iax.width
    y -= anchor[1] * position_iax.height
    # trans_ax.transform(x,y)
    iax.set_position([x, y, position_iax.width, position_iax.height])
```

cvanelteren:
Closes #770 

This PR extends the options for annotation on `GeoAxes`. It add specific "hawkeye" insets for GeoAxes to highlight specific areas of the map. It both adds square as well as circular annotations. Examples:

test2

test1

Succinctly this is now wired in as:

  - Scalar or (width, height) size in parent-axes fractions.
  - Anchoring with xy, transform, and anchor.
  - Geographic extent plus automatic detail/overview relation inference.
  - Extent indicators and either corner connectors or a single leader line.
  - Rectangular or circular inset/target shapes.
  - Projection-preserving circular callouts that pad the shorter map dimension;
    aspect="auto" instead fits an exact extent with distortion.
  - Optional alternative inset projection via proj=.
  - Gridlines disabled by default.

### DIFF
--- a/ultraplot/axes/base.py
+++ b/ultraplot/axes/base.py
@@ -166,8 +166,9 @@ This is similar to `matplotlib.axes.Axes.inset_axes`.
 
 Parameters
 -----------
-bounds : 4-tuple of float
-    The (left, bottom, width, height) coordinates for the axes.
+bounds : 4-tuple of float or (4-tuple, transform)
+    The (left, bottom, width, height) coordinates for the axes. To specify the
+    coordinate system alongside the coordinates, pass ``(bounds, transform)``.
 %(axes.transform)s
     Default is to use the same projection as the current axes.
 %(axes.proj)s
@@ -1071,9 +1072,11 @@ class Axes(_ExternalModeMixin, maxes.Axes):
         """
         Add an inset axes using arbitrary projection.
         """
-        # Converting transform to figure-relative coordinates
-        bounds_input = bounds
-        transform = self._get_transform(transform, "axes")
+        # Convert coordinates to figure-relative coordinates while retaining the
+        # original bounds for aspect-aware geographic inset locators.
+        bounds_input, transform = self._parse_anchor(
+            bounds, transform, default="axes"
+        )
         locator = self._make_inset_locator(bounds_input, transform)
         bounds = locator(self, None).bounds
         label = kwargs.pop("label", "inset_axes")
@@ -1764,6 +1767,22 @@ class Axes(_ExternalModeMixin, maxes.Axes):
             return self.figure.transSubfigure
         else:
             raise ValueError(f"Unknown transform {transform!r}.")
+
+    def _parse_anchor(self, coordinates, transform=None, default="data"):
+        """
+        Parse coordinates and their transform.
+
+        Coordinates can be passed with the transform separately or packaged as
+        a ``(coordinates, transform)`` tuple. The latter is useful for APIs
+        that accept a single anchor argument, such as ``inset_axes``.
+        """
+        if (
+            transform is None
+            and isinstance(coordinates, tuple)
+            and len(coordinates) == 2
+        ):
+            coordinates, transform = coordinates
+        return coordinates, self._get_transform(transform, default)
 
     def _register_guide(self, guide, obj, key, **kwargs):
         """
--- a/ultraplot/axes/base.py
+++ b/ultraplot/axes/base.py
@@ -1074,9 +1074,7 @@ class Axes(_ExternalModeMixin, maxes.Axes):
         """
         # Convert coordinates to figure-relative coordinates while retaining the
         # original bounds for aspect-aware geographic inset locators.
-        bounds_input, transform = self._parse_anchor(
-            bounds, transform, default="axes"
-        )
+        bounds_input, transform = self._parse_anchor(bounds, transform, default="axes")
         locator = self._make_inset_locator(bounds_input, transform)
         bounds = locator(self, None).bounds
         label = kwargs.pop("label", "inset_axes")
--- a/ultraplot/axes/base.py
+++ b/ultraplot/axes/base.py
@@ -750,6 +750,20 @@ class _TransformedBoundsLocator:
         return bbox
 
 
+class _AspectAwareTransformedBoundsLocator(_TransformedBoundsLocator):
+    """Preserve an inset's lower-left anchor after box-aspect adjustment."""
+
+    def __call__(self, ax, renderer):
+        bbox = super().__call__(ax, renderer)
+        aspect = ax.get_aspect()
+        if aspect == "auto" or ax.get_adjustable() != "box":
+            return bbox
+        box_aspect = float(aspect) * ax.get_data_ratio()
+        fig_bbox = ax.figure.bbox
+        fig_aspect = fig_bbox.height / fig_bbox.width
+        return bbox.shrunk_to_aspect(box_aspect, bbox, fig_aspect)
+
+
 class _SideColorbarLocator:
     """Position a side colorbar beyond its parent axes decorations."""
 
@@ -1057,8 +1071,9 @@ class Axes(_ExternalModeMixin, maxes.Axes):
         Add an inset axes using arbitrary projection.
         """
         # Converting transform to figure-relative coordinates
+        bounds_input = bounds
         transform = self._get_transform(transform, "axes")
-        locator = self._make_inset_locator(bounds, transform)
+        locator = self._make_inset_locator(bounds_input, transform)
         bounds = locator(self, None).bounds
         label = kwargs.pop("label", "inset_axes")
         zorder = _not_none(zorder, 4)
@@ -1076,6 +1091,8 @@ class Axes(_ExternalModeMixin, maxes.Axes):
         # automatically if we used data coords. Called by ax.apply_aspect()
         cls = mproj.get_projection_class(kwargs.pop("projection"))
         ax = cls(self.figure, bounds, zorder=zorder, label=label, **kwargs)
+        if ax._name in ("cartopy", "basemap"):
+            locator = _AspectAwareTransformedBoundsLocator(bounds_input, transform)
         ax.set_axes_locator(locator)
         ax._inset_parent = self
         ax._inset_bounds = bounds
--- a/ultraplot/axes/geo.py
+++ b/ultraplot/axes/geo.py
@@ -165,15 +165,15 @@ def _parse_hawkeye_extent_transform(transform):
     raise ValueError("extent_transform must be 'map' or a cartopy CRS.")
 
 
-def _parse_hawkeye_connectors(connectors):
+def _parse_hawkeye_connector(connector):
     """Normalize connector shorthand to a named presentation mode."""
-    if connectors is False:
+    if connector is False:
         return None
-    if connectors is True:
+    if connector is True:
         return "corners"
-    if connectors in ("corners", "line"):
-        return connectors
-    raise ValueError("connectors must be False, True, 'corners', or 'line'.")
+    if connector in ("corners", "line"):
+        return connector
+    raise ValueError("connector must be False, True, 'corners', or 'line'.")
 
 
 def _parse_hawkeye_shape(value, name):
@@ -375,7 +375,7 @@ class _HawkeyeSpec:
     extent: Optional[tuple[float, float, float, float]]
     extent_transform: Any
     relation: str
-    connectors: Optional[str]
+    connector: Optional[str]
     shape: str
     target: str
 
@@ -617,13 +617,13 @@ relation : {'auto', 'detail', 'overview'}, default: 'auto'
     where the extent outline and connectors are drawn.
 indicator : bool, default: True
     Whether to outline *extent* on the parent map when an extent is supplied.
-connectors : {False, True, 'corners', 'line'}, default: False
+connector : {False, True, 'corners', 'line'}, default: False
     The connector presentation. ``True`` and ``'corners'`` draw corner links between
     the extent outline and inset. ``'line'`` draws one leader from the inset boundary
     to the target centre. Requires *extent*.
 shape, target : {'box', 'circle'}, default: 'box'
     The inset clipping shape and target marker shape. Circular targets require
-    ``connectors='line'`` or no connectors.
+    ``connector='line'`` or no connector.
 indicator_kw : dict-like, optional
     Patch properties for the extent outline and connector lines.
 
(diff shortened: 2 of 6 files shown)


cvanelteren:
TODOS 
- [x] fix the indicator inset behaving differently on mpl 3.9
- [x] Do a refactor on `hawkeye` as it is much too long
- [x] Run through the docs to check the examples and visual fidelity
- [x] Need to check what is causing the visual regression
- [x] Get a review

ultraplot/axes/geo.py
```diff
@@ -237,6 +546,60 @@
...
+    The geographic scope ``(west, east, south, north)`` displayed by the inset.
+extent_transform : {'map'} or cartopy CRS, default: 'map'
+    Coordinate system for *extent*. ``'map'`` uses `cartopy.crs.PlateCarree`.
+relation : {'auto', 'detail', 'overview'}, default: 'auto'
+    Whether the inset is a zoomed detail of the parent or an overview containing the
+    parent extent. ``'auto'`` compares the rectangular extent sizes. This determines
+    where the extent outline and connectors are drawn.
+indicator : bool, default: True
+    Whether to outline *extent* on the parent map when an extent is supplied.
+connectors : {False, True, 'corners', 'line'}, default: False
```
gepcel: Would it be better to use singular `connector` as parameter? I am often confused by the singular and plural forms of many parameters in ultraplot. For function names like subplot and subplots, the singular and plural have clearly different meanings. However, for some parameters, the difference in meaning between singular and plural is not obvious at all, but makes it very troublesome to remember.

cvanelteren: I agree on the overload of parameters. Personally I would like to see it reduced to a lot less. There are some for loops that had crazy tuples when iterating over different combinations. Alas as it is apart of the public API reverting it is tricky -- I don´t have the capacity to drive a proper release and deprecation schedule.

cvanelteren: Changed it to singular fyi.

cvanelteren:
I think we can ignore the visual regression as I didn't touch the code for the inset. I am opening this up.

ultraplot/axes/geo.py
gepcel: Alias parameters are the part that I like from proplot. Makes it easier to memerise.

ultraplot/axes/geo.py
```diff
@@ -71,6 +75,350 @@
...
+            height = side / figure_bbox.height
+        return mtransforms.Bbox.from_bounds(
+            x - self._anchor[0] * width,
+            y - self._anchor[1] * height,
+            width,
+            height,
+        )
+
+
+def _parse_hawkeye_anchor(anchor):
```
gepcel: Can you add a paramter for anchor transform, to support point of coordinates to be anchor. Like `ax.hawkeye(anchor=(103, 38), anchor_transform='map/or data/or crs'`

gepcel:
Now the inset axes isn't including in the ultraplot layout engine. If I add an inset axes, it must be inside the figure space, otherwise, it will be cutted.

image

cvanelteren:
I don´t think it has too, placement can be controlled by the user. For example for Cartesian this is possible: 
image

ultraplot/axes/geo.py
cvanelteren: This could be possible as this:

```
ax.hawkeye(
      (103, 38),
      size=0.2,
      transform="map",        # longitude/latitude (PlateCarree)
      anchor="upper right",   # inset box corner
  )
```

cvanelteren:
Ah wait I musunderstood you; yeah  I agree that it has to account for it -- do you have the source for this cuz the example code for me renders correctly:

image

gepcel:
inax = ax.hawkeye(
   ...
    anchor="c",    # Just changed this single line, with your example.
    ...
)

ultraplot/axes/geo.py
gepcel: No, it's not now.

cvanelteren:
The circular inset will now prompt redraw on the limits similar to how the inset do it.

ultraplot/axes/base.py
```diff
@@ -1057,9 +1072,10 @@ def _add_inset_axes(
         """
         Add an inset axes using arbitrary projection.
         """
-        # Converting transform to figure-relative coordinates
-        transform = self._get_transform(transform, "axes")
-        locator = self._make_inset_locator(bounds, transform)
+        # Convert coordinates to figure-relative coordinates while retaining the
+        # original bounds for aspect-aware geographic inset locators.
+        bounds_input, transform = self._parse_anchor(bounds, transform, default="axes")
```
gepcel, line 1077: I didn't understand the logic here. Can you explain it? Or simple example?

gepcel:
I'd like to discuss a few coordinate‑system design aspects of the `hawkeye` function. For clarity, let the parent axes be `ax` and the hawkeye (inset) axes be `iax`.

The `hawkeye` function currently takes three parameters that involve coordinate transformations:

1. **`xy`** – the position of `ax` to put `iax` on. Its coordinate system is specified by the `transform` parameter.  
2. **`extent`** – the data extent of `iax`. Its coordinate system is specified by `extent_transform`.  
3. **`anchor`** – the point on `iax` that will be placed at `xy`. When `anchor` is a tuple of floats, it is always interpreted in `iax.transAxes` coordinates (axes‑relative).  

There is also a `proj` parameter to set the map projection of `iax`.

My questions and suggestions are as follows:

**1. About the `transform` parameter**  
- `transform=ax` → `xy` uses `ax.transAxes` (parent axes coordinates). This is correct.  
- `transform='data'` → `xy` should use `ax.transData` (parent data coordinates). Does `'data'` effectively mean `ax.projection` (i.e., the parent's map projection), or others? I didn't dig deep into the source code and didn't do more tests.  
- `transform='map'` – I left a poor comment about this earlier, I haven't thought it through, please disregard it if I mislead you.  
- Ideally, I would like `transform` to accept projection names or crs directly (similar to the [Proj](https://ultraplot.readthedocs.io/en/latest/api/ultraplot.constructor.Proj.html#proj-table)), so that `xy` can be specified in arbitrary projected coordinates.

**2. About the `anchor` coordinate system**  
I would like to have an additional parameter, say `anchor_transform`, to define the coordinate system for the `anchor` point. Its default should be `'axes'` (matching the current behaviour with `iax.transAxes`, and if `anchor` is str, it is converted to 0~1, also use `iax.transAxes` as coordinate system).  

However, the current implementation forces `anchor=(0.2, 0.3)` to always be axes‑relative. This makes it impossible to:  
- set the anchor to a specific geographic point (e.g., `(103.8, 1.3)` as lon/lat) unless I can pass `anchor_transform='cyl'` (or another projection),  
- Or use a point defined in other projections as the anchor.

I hope this clarifies my design intent. I look forward to your thoughts.

cvanelteren:
1.2 'data' -> ax.transData, that is, I assume here that your input will already be in your projected units. A CRS will go through `projection._as_mpl_transform(ax)` I believe which is cartopy's internal way of casting it into a native mpl transform. A `transform="data"` I read as "coordinates in ax.projection". The transform "data" will only be "PlateCarree" when the parent also has this.
1.3 Ok. Just for reference when set to `map` this will work as `ccrs.PlateCarree` which is how I understood your earlier comment. That is, it is read as "xy" are lon/lat coordinates. 
1.4: A CRS can be directly accepted now, but you are right that I have not wired in the full strings yet. I need to add this in.

For 2 I think this is a good addition to also allow pinning to the data coordinates, it currently defaults to axes.

Does this clear things up?

cvanelteren:
So my read is that

- [x] Add project parsing of strings
- [x] Add `anchor_transform` to place anchor within arbitrary transform

gepcel:
Everything is fine now, apart from the "map" transform. I apologize for my earlier comment—it was poorly considered and may have misled you. Since "map" is not supported as a transform name in the GeoAxes transform system, I believe we should drop it from hawkeye.

Let's keep the transform parsing consistent for both GeoAxes and the hawkeye method.

All others works fine as I tested.

cvanelteren:
Note that the `map` option already predates ultraplot and was added so adding it here actually aligns the api

gepcel (approved):

gepcel (approved):

cvanelteren:
Cool we can test drive it now on main -- be sure to open new issue if you have some annoyances in your day-to-day, then we can fix those too.

PR #783, opened 2026-07-31
Add line to geolegend

gepcel:
Suggest to add "line" to geolegend.


cvanelteren:
Ah damn forgot to check. This needed black. Will fix no worries. Thanks!

PR #793, opened 2026-08-13
feat: add remove to legend

### LINKED ISSUE #792: Question about working other plotting packages like seaborn.
Many other packages provide numerous supplements for Matplotlib. Occasionally, I need a feature that Ultraplot lacks. For instance, I miss useful statistical methods and finely tuned color themes of seaborn. I learned about `ax.external()`, so I gave it a try. Let's look at an example:

```
# %% import seaborn and prepare data
import seaborn as sns
x = np.random.randn(1000)+3
df = pd.DataFrame(x, columns=["Test Values"])

# %% plot with only seaborn and move legend
# legend was moved successfully
ax = sns.histplot(df, kde=True, legend=True)
sns.move_legend(ax, "upper left")

# %% plot with ultraplot and seaborn
# legend cannot be moved.
fig, ax = pplt.subplot()
with ax.external():
    sns.histplot(df, ax=ax, kde=True, legend=True)
    sns.move_legend(ax, "upper right")

```

I understand that working with other packages isn't really Ultraplot's agenda, and cooperation isn't always straightforward. I was also surprised to find that I can't control the legend position in Seaborn's `histplot` with some kind of `loc` argument—perhaps I missed something. Are there other ways to integrate Seaborn with Ultraplot that I'm unaware of?

Maybe we could consider adding some useful statistical fitting methods directly into Ultraplot?

cvanelteren:
Closes #792 
Add missing method to UltraLegend to make it work with seaborn.

### DIFF
--- a/ultraplot/__init__.py
+++ b/ultraplot/__init__.py
@@ -379,3 +379,38 @@ def __dir__():
 
 # Prevent "import ultraplot.figure" from clobbering the top-level callable.
 install_module_proxy(sys.modules.get(__name__))
+
+
+def _patch_seaborn_move_legend():
+    """Compatibility shim so seaborn can move legends on single-axis grids."""
+    try:
+        import matplotlib.axes
+        import matplotlib.figure
+        import seaborn as sns
+        from seaborn.axisgrid import Grid
+
+        from .gridspec import SubplotGrid
+    except Exception:
+        return
+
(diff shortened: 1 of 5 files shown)


gepcel (approved):

cvanelteren:
Not too happy with the monkey patch but it is good enough for now.

PR #795, opened 2026-08-14
Add kde support for hist

### LINKED ISSUE #794: Suggest to add kde to histogram plot.
I noticed that you mentioned something about this earlier in [here](https://ultraplot.readthedocs.io/en/latest/stats.html#Histograms-and-kernel-density) that:

```
In the future, UltraPlot will include options for adding “smooth” 
kernel density estimations to histograms plots using a kde keyword. 
It will also include separate ultraplot.axes.PlotAxes.kde and 
ultraplot.axes.PlotAxes.kde2d commands. The violin() and violinh() 
commands will use the same algorithm for kernel density estimation 
as the kde commands.
```

Adding all kinds of statistical plots to all plotting methods would be a huge task, so perhaps we can tackle it bit by bit. Currently, I'm working on some plots that require KDE for histograms. I've opened a pull request for you to review.

With this pr, make the following examples:

```
# %% Example 1
import numpy as np
import ultraplot as uplt
import seaborn as sns
# Sample data
M, N = 50, 3
x = state.normal(size=(M, N)) + state.rand(M)[:, None] * np.arange(N) + 2 * np.arange(N)

# Sample overlayed histograms
fig, ax = uplt.subplots(ncols=2, nrows=2, refwidth=4, refaspect=(3, 2))
ax[0, :].format(title="Overlaid histograms with density=False", xlabel="distribution", ylabel="count")
ax[1, :].format(title="Overlaid histograms with density=True", xlabel="distribution", ylabel="density")

ax[0].hist(
    x, uplt.arange(-3, 8, 0.2), filled=True, alpha=0.7, edgecolor="k",
    cycle=("indigo9", "gray3", "red9"), labels=list("abc"),
    legend="ul", kde=True)

with ax[1].external():
    sns.histplot(x, kde=True, bins=np.arange(-3, 8, 0.2), ax=ax[1])

ax[2].hist(x[:,0], bins=10, density=True, kde=True,ec='white')
with ax[3].external():
    sns.histplot(x[:,0], kde=True, bins=10,ax=ax[3],stat='density')
```

Image

gepcel:
Are you instrested in adding a kde support for hist for now?

Should close #794

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -22,6 +22,5 @@ dependencies:
   - cftime
   - markdown
   - requests
-  - scipy
   - pip:
       - pycirclize
--- a/environment.yml
+++ b/environment.yml
@@ -22,5 +22,6 @@ dependencies:
   - cftime
   - markdown
   - requests
+  - scipy
   - pip:
       - pycirclize
--- a/ultraplot/axes/plot.py
+++ b/ultraplot/axes/plot.py
@@ -1514,7 +1514,7 @@ stack, stacked : bool, optional
     Whether to "stack" successive columns of {y} data for bar-type histograms
     or show side-by-side in groups. Setting this to ``False`` is equivalent to
     ``histtype='bar'`` and to ``True`` is equivalent to ``histtype='barstacked'``.
-kde : book, optional, default: False
+kde : bool, optional, default: False
     Whether to compute and draw a kernel density line to estimate and smooth the
     distribution on the plot.
 kde_kw : dict, optional
--- a/ultraplot/axes/plot.py
+++ b/ultraplot/axes/plot.py
@@ -65,10 +65,6 @@ __all__ = ["PlotAxes"]
 # This is half of rc['patch.linewidth'] of 0.6. Half seems like a nice default.
 EDGEWIDTH = 0.3
 
-# NOTE: Shared by every command that draws a kernel density estimate so that
-# 'hist' and 'ridgeline' curves are sampled identically by default.
-KDE_POINTS = 200
-
 DataInput: TypeAlias = ArrayLike
 ColorTupleRGB: TypeAlias = tuple[float, float, float]
 ColorTupleRGBA: TypeAlias = tuple[float, float, float, float]
@@ -1948,7 +1944,7 @@ def _parse_kde_kw(kde_kw=None, *, points=None, weights=None):
         "points": _not_none(
             points=kw_line.pop("points", None),
             stepsize=kw_line.pop("stepsize", None),  # backwards compatible alias
-            default=_not_none(points, KDE_POINTS),
+            default=_not_none(points, inputs.KDE_POINTS),
         ),
         "bw_method": kw_line.pop("bw_method", None),
         "weights": _not_none(kw_line.pop("weights", None), weights),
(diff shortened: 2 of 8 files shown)


cvanelteren:
This would require a few extra things. First an optional dependency in the install for scipy; a fallback + handling in the code added here and third tests.

ultraplot/axes/plot.py
```diff
@@ -7157,6 +7159,24 @@ def _apply_hist(
...
+        from scipy.stats import gaussian_kde
+        edges = obj[1]
+        kde_kw = dict(kde_kw or {})
+        density = kw.get('density', False)
+        data2d = xs if xs.ndim > 1 else xs[:, None] # (M, N) data
+        stepsize = kde_kw.pop('stepsize', 300)
+        for i in range(data2d.shape[1]):
+            _x = data2d[:, i]
+            xa = np.linspace(_x.min(), _x.max(), stepsize)
+            ya = gaussian_kde(_x)(xa)
```
cvanelteren: we would need to forward gaussian_kde kwargs too

ultraplot/axes/plot.py
```diff
@@ -7157,6 +7159,24 @@ def _apply_hist(
                 if type(sub) is list:
                     res[i] = cbook.silent_list("Polygon", sub)
         self._update_guide(res, **guide_kw)
+        # add kde line
+        if not kde:
```
cvanelteren: actually we may just want to move this block out in helpers somewhere and call it from there

gepcel:
- [x] First an optional dependency in the install for scipy; 
- [x] a fallback + handling in the code added here and 
- [x] third tests.

3/3 are done.

cvanelteren:
> * [x]  First an optional dependency in the install for scipy;
> 
>     * [ ]  a fallback + handling in the code added here and
> 
>     * [x]  third tests.
> 
> 
> 2/3 are done. I don't know how to add fallback handling, raise an error if scipy is not installed? And I'm not sure if it's good practise to put one assert in one method or it's ok to put them together.
> 
> It's fine if you take this over, or you can leave comments about what else to do, and how to do it.

shouldn't the second bullet be done?

gepcel:
I don't know what you mean by "fallback handling". Maybe you can explain a little or you can take it over.

cvanelteren:
What I would do is make the scipy import optional and do a try except block and produce a warning if it is not installed; fallback  may have been wrongly stated as this would imply we provide the kde itself

ultraplot/axes/plot.py
cvanelteren: ^this still needs to happen

environment.yml
```diff
@@ -22,5 +22,6 @@ dependencies:
   - cftime
   - markdown
   - requests
+  - scipy
```
cvanelteren: conda does not allow for optional deps we could add this to the pyproject.toml and not bloat conda installs

gepcel:
> What I would do is make the scipy import optional and do a try except block and produce a warning if it is not installed; fallback may have been wrongly stated as this would imply we provide the kde itself

Ok, I'll take a look, and add this. I actually did do a research how to calculate kde using pure numpy.

gepcel:
Now you can review.

cvanelteren:
I put some notes down for you to look at!

gepcel:
> I put some notes down for you to look at!

Where did you put notes at? I saw no notes.

image

cvanelteren:
image

cvanelteren (changes requested):

cvanelteren:
Ah they were pending my bad!

gepcel:
I have to clear out that, comments like the following 4 days ago, I never saw it until just now. I'll still provide a long screen clip to prove it. I don't know the reason. Screen clips can be deleted later.


image

cvanelteren:
error was on my end @gepcel !

ultraplot/axes/plot.py
gepcel: > we would need to forward gaussian_kde kwargs too

About this, scipy.stats.gaussian_kde only takes 3 args: `class gaussian_kde(dataset, bw_method=None, weights=None)`[[source]](https://github.com/scipy/scipy/blob/v1.18.0/scipy/stats/_kde.py#L38-L742), ax.hist also takes weights args, these should be same. so `weights` should be extract from kwargs of ax.hist.

So there will be a kw named eigher kde_kw or gaussian_kde like you suggest only accecpted 1 args: bw_method. I suggest add bw_method as a args to ax.hist, if kde is false, just ignore it. I will submit it later.

environment.yml
gepcel: Will do.

gepcel:
The implementation is complete and the code is fairly straightforward. However, I'd like to highlight a few points:

1. In addition to the boolean argument `kde` that controls whether to plot the KDE line, I added a single `**kwargs` named `kde_kw`. This argument accepts two types of parameters: one for generating the KDE line (including `stepsize`, `bw_method`, passed to `_kde_line1d`) and the other for plotting it (all other args passed to `ax.plot`). I decided against adding separate `gaussian_kwargs` and `line_plotting_kwargs` because it would create confusion and unnecessary complexity.

2. There is an import check that raises an error with an explanation if SciPy is not installed. However, there is also another `gaussian_kde` import inside `_apply_ridgeline`. I think that part should be dealt with too. And to avoid duplication, the helper function should probably be refactored to serve both use cases.

3. Currently, the KDE functionality only handles univariate data for 1d hist plotting,  but it can accept multiple columns to generate multiple kde lines. However, `scipy.stats.gaussian_kde` supports bivariate data too, so you might consider extending it for `hist2d` in the future. I have only worked with 1D histogram plotting so far, but I can dig in a little if needed.

4. I added SciPy as a dependency in `pyproject.toml` under a new optional group called `stats`. In the future, any additional dependencies for statistical plotting can be placed there.

ultraplot/axes/plot.py
gepcel: done

ultraplot/axes/plot.py
gepcel: done.

cvanelteren (approved):

cvanelteren:
Feel free to merge

gepcel:
> Feel free to merge

Can I? I don't see the merge option. The refactor codes are too many, I did some simple tests, seems right. You can merge.

Whenever code is refactored to integrate into UltraPlot’s codebase and style, it becomes harder to read. I can only do some tests.

PR #808, opened 2026-09-01
Feature: add signature exposure for static inference

cvanelteren:
UltraPlot does a lot of runtime composition (lazy loading, dynamic dispatch, shared doc merging, and compact documentation signatures), which makes static analysis difficult because analyzers see only source-level declarations rather than the final runtime-expanded API.

This PR now takes a single-source approach:

1. Runtime call signatures remain visible instead of being replaced for documentation.
2. The checked-in `.py` files remain the only authored API representation.
3. Reusable docstring snippets stay DRY in the existing snippet registry.
4. During wheel builds, those snippets are materialized into literal docstrings in the copied build tree.
5. Installed wheels therefore expose ordinary Python docstrings to Pylance without shipping or maintaining generated `.pyi` files.
6. CI builds an isolated wheel and verifies that registered placeholders are fully expanded and that no generated stubs are shipped.

Editable/source installs keep the existing runtime docstring expansion behavior; normal wheel installs get statically visible expanded docstrings.


cvanelteren:
Does this work @munechika-koyo ?

cvanelteren:
TODO: get an editor with pylance

munechika-koyo:
I checked how the static type-checking tool (`pyrefly` in my case) behaves in the PR environment.
The results are below:
Screenshot 2026-09-03 at 9 35 29
Screenshot 2026-09-03 at 9 18 18

Since the docstrings are dynamically created when importing ultraplot at runtime, is it difficult for such tools to handle this information statically?

cvanelteren:
From my understanding yes. Custom stubs would work but that is very not handy to maintain.

cvanelteren:
So had a bit more digging. The solution it seems is that one would have to generate and comit the stubs to the repo in order for it to work with the static viewer properly. This is also how mpl does it.

munechika-koyo:
It's true that management costs might go up, but that seems like the better option for generating stubs in a runtime environment, for example, using mypy's `stubgen` tool.

cvanelteren:
Can you check the editor with the stubs added? (30k jeezzzz)

munechika-koyo:
In my environment with `pyrefly`, I couldn't see the correct generated docstrings on hover because it prioritizes using docstrings in `*.py` files, not `.pyi` files.
I submitted the issue to pyrefly: https://github.com/facebook/pyrefly/issues/4803.

Anyway, there appear to be many `Incomplete` types in `.pyi` files. https://github.com/Ultraplot/UltraPlot/blob/f2571e1f5fe92c49bed1c241a0bd208da9fd55ed/ultraplot/constructor.pyi#L118
Were they caused by automatic generation?
Is it possible to fix the types for them?

cvanelteren:
Most of the codebase is not type annotated yet and the automatic inference fallsback to Incomplete I think

cvanelteren:
Wouldn't this solve all the issues: https://marketplace.visualstudio.com/items?itemName=KiidxAtlas.python-hover

munechika-koyo:
Screenshot 2026-09-03 at 16 34 57
Looks like it doesn't solve it...

cvanelteren:
hmm annoying. `help` ofc works fine

cvanelteren:
I added a dynamic inspector prior to dumping the stubs to expand the stubs, does this work @munechika-koyo ?

cvanelteren:
Note there is another small issue that proplot clearly intended the docs to be used as a replacement. So many of the docstrings are near novels.

cvanelteren:
Nvm I noticed I broker a bunch of stuff for myself now. I will fix that first before I ping again.

cvanelteren:
Ok I got the tool complete to work again in zed: 
image

cvanelteren:
I had to run this

```
  pip install --force-reinstall --no-deps \
    --config-settings editable_mode=compat \
    -e /home/casper/projects/ultraplot
```

cvanelteren:
Ok this is turning more of mess than I wanted it to be. I knew the doc strings were complicated in pro/ultraplot but the extension is something that should be addressed. I would love to go back and just make it verbose and remove a bunch of aliases (we have way too many I think) but this would be so much work..

munechika-koyo:
When hovering over `scatter`,
Screenshot 2026-09-04 at 11 52 16

When hovering over the underwaveline,
Screenshot 2026-09-04 at 11 52 31
I also got better hovering and static type analysis, since `pyrefly` can suggest errors. I presume that it is because `pyrefly` referred to the types implemented in `matplotlib`.

As with many parameter aliases, I think it is important to provide clear, compatible alternatives to make it easy for matplotlib users to migrate. (I'm sometimes confused about `xlocator`, `xformatter`, `xticks`, `xticklabels`, etc.)
It might be good to be compatible with users of older versions, but reducing the available options could make it straightforward to learn how to use `ultraplot`.

cvanelteren:
I personally would love to get rid of all the aliases, I think they are unnecessary and confusing at times. But this is tricky as it was part of the public api in proplot, so dunno if we can just remove them without a deprecation phase. I find it a bit hard to judge how many people are actually using this package other than a handful.  In any case, rewriting it to remove all of those aliases, would be addressing the public facing api which is never that easy to do -- some may like the aliases, some may not.

cvanelteren:
My count says that we have 944 aliases that can be removed in the whole package. For style alone that would be a 74 percent reduction from 179 to 47. Others are less severe.

gepcel:
I use ultraplot a lot, and I use aliases a lot; I couldn't even tell which are aliases and which are not. I understand that this makes things harder for development and many other aspects. But I think it's also not easy to tell which names are more intuitive, and which save more keystrokes (which is part of ultraplot's philosophy).

Maybe you could start by not showing aliases in the documentation, rather than deprecating them all. 

Is it possible to make a mapping table or a rename_helper_function that converts them before passing them into format? May this help make codes clearer?

cvanelteren:
I think one way forward is to (i) separate the aliases from the function parsing, (ii) provide a mapping table on the docs, (iii) maybe deprecate it later. The first step would fix the busy static inference here and help some with the setup. It will also be easier to deprecate (or add) aliases. 

I don't want to remove stuff that is already there unless it is necessary, but some of these functions have crazy number of parameters.  What do you think?

gepcel:
I think with steps (i) separate the aliases from the function parsing, (ii) provide a mapping table on the docs, that will give us more time to think about deprecation, and maybe even better, it might even make deprecation less necessary.

gepcel:
Although I mentioned that I use aliases frequently, I probably only use a small fraction of them. However, my point is that with all 944 aliases available, different users can always pick different aliases they like best—and what suits for one person may not be the same for another. After all those long times, deperating them all will make people confusing too, because, some names they are using are aliases without them knowing it.

When we speak of alias names, we tend to assume that there is a single official name—the one that all aliases are ultimately converted to internally in the codebase. Everything else are just shortcuts that get mapped away.

However, if you go through all those official name and their aliases, you'll quickly notice that they aren't all chosen by the same logic, and are more subjective. This lack of consistency and the fact of subjective can itself become a source of confusion.

For Ultraplot (and Proplot before it), one of the goals has always been to make daily plotting easier. This is achieved not only through the `format` philosophy, but also through the alias system, which plays a crucial role in that ease of use.

cvanelteren:
I agree with that the aim should be to make plotting easier. I have tried to convince people to use ultraplot as I think it is great, but people get kinda lost in the many options that one is faced with -- the aliases don't help there. Tutorials will help ofc, but I dunno myself on exactly what magic bullet would help understanding all the keywords that are there. Potentially a rule would help. Like `linewidth` to `lw`  is sensible but `width` to `figwidth` is less clear if the usage is something outside `Figure.format` or whatever.  They can also bloat clean signatures, like what this PR shows a bit.

I think separating them so that they do exist, but the user can see a cleaned function description may work. This, combined with a map of the available aliases, would then give the best of both worlds I think.

gepcel:
> I think separating them so that they do exist, but the user can see a cleaned function description may work. This, combined with a map of the available aliases, would then give the best of both worlds I think.

I agree, as I said, maybe starting by less alias in documents, less confusing for new uses. Let them keep exising by separating for people get used to them.

cvanelteren:
I think that is a good way forward tbh.

munechika-koyo:
> I think separating them so that they do exist, but the user can see a cleaned function description may work. This, combined with a map of the available aliases, would then give the best of both worlds I think.

I also agree.
Reducing the time required to review the documents (only seeing docstrings or annotations from assisting tools) would help users focus on coding tasks.

cvanelteren:
Merged #818  with this PR and it shows for example the complete like this now (basedpyright):

image

cvanelteren:
@munechika-koyo can you check this PR again. At least on basedpyright is working now.

munechika-koyo:
`pyrefly` is now handling [the issue of prioritizing `pyi`'s docstrings ](https://github.com/facebook/pyrefly/pull/4804), so I will confirm it after the change is released.
However, I presume it should be fine if other tools like basedpyright work well.

cvanelteren:
Need to explore if there is not another way. Potentially we would need to get rid of how the docstrings are expanded.

munechika-koyo:
OK, to avoid adding way too many files, it may be better to offer the "stub" package instead, like `scipy-stubs`.

cvanelteren:
I preferably don't want to manually keep track of the stubs. I can automate it for deving through workflows but it will still be annoying as autocommits would need to be pulled back down to sync up (or regenerated). The major issue is that ProPlot was using DRY principles to weave in docstrings that are fine for run time inspection but not for these static inspectors. Redoing all the doc strings violates that principle, but that would gain having the doc strings where they actually belong (in the function description). What I don't like right now is that the doc strings are defined early on in the document and not where the function actually sits, but I recognize the complexity of the dynamic and DRY stuff

munechika-koyo:
I presume the dynamic strategy in this library is key from the perspective of API simplicity at runtime. (e.g. `format()` method associated with different classes.)
Meanwhile, statically typed implementations in Python seem to be growing increasingly popular, which makes the dynamic one more difficult.

If it is hard to build the stub files into the CI automation, what about moving it into the pre-CD workflow?
I understand it's dangerous since we don't confirm stubs in advance, but we can at least check them using the `stubtest` command.

Anyway, we probably need to find the best way to automate stub creation in CI.

cvanelteren:
pre-commit.ci autofix

cvanelteren:
I have added the stubs for wheel releases -- this way users have it while not needing to add the stubs in the dev version.

PR #817, opened 2026-09-04
Feat/format sequence labels

### LINKED ISSUE #815: Feature: expand format to allow for expansion of xlables and ylabels similar to how titles is run
Titles can be passed as lists, but xlabel will pass it on to all axes. This creates a disconnect when we format with one call when we have non-shared axes. This should be changed.

For example

```
axs.format(titles = [..], xlabel = [..])
```
will split the titles to the available axes but not work for xlabel(etc).

cvanelteren:
Closes #815 
By default ultraplot shares axes which is great but creates an awkward setup when forcing everything through format. Especially for newer users. Suppose we want to make an axis in a usually 2x2 grid, we write:

```
fig, ax = uplt.subplots(nrows = 2, ncols = 2)
```

If we then want to format the titles, we can do so with

```
ax.fomat(titles = "a b c d".split())
```
However, say a novice reader would also like to setup xlabels separately (for whatever reason) then this will not work

```
ax.format(xlabel = "a b c d".split())
```
It will spit out ["a", "b", "c", "d"] on the shared axes -- this seems silly. I think what we should instead assume is that the user wants to put a, b, c, d as xlabels for each axes. This implies two things. First, the sharing level would need to be auto adjusted. Second, we need to parse and pass this across the axes. Limits are still shared as usual (unless passed on as well).

This warrants therefore a separate (and new) sharing level where de decouple labels and limits. 

So we have a couple of new situations

```
ax.formatl(titles =[..] # works as normal
ax.format(xlabel = "x") # works as normal
ax.format(xlabel = [...]) # turns xlabel sharing off but y label is unaffected unless subplot share is changed before.
```

Behind the scenes it will add new bools to make label and tick sharing independent. Overrides can be provided by new api key words

```
ax.format(sharexlabels = False) # or as above
```

The sharing levels are not changed (maybe we should?)

It also adds dict parsing for inputs so we can set directly a particular axis, i.e. 

```
axs.format(xlabel = {(1,2) : "test"}, # for pair indexing
                                     ylabel = {2: "test"} # for linear indexing
)
```


cvanelteren:
This is now combining a few different issues. One being a refactor of `format`. 

We maintain the same sharing levels, but add option for disconnecting sharing limits and labels. Indexing to axis may remove sharing if limits are reset. This creates a stronger identity on the subplotgrid and the individual axes. The subplotgrid can be used to delegate the formatting or plotting, whereas setting particular things that should be shared on index axes, will stop the sharing. Intention of the user is trusted here.

PR #818, opened 2026-09-04
Refactor keyword aliases behind a compatibility layer

cvanelteren:
Continuation of our discussion in #808 

UltraPlot has many (many many) aliases. Some love them, some hate them. This PR proposes to move them away from function signatures, to remain with a smaller set, and pushing the rest into a specific decorator. This allows for a smaller function description in stuff like pylance. 

I added a small applet in the docs that shows all the aliases, and shows where they can be added etc. Open for feedback!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -22,6 +22,9 @@ docs/_build
 docs/_static/ultraplotrc
 docs/_static/rctable.rst
 docs/_static/*
+!docs/_static/alias-explorer-v3.css
+!docs/_static/alias-explorer-v3.js
+!docs/_static/alias-map.svg
 !docs/_static/why_plots/
 *.html
 docs/gallery/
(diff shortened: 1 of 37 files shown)


cvanelteren:
@gepcel  can you test drive this in case I missed some of the aliases?
Basically if you hit errors because some of the aliases are missing is fine. I think this is relatively thorough tho.

gepcel:
I tested on some of my snippets, no errors were raised.

cvanelteren:
What do you think of the applet in the docs(https://ultraplot--818.org.readthedocs.build/en/818/aliases.html) @gepcel?

gepcel:
Excellent job.

cvanelteren:
Ok great then we are merging this but we can always return to this commit in case we encounter errors.

PR #819, opened 2026-09-07
Modernize pyproject.toml configuration and formatting

munechika-koyo:
Update deprecated lint and packaging configuration in `pyproject.toml`.

- Move Ruff settings to `[tool.ruff.lint]` and ignore the isort group with `"I"`.
- Replace setuptools_scm's `write_to` and `write_to_template` with `version_file` and `version_file_template`, preserving the generated version format.
- Use the SPDX license expression `"MIT"` and remove the deprecated license classifier.
- Require `setuptools>=80` and remove the redundant `[toml]` extra and `wheel` build dependency.
- Format `pyproject.toml` using [Tombi](https://github.com/tombi-toml/tombi), a TOML formatter.

Validation:
- Built an sdist and a wheel from that sdist.
- Passed `twine check --strict` for both artifacts.
- Verified wheel version and license metadata.
- Verified Ruff configuration loading and `git diff --check`.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,8 +1,7 @@
 [build-system]
 requires = [
-    "setuptools>=64",
-    "setuptools_scm[toml]>=8",
-    "wheel",
+    "setuptools>=80",
+    "setuptools-scm>=8",
 ]
 build-backend = "setuptools.build_meta"
 
@@ -19,9 +18,8 @@ maintainers = [
 description = "A succinct matplotlib wrapper for making beautiful, publication-quality graphics."
 readme = "README.rst"
 requires-python = ">=3.10,<3.15"
-license = {text = "MIT"}
+license = "MIT"
 classifiers = [
-    "License :: OSI Approved :: MIT License",
     "Operating System :: OS Independent",
     "Intended Audience :: Science/Research",
     "Programming Language :: Python :: 3 :: Only",
@@ -50,16 +48,18 @@ packages = {find = {exclude=["docs*", "baseline*", "logo*"]}}
 include-package-data = true
 
 [tool.setuptools_scm]
-write_to = "ultraplot/_version.py"
-write_to_template = "__version__ = '{version}'\n"
+version_file = "ultraplot/_version.py"
+version_file_template = "__version__ = '{version}'\n"
 
 [tool.ultraplot.core_versions]
 python = ["3.10", "3.11", "3.12", "3.13", "3.14"]
 matplotlib = ["3.9", "3.10"]
 
 
-[tool.ruff]
-ignore = ["I001", "I002", "I003", "I004"]
+[tool.ruff.lint]
+ignore = [
+    "I",  # isort (import order)
+]
 
 [tool.basedpyright]
 exclude = ["**/*.ipynb"]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,40 +1,33 @@
-[build-system]
-requires = [
-    "setuptools>=80",
-    "setuptools-scm>=8",
-]
-build-backend = "setuptools.build_meta"
-
 [project]
 name = "ultraplot"
-authors = [
-    {name = "Casper van Elteren", email = "caspervanelteren@gmail.com"},
-    {name = "Luke Davis", email = "lukelbd@gmail.com"},
-]
-maintainers = [
-    {name = "Casper van Elteren", email = "caspervanelteren@gmail.com"},
-    {name = "Matthew R. Becker", email = "becker.mr@gmail.com"},
-]
 description = "A succinct matplotlib wrapper for making beautiful, publication-quality graphics."
 readme = "README.rst"
 requires-python = ">=3.10,<3.15"
 license = "MIT"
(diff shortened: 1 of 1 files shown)


pyproject.toml
```diff
@@ -1,42 +1,33 @@
...
-    {name = "Matthew R. Becker", email = "becker.mr@gmail.com"},
+  { name = "Casper van Elteren", email = "caspervanelteren@gmail.com" },
+  { name = "Matthew R. Becker", email = "becker.mr@gmail.com" },
 ]
-description = "A succinct matplotlib wrapper for making beautiful, publication-quality graphics."
-readme = "README.rst"
-requires-python = ">=3.10,<3.15"
-license = {text = "MIT"}
 classifiers = [
-    "License :: OSI Approved :: MIT License",
```
cvanelteren, line 24: License is gone; is the new default to not add it in the toml?

pyproject.toml
```diff
@@ -1,42 +1,33 @@
...
 ]
-dependencies= [
-    "numpy>=1.26.0",
-    "matplotlib>=3.9,<3.11",
-    "pycirclize>=1.10.1",
-    "typing-extensions; python_version < '3.12'",
+dependencies = [
+  "matplotlib>=3.9,<3.11",
+  "numpy>=1.26.0",
+  "pycirclize>=1.10.1",
```
cvanelteren: I should make this optional. Non-blocker here.

pyproject.toml
```diff
@@ -45,53 +36,62 @@ dynamic = ["version"]
 "Issue Tracker" = "https://github.com/ultraplot/ultraplot/issues"
 "Source Code" = "https://github.com/ultraplot/ultraplot"
 
-[tool.setuptools]
-packages = {find = {exclude=["docs*", "baseline*", "logo*"]}}
-include-package-data = true
-
-[tool.setuptools_scm]
-write_to = "ultraplot/_version.py"
-write_to_template = "__version__ = '{version}'\n"
```
cvanelteren, line 53: Is this deprecated in the newer setuptools?

cvanelteren: Nvm it's below

pyproject.toml
cvanelteren: Nvm it's above

cvanelteren:
LGTM

cvanelteren (approved):

pyproject.toml
```diff
@@ -1,42 +1,33 @@
...
-    "wheel",
-]
-build-backend = "setuptools.build_meta"
-
 [project]
 name = "ultraplot"
+description = "A succinct matplotlib wrapper for making beautiful, publication-quality graphics."
+readme = "README.rst"
+requires-python = ">=3.10,<3.15"
+license = "MIT"
```
cvanelteren, line 6: may need to add license file to this

cvanelteren (dismissed):

pyproject.toml
cvanelteren: I think this is how you do it:
```
license-files = ["LICENSE"]
```

pyproject.toml
munechika-koyo: I found the description about it:
https://github.com/Ultraplot/UltraPlot/blob/99ff644ebd8580f592a782e40caee4b8fa72b2a9/ultraplot/axes/plot_types/circlize.py#L55-L59

So, I moved `pycirclize` into the `circos` optional dependency table.

pyproject.toml
munechika-koyo: I added it.

pyproject.toml
cvanelteren: Yeah I would ignore and remove the edits for this as it would require making it optional in the src. I think I need to do a pass on the deps for the docs again as. I believe I did it like this as conda does not allow to separate these deps, but pip does.

pyproject.toml
```diff
@@ -44,60 +36,68 @@ dynamic = ["version"]
...
-
-[tool.pytest.ini_options]
-filterwarnings = [
-    "ignore:'parseString' deprecated - use 'parse_string':DeprecationWarning:matplotlib._fontconfig_pattern",
-    "ignore:'resetCache' deprecated - use 'reset_cache':DeprecationWarning:matplotlib._fontconfig_pattern",
-]
-mpl-default-style = { axes.prop_cycle = "cycler('color', ['#4c72b0ff', '#55a868ff', '#c44e52ff', '#8172b2ff', '#ccb974ff', '#64b5cdff'])" }
 [project.optional-dependencies]
 all = [
     "ultraplot[circos,docs,stats]",
```
munechika-koyo, line 41:
```suggestion
  "ultraplot[circos,docs,stats]",
```

munechika-koyo:
Should we add [`tombi-pre-commit`](https://github.com/tombi-toml/tombi-pre-commit) to pre-commit-config?

cvanelteren:
sure why not, can you write it up @munechika-koyo ?

munechika-koyo:
OK, I will submit a different PR since I deleted this branch.

PR #823, opened 2026-09-08
Fix #639: unskip and harden the vcenter colormap test

### LINKED ISSUE #639: Turn a skipped vcenter 2D plot check into a deterministic regression test
`ultraplot/tests/test_2dplots.py` has skipped tests with notes like "not sure what this does" around `vcenter` / auto-diverging behavior. A good first step is to take one of those cases and make the expected behavior explicit.

This issue is intentionally scoped for newcomers to the project. It is low complexity, focused on a single test case, and should help someone get familiar with the 2D plotting API and the current test conventions.

Suggested scope:
- pick one skipped case such as `test_colormap_vcenter()` or `test_autodiverging2()`
- write down the intended behavior in the test docstring or assertions
- replace the skipped placeholder with deterministic assertions or a narrowly scoped image-comparison test

Acceptance criteria:
- at least one skipped `vcenter` / auto-diverging test is unskipped
- the expected behavior is clearly documented in the test itself
- the new test passes reliably in CI

jessiedoe03-cpu:
Fixes #639.

`test_colormap_vcenter` was skipped with *"not sure what this does"*. The test
body already called `pcolor(data, vcenter=...)`, so the intent was there - but
`vcenter` was never plumbed through `PlotAxes._parse_cmap`. It fell through to
the artist and raised:

```
AttributeError: PolyQuadMesh.set() got an unexpected keyword argument 'vcenter'
```

so the test could not simply be unskipped; the parameter had to work first.

### What this does

- Unskips `test_colormap_vcenter` and replaces the image comparison with
  deterministic assertions: the centre maps to 0.5 in both discrete and
  continuous mode, and `vcenter=` matches `norm_kw={'vcenter': ...}`.
- Accepts `vcenter` in `_parse_cmap` alongside `vmin`/`vmax`, forwarding it into
  `norm_kw`, and extends the existing "don't combine these with an explicit
  `Normalize` instance" guard to cover it.

### One decision worth your call

When `vcenter` is given and no `norm` is specified, this selects the diverging
normalizer. That seemed the only sensible reading, and it is what the original
skipped test implied - but it is a behaviour choice rather than pure plumbing,
so say if you would rather it were explicit or named differently.

### Testing

- `test_colormap_vcenter`: 1 passed, 1 warning in 1.66s
- Whole file: 1 failed, 32 passed, 1 skipped, 97 warnings in 11.66s
- With `ultraplot/axes/plot.py` reverted and the new test kept: 1 failed, 1 warning in 1.50s -
  so the test genuinely covers the change rather than passing regardless.

Any remaining failures in that file are pre-existing `_tkinter.TclError` issues
in my environment (no display); they fail on the base commit too, and which one
trips varies with run order. Please let CI have the final word.

---

*This patch was written with AI assistance (the code by Google Antigravity, the
review and this pull request prepared with Claude Code) and was reviewed and
tested before submitting by @jessiedoe03-cpu. Happy to revise or close it if it is not the
direction you want.*

🤖 Generated with [Claude Code](https://claude.com/claude-code)


cvanelteren:
Goes against our AI policy. Do need to address this separately if this truly solves it or not.

PR #826, opened 2026-09-10
Create one master baseline and compare against that

cvanelteren:
This PR re-organizes the tests such that we create the baseline images once, and then test the test matrix  against this


cvanelteren:
Closing this as the drift is caused by small difference in for example text rendering and on our end having some rounding for nice effects.

